PR #14, opened 2023-12-05
Add testing with `pytest`

windymilla:
Also add to Github Actions

Make Preferences use different JSON file (and delete it) when testing, to avoid testing affecting normal use, and vice versa.


windymilla:
Fixes #12

PR #22, opened 2023-12-09
Use PanedWindow to hold text and image

windymilla:
User can now adjust proportion of window used for text & image.


windymilla:
Fixes #18

PR #24, opened 2023-12-11
Add macOS dev instructions

cpeel:
This adds dev instructions for macOS and some slight restructuring to accommodate two platforms plus some common bits.

I'm guessing at the `poetry use` incantation for Windows -- it should point to the python binary. Without this poetry will use the first version of python on the path. For Windows this might not be a problem, but for macOS it is as macOS comes with a version and multiple can be installed concurrently from Homebrew. To keep the instructions parallel I added the Windows example but we can leave that out if it's too confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,17 @@
 
 Guiguts rewrite using Python/tkinter
  
- 
-## Windows-focused dev installation
+## Windows Development Setup
 
 Minor alterations probably needed for other platforms
 
-## Install Python
+### Install Python
 
-    1. Download 3.10 from python.org
-    2. Install – default dir is `C:\Users\<username>\AppData\Local\Programs\Python\Python310`
-    3. Ensure this dir is in PATH variable
+1. Download 3.10 from [python.org](https://www.python.org/).
+2. Install – default dir is `C:\Users\<username>\AppData\Local\Programs\Python\Python310`
+3. Ensure this dir is in PATH variable
 
-## Install Poetry
+### Install Poetry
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,32 +2,67 @@
...
 2. Clone the [GG2 Github repo](https://github.com/windymilla/Guiguts-2.0-tkinter)
-3. In the cloned GG2 directory, `poetry install`. The message
-   ".../guiguts does not contain any element" can be ignored (or suppressed by using
-   `poetry install --no-root` instead).
+   or a fork thereof.
+3. In the cloned GG2 directory, create a virtual environment using the version of
+   python you installed above.
+   * Windows
+     ```
+     poetry env use C:\Users\<username>\AppData\Local\Programs\Python\Python310\python3
```
cpeel, line 53: If this is executed within a bash shell of some sort on Windows, it should probably use `/`s. And there's very likely some env variable that could be used to shorten the user's home directory and wouldn't require them updating `<username>` but I don't have a clue what that is.

windymilla: In the Git Bash shell, the necessary command is
` poetry env use ~/AppData/Local/Programs/Python/Python310/python.exe`
Note that `python3` doesn't exist, and it fails without the `.exe`

windymilla (changes requested):

README.md
cpeel: Thanks - I suspected my wild-ass guess was going to be wrong 😁

windymilla (approved):

PR #30, opened 2023-12-14
Enable package building

cpeel:
This PR enables GG2 to be built and used as a python package. Notable changes:

* Package name in `pyproject.toml` was renamed from `guiguts-2-0-tkinter` to `guiguts`. Note that there are no other published packages with the name `guiguts` at [pypi](https://pypi.org/search/?q=guiguts) so this is a pretty safe move.
* Python code was moved from `/src` to a `/src/guiguts` subdirectory as `guiguts` is our package name (see also  [Packaging Python Projects](https://packaging.python.org/en/latest/tutorials/packaging-projects/)).
* The `guiguts.py` file was renamed `application.py`, this is necessary to disambiguate imports from the `guiguts` package and the `guiguts` module. Open to other names.
* `application.py` was updated to allow it to be run during development just like it was before (`python application.py`) and also it's now executable on *nix systems so it can be run as `./application.py`.
* Imports were changed to package-level imports (`guiguts.preferences` vs `preferences`), this is required as part of using this code as a package.
* When installed as a package, the package will include a script where it can be run simply with the command `guiguts`. Also as a module it can be run with `python -m guiguts`.
* Now that it can be installed as a package, we no longer need to use `--no-root`. Poetry will now install this as an editable package, meaning it can be used as a package from within the virtual environment "live" so any code changes are immediately available.

Fixes https://github.com/windymilla/Guiguts-2.0-tkinter/issues/28


src/guiguts/application.py
```diff
@@ -74,7 +79,7 @@ def open_document(self, args):
 
         Accepts a list of filenames, but only loads the first.
         """
-        file.load_file(args[0])
+        self.file.load_file(args[0])
```
cpeel, line 82: This is an untested change but I think this is correct as `file.py` does not have a `load_file()` function -- the `File` class does though.

cpeel:
I broke the doc builds and will work on getting that fixed in this PR here in a bit.

cpeel:
Done -- apologies for the whack-a-mole there at the end.

src/guiguts/application.py
windymilla: Thanks for catching that

windymilla (approved):

PR #33, opened 2023-12-15
Use dynamic homebrew path

tangledhelix:
The install directory for Homebrew varies on Intel and Apple Silicon systems. This command works for either.

Technically, it's possible to install Homebrew in any path. So this approach would also find Homebrew if installed in a custom path that is not the normal platform default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ brew install poetry
      ```
    * macOS
      ```bash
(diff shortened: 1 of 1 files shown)


cpeel:
Awesome, thanks for the update!

srjfoo (approved):

windymilla (approved):
Thanks!

PR #67, opened 2023-12-19
Image improvments

### LINKED ISSUE #62: Prompt to set image (pngs) directory
Currently, the directory that contains the page images must be in the same location as the text file. It's useful to be able to leave the images where they are, and point to where the directory is, instead of having to move them. Not critical, but nice to have.

windymilla:
1. Allow user to choose png dir by right-clicking "See Img"
2. Enable Auto Img by double-clicking "See Img"

Fixes #62

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -49,7 +49,7 @@ class Guiguts:
         maintext().focus_set()
         maintext().add_modified_callback(self.update_title)
 
-        self.auto_image = preferences["AutoImage"]  # For side effects
+        preferences.run_callbacks()
 
     @property
     def auto_image(self):
@@ -163,7 +163,12 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         """
         preferences.set_default("ImageWindow", "Docked")
         preferences.set_default("Bell", "VisibleAudible")
+
+        def set_auto_image(value):
+            self.auto_image = value
+
         preferences.set_default("AutoImage", False)
+        preferences.set_callback("AutoImage", set_auto_image)
 
     # Lay out menus
     def init_menus(self, menubar):
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -49,6 +49,28 @@ class Guiguts:
         maintext().focus_set()
         maintext().add_modified_callback(self.update_title)
 
+        self.set_auto_image(preferences["AutoImage"])
+
+    def get_auto_image(self):
+        """Return whether to show scan images automatically.
+
+        Returns:
+            True if scan images should be shown automatically.
+        """
+        return preferences["AutoImage"]
+
+    def set_auto_image(self, value):
+        """Set whether to show scan images automatically."""
+        preferences["AutoImage"] = value
+        statusbar().set("see img", "Auto Img" if value else "See Img")
+        if value:
+            self.auto_image_check()
+
+    def auto_image_check(self):
+        if preferences["AutoImage"]:
+            self.load_image()
+            root().after(200, self.auto_image_check)
+
     def run(self):
         """Run the app."""
         root().mainloop()
@@ -113,6 +135,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         else:
             self.mainwindow.float_image()
 
+    def toggle_auto_image(self):
+        self.set_auto_image(not self.get_auto_image())
+
     def spawn_process(self, *args):
         """Spawn a subprocess.
 
@@ -142,6 +167,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         """
         preferences.set_default("ImageWindow", "Docked")
         preferences.set_default("Bell", "VisibleAudible")
+        preferences.set_default("AutoImage", False)
 
     # Lay out menus
     def init_menus(self, menubar):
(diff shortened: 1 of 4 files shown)


src/guiguts/application.py
```diff
@@ -49,6 +49,28 @@ def __init__(self):
         maintext().focus_set()
         maintext().add_modified_callback(self.update_title)
 
+        self.set_auto_image(preferences["AutoImage"])
+
+    def get_auto_image(self):
```
cpeel, line 54: Python has a cool featured called properties that can streamline things like this to make them more intuitive to use, eg:
```python
@property
def auto_image(self):
    return preferences["AutoImage"]

@auto_image.settr
def auto_image(self, value):
    preferences["AutoImage"] = value
    statusbar().set("see img", "Auto Img" if value else "See Img")
    if value:
        self.auto_image_check()
```
Then you can use them like:
```python
self.auto_image = not self.auto_image
```

src/guiguts/application.py
```diff
@@ -49,6 +49,28 @@ def __init__(self):
...
+
+    def set_auto_image(self, value):
+        """Set whether to show scan images automatically."""
+        preferences["AutoImage"] = value
+        statusbar().set("see img", "Auto Img" if value else "See Img")
+        if value:
+            self.auto_image_check()
+
+    def auto_image_check(self):
+        if preferences["AutoImage"]:
```
cpeel, line 70: I would use `self.get_auto_image()` here -- or `self.auto_image`.

src/guiguts/application.py
cpeel: Also I need to look at the preferences class a bit closer as this seems like a bit of an odd pass-through but clearly we are taking action on changes not just changing the value.

src/guiguts/application.py
```diff
@@ -234,8 +260,12 @@ def init_statusbar(self, statusbar):
         statusbar.add("prev img", text="<", width=1)
         statusbar.add_binding("prev img", "<ButtonRelease-1>", self.file.prev_page)
 
-        statusbar.add("see img", text="See Img", width=7)
+        statusbar.add("see img", text="See Img", width=9)
         statusbar.add_binding("see img", "<ButtonRelease-1>", self.load_image)
+        statusbar.add_binding(
+            "see img", "<ButtonRelease-3>", self.file.choose_image_dir
+        )
+        statusbar.add_binding("see img", "<Double-Button-1>", self.toggle_auto_image)
```
cpeel, line 268: There's nothing wrong with this, but if we only plan to use `toggle_auto_image()` in this one place, we can probably just make it a lambda (assumes code is using properties):
```python
statusbar.add_binding("see img", "<Double-Button-1>", lambda: self.auto_image = not self.auto_image)
```

cpeel (approved):
Code looks good to me. I've added comments about some python-esque ways you might not be aware of.

src/guiguts/application.py
windymilla: Funny, cos I was using property for this and elsewhere, but I was concerned the side effects of this one were too big to be a good use of properties, so reverted to regular methods. Is it OK to use properties  when there are significant side effects? Thanks.

cpeel: IMHO this is a fine use-case for properties. I'd probably look askance at something that was doing something potentially high-latency (external system calls, network traffic, etc) because I suspect it would be a surprise to a developer to have an get/set take a "long time" but this feels reasonable -- particularly since the only non-setter stuff happening are UI changes and this is a UI app. In general it's a better practice in python to use properties over getters/setters.

**However** one could easily that a property should do nothing that isn't necessary to get/set the attribute and activities like calls to update the UI violate that, and this clearly falls into that case.

windymilla: Thanks. I'll revert to using property along those lines. It was when I added the call to the repeating image check, I bottled out. Good to know that "reasonable" size effects are ok, thanks.

cpeel: You bet. And apologies for mansplaining properties to you ☹️

windymilla: Not at all. You've added to my understanding, as usual. 
And of course I have no experience to rely on about where to draw the line or how to organise things pythonically, so please keep telling me stuff, even if you suspect I might have seen it in a tutorial.
I'm sure when you or other experienced devs get a chance to look at other stuff I've already done, you'll see quite a lot to improve. I keep thinking of things, even as I use what I've already done, so I'm sure you will.

srjfoo (approved):
@windymilla -- Current version tested and approved, but I'm going to assume that I should re-test it if you make changes.

One comment about assigning the image directory -- I don't know if it needs to be fixed or not, but if the cursor is above the first page separator, no image is shown. Locating the image directory does appear to work, even so, because (at least with Auto Img turned on) moving the cursor to some actual page text does bring up the correct image.

Comparing against GG1, having the cursor above the first page separator, it is assumed that it's the first page, and the first image opens.

windymilla:
Thanks for mentioning that - I want to leave the behavior as it is for now
- it shouldn't show you are on page 1 if you are above it, nor show you the
page 1 image. I consider that a bug in GG1, albeit one that won't affect
much real-life work.

srjfoo:
> Thanks for mentioning that - I want to leave the behavior as it is for now - it shouldn't show you are on page 1 if you are above it, nor show you the page 1 image. I consider that a bug in GG1, albeit one that won't affect much real-life work.

No problem. It's just that it stands out as a difference between the two. I honestly don't know if anyone but someone who's just testing would be likely to try to open the images immediately, without having navigated to elsewhere in the file.

src/guiguts/application.py
windymilla: @cpeel , apparently you're not allowed to make assignments in a lambda, so I can't do exactly that. Tempted to leave the toggle function for now, since it's also possible we'll make a decision that everything available in the status bar should also be available in the menus, so the toggle function would appear twice. Open to other suggestions though.

cpeel: TIL! I'd leave it as a function. 👍

cpeel (approved):

windymilla:
@cpeel - I had a go at doing something slightly different with preferences following your comment:

> Also I need to look at the preferences class a bit closer as this seems like a bit of an odd pass-through but clearly we are taking action on changes not just changing the value.

It seems like there are two options: previously, after creating the UI, the application had to update any necessary bits of the UI (e.g. the label for "See Img"/"Auto Img") based on the preferences loaded from the JSON file. This latest commit lets the application define a callback for any preferences that have side effects. Then when the UI has been created, all the callbacks are called. 

I'd be grateful for your opinion on whether this is better/worse than the previous one. Although it's only currently relevant to one pref, there will undoubtedly be more as development continues.

Thanks

cpeel:
The introduction of the callback is demonstrably better. I was just poking at the class with an eye to that and attributes vs arrays and will continue noodling that, but you should 100% ship this as-is.

PR #68, opened 2023-12-19
Add Close to File menu

### LINKED ISSUE #59: Add the ability to close active doc and open new one
Current options are Open and Quit. Close should close the current document without quitting the application (prompting for save if there are changes since the last save) and open a new blank document, or leave it with no window, but allow the user to open a new document.

windymilla:
Fixes #59

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -160,7 +160,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_file.add_button("~Open...", self.file.open_file, "Cmd/Ctrl+O")
         menu_file.add_button("~Save", self.file.save_file, "Cmd/Ctrl+S")
         menu_file.add_button("Save ~As...", self.file.save_as_file, "Cmd/Ctrl+Shift+S")
-        menu_file.add_button("~Close", self.close_file)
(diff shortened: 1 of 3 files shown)


cpeel (approved):

srjfoo (approved):
Works beautifully!

A couple of related questions.
* Should closing a document also close an image if there are images open? If you open a text file for a different project, it might be disconcerting to have mismatched image and text.
* Is "close document but keep the application open" mostly a Mac thing? I know GG1 does it, but for Windows applications, I got the idea that when a document is closed, so is the application, and that if you have multiple documents open, you have multiple instances (which is true for GG1, also, actually). So, I wonder how many Windows users actually expect the close functionality. Anyway, the keyboard shortcut for "close the window but keep the application open" is `cmd+w`, if that's of any use.

windymilla:
@srjfoo - requesting re-test with the two features you requested.

srjfoo (approved):

tangledhelix:
> * Is "close document but keep the application open" mostly a Mac thing? I know GG1 does it, but for Windows applications, I got the idea that when a document is closed, so is the application, and that if you have multiple documents open, you have multiple instances

Windows generally does close the app along with the window. IIRC it varies on Linux depending which windowing system you're using.

PR #69, opened 2023-12-20
Show "page 1" if before all page markers

windymilla:
In fact, shows the first page number, even if it's not "1"


windymilla:
Think this is what you wanted, @srjfoo. This PR doesn't have all the other stuff in - just this feature. I'll merge it in if all is OK, then we'll have everything in master.

srjfoo (approved):

cpeel (approved):

PR #77, opened 2023-12-22
Avoid Auto Img dir selection loop

### LINKED ISSUE #70: Auto Img can cause loop
If Auto Img is trying to load an image but can't find it, it pops a directory selection dialog. If the users cancels this, it just pops again.
One solution would be turn off Auto Img if user cancels dir selection.
Another would be not to pop the dir selection at all if because of Auto Img, but that would require the user to manually select the folder.

windymilla:
Previous test was wrong: instead of checking if the scan file exists, it makes sense to check if a scan dir has been chosen.

Now avoids loop, and doesn't turn off Auto Img just because you haven't set up the scan dir.

Fixes #70

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -2,6 +2,7 @@
 """Guiguts - an application to support creation of ebooks for PG"""
 
 
+import os.path
 import re
 import subprocess
 from tkinter import messagebox
@@ -59,6 +60,7 @@ class Guiguts:
         preferences["AutoImage"] = value
         statusbar().set("see img", "Auto Img" if value else "See Img")
         if value:
+            self.image_dir_check()
             self.auto_image_check()
 
     def auto_image_check(self):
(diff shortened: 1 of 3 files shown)


srjfoo (approved):

srjfoo:
Hold off on merging -- It seemed to me in my initial testing that it worked, but I think I finally figured out what was going on for one of the errors that I only saw in the iTerm window after the fact. Error message:
``` python

  File "/opt/homebrew/Cellar/python@3.10/3.10.13_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/tkinter/__init__.py", line 1921, in __call__
    return self.func(*args)

  File "/Users/sharonjoiner/dpdev/guiguts-py/src/guiguts/application.py", line 151, in close_file
    self.mainwindow.clear_image()

AttributeError: 'MainWindow' object has no attribute 'clear_image'
```

I was closing the file to keep gg2 open in order to open another file (since I needed to set auto-img in order to test it. The text and image are cleared, and in this case there was a popup with the same information that was in the iTerm window.

Opening a second file and testing it for absence of infinite loop on canceling dialog worked correctly. File->Close generated the same error (the image did clear, but only after closing the error popup).

Reopening one of the two text files used for testing did not (as in previous versions) remember the location of the image directory). "Quit" quits python without prompting to save, and without further error. (Because it's quitting Python and not the application?)

srjfoo (commented):
The latest commit fixed the error I was seeing, but I got a ton of

> 2023-12-28 14:13:02.309 Python[35823:4571791] +[CATransaction synchronize] called within transaction

when moving the dialog box out of the way (it came up covering the gg2 window, and I wanted to move it out of the way of that and the active iTerm2 window.

windymilla:
That `CatTransaction` error appears to be this: https://stackoverflow.com/questions/74437765/how-can-i-stop-tkinter-from-outputting-catransaction-synchronize-called-withi

It's not related to the GG2 code (though it is related to Tk). It's a macOS error, and there is a comment [here](https://developer.apple.com/forums/thread/722889)  that says:

> I no longer see these log messages after updating to Xcode 15 / macOS 13.6.

PR #78, opened 2023-12-22
Fix sphinx coverage check

windymilla:
Renaming of guiguts module to application stopped the coverage checks from working.

Fixing it showed a couple routines that needed docstring.


cpeel (approved):
Whoops - thanks for fixing this!

PR #79, opened 2023-12-22
Focus on root window on startup

### LINKED ISSUE #58: Make document window live on file open
If you just start up a gg2 instance and start typing, this problem doesn't appear to exist, but currently (confirmed at least on Mac and Windows), when a new document is first opened in gg2 (i.e. a new project's text CTF), the cursor is visible, but not active. This means that
* you can't move the cursor with the arrow keys
* you can't load images

To make the window active, you need to click in the GG window, or, on a Mac, cmd-tab away from and back into the python instance that is running (@windymilla, can you describe the Windows method?).

windymilla:
All 3 additional calls seem necessary.

Thought it might help with #58, but apparently not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ brew install poetry
 ## Common Development Setup
 
 1. Install Python & Poetry (above)
-2. Clone the [GG2 Github repo](https://github.com/windymilla/Guiguts-2.0-tkinter)
+2. Clone the [GG2 Github repo](https://github.com/DistributedProofreaders/guiguts-py)
    or a fork thereof.
 3. In the cloned GG2 directory, create a virtual environment using the version of
    python you installed above.
@@ -63,8 +63,8 @@ brew install poetry
    poetry install
    ```
 
-You can access the virtual environment shell with `poetry shell` and run
-GG2 directly with `poetry run guiguts`.
+You can run then GG2 directly with `poetry run guiguts`. For advanced users, 
+you can access the virtual environment shell with `poetry shell`.
 
 ## Code style
 Guiguts 2 uses [flake8](https://pypi.org/project/flake8) for static code analysis
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -5,7 +5,7 @@ Guiguts Documentation
 Guiguts
 =======
 
-.. automodule:: guiguts
+.. automodule:: application
 
 MainWindow
 ==========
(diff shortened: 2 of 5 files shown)


windymilla:
Converted to draft, since it fixed a problem on Windows, but not the reported problem on macOS from #58 

For the record, on Windows, after typing `poetry run guiguts` in the shell, the GG window did not come to the front and receive focus - the window manager focus remained on the shell window, so you had to Alt-tab or click on the GG window before you began typing into the text window.

windymilla:
OK - PR is live again.

srjfoo:
The "Focus on root window on startup" is working. On macOS it also works in the current `master`. What does not work (and still doesn't, even after the most recent commit), is the focus in a newly opened text document that has not been opened before. I can get the focus there by switching to another app and then back again, but I still get a BELL when I try to move the cursor with the arrow key if trying to use the arrow keys to navigate in a file that I've not opened before.

I'm happy to approve this PR, as it does what the title of the PR says, but it does not fix #58 as far as I can tell.

Actually, looking back at the issue description, my second point was fixed by another PR: if Auto Img is on, and the image directory has been identified, the image comes up just fine. So, now it's just arrow navigation in a newly-opened-for-the-first-time text file. (I.e. previously saved in another text editor but never opened and saved in GG2.)

windymilla:
@srjfoo, I slightly altered the time at which the focus gets set to the text window - instead of doing it straightaway it should now do it when the process is idle. Worth a try, if you don't mind retesting. Thanks

srjfoo:
> @srjfoo, I slightly altered the time at which the focus gets set to the text window - instead of doing it straightaway it should now do it when the process is idle. Worth a try, if you don't mind retesting. Thanks

Sorry for the delay. Retested, and still the same results:
1. Initial blank page with no attempt to load a document, I can type in an empty document just fine.
2.  Loading a file causes the cursor to become unavailable to typing or using the arrow keys
3.  Same for clearing a file (i.e. using "Close" to get an empty document).
4.  In both of cases 2 and 3, navigating away from the window and then back, or physically clicking in the window, fixes the issue.

(macOS Monterey) (I just noticed -- at least on this one, I was prompted for saving -- maybe on when testing #77, I hadn't made any changes.)

srjfoo:
@windymilla -- Regarding this PR -- if I remember correctly, it fixes at least the initial problem that you were having on Windows (i.e. not having focus on even an initial empty document), which, interestingly, was not a problem on macOS. Is it reasonable to merge the fix for Windows, and leave issue #58 open with the comment that the focus issue is fixed for Windows, but not for Macs? It would be good to get the Windows fix merged in.

PR #80, opened 2023-12-23
(Mac) Remove redundant "Quit" item from File menu

tangledhelix:
In Mac applications, the standard location for the "Quit" menu item is under the so-called "Application" menu. Tkinter automatically supplies this by default. The second Quit item in the file menu is redundant.

The default Quit menu item binds Cmd-Q, so the menu item removed by this commit was not actually able to bind Cmd-Q, though it clearly attempted to do so.

I tested both of the Quit menu items. Both behaved in the same way, asking if the user wished to save their changes before exit (if the file state was dirty). As far as I can tell, this removal doesn't create any behavioral change.

Screenshot 2023-12-22 at 11 26 54 PM

Screenshot 2023-12-22 at 11 27 08 PM

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -213,10 +213,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         self.menu_file.add_separator()
         self.menu_file.add_button("Spawn ~Process", self.spawn_process)
-        self.menu_file.add_separator()
-        self.menu_file.add_button(
-            "~Quit", self.quit_program, "Cmd+Q" if is_mac() else ""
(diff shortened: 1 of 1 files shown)


windymilla (changes requested):
Code looks fine and works on Windows. 
Since the reason the button is labelled `Quit` rather than `Exit` is for Mac users, please could you change the label of the button to `"E~xit"` now it won't be shown on Macs - from online searches it seems Linux users will cope with either :)

tangledhelix:
Ok, label has been changed

windymilla (approved):

PR #86, opened 2023-12-28
Allow filename to be given on command line

windymilla:
1. Parse command line args
2. Allow `guiguts -h`
3. Allow `guiguts file.txt` to load file.txt
4. Allow `guiguts -r 1` or `guiguts --recent 9` to load most nth most recent file (up to 9)

Revealed a known tkinter issue where if a dialog is popped before `update_idletasks` is called, focus will not return to the main text window. Fixed, since this commit can cause that to happen if the filename given on the command line does not exist.

Reversed the stored order of recent files, since it made more sense - the first is now the most recent.


srjfoo (approved):
This may be because the branch pre-dated fixing the earlier image problems (i.e. wrong image not cleared when loading a different text file, with Auto Img turned off), but one of the things I wanted to test needed to start out with no image (part of setting up my "recent documents" list). The previous load had text and image both:
image

Other than that, which, is hopefully unrelated, all the commands look good.

Interesting observation: When I open a file from the command line, the arrow keys work immediately just fine. They still do not when opened from within GG, though.

PR #87, opened 2023-12-29
Upgrade python version to 3.11

### LINKED ISSUE #57: Evaluate upgrading python version to 3.12
GG2 is currently being developed against python 3.10 per the `README.md` and it is currently supported for security releases. 👍. However, binaries aren't available for [the most recent 3.10 release](https://www.python.org/downloads/release/python-31013/) on Windows. python 3.12 is the most recent supported version -- see the [python version status](https://devguide.python.org/versions/).

We should consider updating the minimum version to 3.12 so we're developing against the latest version. Windows and macOS (via Homebrew) have ready access to python 3.12. 

The biggest downside and thing to consider is for Linux users. Some very common, supported platforms include older versions:
* Ubuntu 20 (in support through mid 2024) uses python 3.8
* Ubuntu 22 uses python 3.10

Getting newer versions installed on those platforms is doable but often not trivial. We should consult with some of our GG1 Linux users and get their thoughts on python version and either proceed with this work or WONTFIX this issue and punt it to the future.

I don't see us targeting new python features that would preclude users from running on older versions, but for support purposes by having a target version we test and validate against is useful -- and that may be sticking with 3.10.

windymilla:
As per discussion in #57, this is a compromise between using 3.12 and not causing undue hassle for Linux users.

Also, while here, update black version to 23.12, which is the last release using the 2023 stable style.

Also update installation documentation to cover python versions.

Fixes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ brew install poetry
 1. Install Python & Poetry (above)
 2. Clone the [GG2 Github repo](https://github.com/DistributedProofreaders/guiguts-py)
    or a fork thereof.
-3. In the cloned GG2 directory, create a virtual environment using the version(s) of
+3. In the cloned GG2 directory, create a virtual environment using a version of
    python you installed above.
    * Windows - single python version (in git bash shell)
      ```bash
--- a/README.md
+++ b/README.md
@@ -8,9 +8,28 @@ Minor alterations probably needed for other platforms
 
 ### Install Python
 
-1. Download 3.10 from [python.org](https://www.python.org/).
-2. Install – default dir is `C:\Users\<username>\AppData\Local\Programs\Python\Python310`
+#### Single (system-wide) version
+
+1. Download 3.11 from [python.org](https://www.python.org/).
+2. Install – default dir is `C:\Users\<username>\AppData\Local\Programs\Python\Python311`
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -46,25 +65,31 @@ brew install poetry
 1. Install Python & Poetry (above)
 2. Clone the [GG2 Github repo](https://github.com/DistributedProofreaders/guiguts-py)
    or a fork thereof.
-3. In the cloned GG2 directory, create a virtual environment using the version of
+3. In the cloned GG2 directory, create a virtual environment using the version(s) of
```
cpeel, line 68: The virtual environment still only uses one python version. Or perhaps `using one of the versions` or `using a version of`?

cpeel (approved):

README.md
windymilla: Yes, that's clearer - thanks

srjfoo (approved):

tangledhelix (approved):
Tested on macOS 14.2.1 using Pyenv and Python 3.11.7, worked as expected. I did run into a problem, but the same problem is observed using Python 3.10 from Homebrew, so it's not related to Python 3.11; I'll file that in a different issue.

PR #89, opened 2023-12-30
Mypy

windymilla:
Use mypy for static type checking.

Mypy set up to install via poetry and to run as Github Action.

Source code changes: 

Primarily function definitions now have type annotations, plus
a few other places that mypy needed them.

There is no output from a default run of mypy now.

In addition, when mypy is run with `--disallow-untyped-defs`
there is also no output, meaning all function defs have type
hints so all code is being type checked.

A couple of small type errors/dangers were highlighted by the
process and have been fixed.

There are 7 instances of `# type: ignore[...]` directives, generally
to do with either incompatibility between Tkinter and its stubs
file, or just too complicated to sort out to satisfy mypy at the
moment.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -39,6 +39,10 @@ jobs:
     - name: Report undocumented code with sphinx
       run: |
         poetry run python -m sphinx -M coverage docs docs/build -W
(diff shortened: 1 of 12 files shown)


cpeel (approved):
This looks good to me!

PR #92, opened 2024-01-03
Add message log

windymilla:
Outputs messages with a timestamp to a dialog as well as to stderr. Shows the dialog if currently hidden.

Currently supports errors. Other message types could be added later if needed.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -1,6 +1,8 @@
 """Define key components of main window"""
 
 
+from datetime import datetime
+from idlelib.redirector import WidgetRedirector  # type: ignore[import-not-found]
 import os.path
 from PIL import Image, ImageTk
 import re
@@ -8,7 +10,8 @@ import sys
 import time
 import traceback
 import tkinter as tk
-from tkinter import ttk, messagebox
+from tkinter import ttk
+
 from types import TracebackType
 from typing import Any, Callable, Optional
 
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes:
An easy way to cause an error is to edit the `.json`  file that goes with a project file (formerly "bin" file), e.g. adding/removing a comma.

src/guiguts/mainwindow.py
```diff
@@ -593,6 +582,99 @@ def add_binding(self, key: str, event: str, callback: Callable[[], None]) -> Non
...
+
+        self.messagelog = ScrolledReadOnlyText(frame, wrap=tk.NONE)
+        self.messagelog.grid(column=0, row=0, sticky="NSEW")
+
+    def append(self, message: str) -> None:
+        """Append a message to the message log dialog."""
+        self.messagelog.insert("end", message)
+
+
+class MessageLog:
```
cpeel, line 643: We might want to think about introducing standard python logging in the code and have one of the handlers output errors to the `MessageLogDialog`. That could happen separately after this PR however.

cpeel (approved):
Looks good to me caveat changing to standard python logging down the road.

srjfoo:
> Testing notes: An easy way to cause an error is to edit the `.json` file that goes with a project file (formerly "bin" file), e.g. adding/removing a comma. Then try opening the project file.

That certainly caused the error window to pop up. I noticed that it did load the CTF, just didn't load the images. I can see why the images don't load, but wanted to make sure that, in this particular case, the text loading is the expected behavior.

windymilla:
I think that's ok, yes. It should only happen if the user messes with their json file. Thanks

src/guiguts/mainwindow.py
windymilla: I've added an issue #94 to use python logging - looks like just what we need, thanks.

srjfoo (approved):

PR #93, opened 2024-01-03
Fix bugs in prefs/recent files

### LINKED ISSUE #90: GGprefs.json is being saved every "update"
Note to self: Probably something in the repeat timer is changing a pref, which causes a save of GGprefs.json.
May be related to Recent files

### LINKED ISSUE #91: Recent files aren't being saved/updated correctly.
This is really a note to self to investigate something not quite working.

windymilla:
Preferences were being saved every time AutoImg was checking for whether to load an image, because the "docked/floated" flag was set.  Altered so that the prefs are only saved when the value is changed, not if it is set to the same value as it formerly was.

When getting a preference that was a list (like the list of recent files, the code was not doing a `copy` (due to my Python ignorance). This meant that when the new recent file was added, it was inserted directly into the prefs list rather than "setting" the list, hence saving wasn't being triggered. Fixed by using `deepcopy` - in fact, a shallow `copy` would be OK for the current settings stored in prefs, but hopefully a deep copy will avoid similar headaches in the future. Performance is not an issue for the relatively small number of prefs values.

Fixes #90 
Fixes #91

### DIFF
--- a/src/guiguts/preferences.py
+++ b/src/guiguts/preferences.py
@@ -77,12 +77,8 @@ class Preferences:
             key: Name of preference.
             value: Value for preference.
         """
-        print(key, flush=True)
-        print(self.get(key), flush=True)
-        print(value, flush=True)
         if self.get(key) != value:
             self.dict[key] = value
-            print("save", flush=True)
             self.save()
 
     def __del__(self) -> None:
--- a/src/guiguts/preferences.py
+++ b/src/guiguts/preferences.py
@@ -1,5 +1,6 @@
 """Handle preferences"""
 
+import copy
 import json
 import os
 from typing import Any, Callable
@@ -57,10 +58,8 @@ class Preferences:
         self._remove_test_prefs_file()
         self.load()
 
-    def __getitem__(self, key: str) -> Any:
+    def get(self, key: str) -> Any:
         """Get preference value using key.
-
-        Provides `value = preferences[key]`
 
         Args:
             key: Name of preference.
@@ -69,19 +68,22 @@ class Preferences:
             Preferences value; default for ``key`` if no preference set;
             ``None`` if no default for ``key``.
         """
-        return self.dict.get(key, self.defaults.get(key))
+        return copy.deepcopy(self.dict.get(key, self.defaults.get(key)))
 
-    def __setitem__(self, key: str, value: Any) -> None:
-        """Set preference value and save to file.
-
-        Provides `preferences[key] = value`
+    def set(self, key: str, value: Any) -> None:
+        """Set preference value and save to file if value has changed.
 
         Args:
             key: Name of preference.
             value: Value for preference.
         """
-        self.dict[key] = value
-        self.save()
+        print(key, flush=True)
+        print(self.get(key), flush=True)
+        print(value, flush=True)
+        if self.get(key) != value:
+            self.dict[key] = value
+            print("save", flush=True)
+            self.save()
 
     def __del__(self) -> None:
         """Remove any test prefs file when finished."""
@@ -145,7 +147,7 @@ class Preferences:
         for key in self.callbacks.keys():
             callback = self.callbacks[key]
             if callback is not None:
-                callback(self[key])
+                callback(self.get(key))
 
     def _remove_test_prefs_file(self) -> None:
         """Remove temporary JSON file used for prefs during testing."""
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -93,17 +93,17 @@ class Guiguts:
             self.file.load_file(self.args.filename)
         elif self.args.recent:
             index = self.args.recent - 1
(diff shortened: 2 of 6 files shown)


windymilla:
Testing notes: check that recent files now behave as expected: stored in GGprefs file, updated correctly when a file is loaded, either  from the File->Open button, the Recent Files menu, or the `-r N` command line option.

Testing the 200 millisecond repeating update of GGprefs.json is probably easiest checked by having GG2 running with AutoImg turned on, and looking at the time stamp on the GGprefs file - it should keep updating in master, but not with this PR.

src/guiguts/preferences.py
```diff
@@ -57,31 +58,32 @@ def __init__(self) -> None:
...
+    def set(self, key: str, value: Any) -> None:
+        """Set preference value and save to file if value has changed.
 
         Args:
             key: Name of preference.
             value: Value for preference.
         """
-        self.dict[key] = value
-        self.save()
+        print(key, flush=True)
```
cpeel, line 80: Leftover debugging 😁

windymilla: How did you guess? 😁

tangledhelix:
Seems to work fine here (via all the mechanisms mentioned in your testing notes).

src/guiguts/preferences.py
windymilla: Fixed :)

cpeel (approved):

srjfoo:
I finally pinned down a message that I've been getting, and hadn't until know figured out where it was coming from. I don't know if it's something that needs to be worried about (not marked as a fatal error):

> 2024-01-06 03:31:48.669 Python[50799:6561479] +[CATransaction synchronize] called within transaction

It appears to be happening when I move the dialog box that opens when I either go to `File->Open...` or `File->Save As...`.

windymilla:
> I finally pinned down a message that I've been getting, and hadn't until know figured out where it was coming from. I don't know if it's something that needs to be worried about (not marked as a fatal error):
> 
> > 2024-01-06 03:31:48.669 Python[50799:6561479] +[CATransaction synchronize] called within transaction
> 
> It appears to be happening when I move the dialog box that opens when I either go to `File->Open...` or `File->Save As...`.

I think I mentioned this before somewhere else (but maybe I just found it but didn't mention it) - it's a macOS NSOpenPanel bug:
https://developer.apple.com/forums/thread/722889
The final comment in the above-linked thread says "I no longer see these log messages after updating to Xcode 15 / macOS 13.6."

srjfoo:
> > I finally pinned down a message that I've been getting, and hadn't until know figured out where it was coming from. I don't know if it's something that needs to be worried about (not marked as a fatal error):
> > > 2024-01-06 03:31:48.669 Python[50799:6561479] +[CATransaction synchronize] called within transaction
> > 
> > 
> > It appears to be happening when I move the dialog box that opens when I either go to `File->Open...` or `File->Save As...`.
> 
> I think I mentioned this before somewhere else (but maybe I just found it but didn't mention it) - it's a macOS NSOpenPanel bug: https://developer.apple.com/forums/thread/722889 The final comment in the above-linked thread says "I no longer see these log messages after updating to Xcode 15 / macOS 13.6."

Oops, my bad. Looks like I'm going to have to live with it, then, and apologies for forgetting that we'd already discussed it.
1. I'd forgotten that I needed to check that out on both the Intel Mac and the M1. I'm seeing this on the Intel Mac.
2. The Intel Mac is at macOS 13.6.3, with the Xcode command line tools version 2397 (which is, as far as I can tell, the latest version, at least for that version of the OS), and that's where I'm seeing the issue. (I don't have the full Xcode installed, just the command line tools)

tangledhelix:
> The Intel Mac is at macOS 13.6.3, with the Xcode command line tools version 2397 (which is, as far as I can tell, the latest version, at least for that version of the OS), and that's where I'm seeing the issue. (I don't have the full Xcode installed, just the command line tools)

If you're using Homebrew Python (which would use a binary "bottle" package) then the Xcode version probably doesn't matter. The macOS version would, though, since it includes the library that defines NSOpenPanel.

If you built Python from source (using either brew's `--build-from-source` or `pyenv`) then they might both matter.

srjfoo:
> If you built Python from source (using either brew's `--build-from-source` or `pyenv`) then they might both matter.

I'm not that savvy 😁  -- just a standard homebrew install. If there's something that might get rid of the message, I'm willing to give it a try. Or, if we don't figure a way around it, it should be documented/footnoted so that anyone who sees it doesn't freak out. Though to be fair, I think we have time to noodle over possible solutions.

srjfoo (approved):
Sorry for taking so long for the review. I finally figured out that I needed to focus just on what the PR is trying to do. Those parts look good. At least some of what I'm seeing of stuff that doesn't quite work has undoubtedly already been merged.

Issue #60, opened 2023-12-16
Add line numbers


srjfoo:
Line numbers are very useful in a text editor. Users should be able to turn them on and off.

PR #95, opened 2024-01-04
Add line numbers to main text window

windymilla:
Line numbers are shown by default. Can be toggled by right-clicking the first (Line:Col) status bar button (as in GG1). Setting is stored in Prefs, so will be remembered next time program is run.

Merging note: Setting a preference has different syntax after #93 is merged

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,7 @@
 [flake8]
 
-# Recommended settings to avoid line-length conflicts with black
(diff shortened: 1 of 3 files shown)


windymilla:
Fixes #60

cpeel (approved):

srjfoo (approved):
Very nice!

PR #96, opened 2024-01-06
Use `logging` module for messages

### LINKED ISSUE #94: Use python `logging` module
Useful resources:

- https://realpython.com/python-logging/
- https://realpython.com/python-logging-source-code/
- https://docs.python.org/3/howto/logging.html

windymilla:
1. Add `-d` or `--debug` command line option which starts GG in debug mode: messages will contain `filename:linenum`. Also, debug messages will be displayed (see below).
2. Add basic message handler when program starts that just outputs to stderr.
3. Once GUI is initialized, add two more handlers: one to show error messages in a popup messagebox dialog; the other to log all messages in the message log window, which can be viewed at any time via the `View` menu.
4. In debug mode, all messages are written to stderr and the message log. In non-debug mode, debug messages are not displayed at all, and info messages are only written to the message log, not to stderr.
5. Messages show timestamp, severity of message, and the message. Also filename:linenum when in debug mode.

Fixes #94

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -7,7 +7,7 @@ import tkinter as tk
 from tkinter import filedialog, messagebox, simpledialog
 from typing import Any, Callable, Final, TypedDict, Literal
 
-from guiguts.mainwindow import maintext, sound_bell
+from guiguts.mainwindow import maintext, sound_bell, logger
 from guiguts.preferences import preferences
 from guiguts.utilities import is_windows
 
@@ -112,9 +112,7 @@ class File:
         try:
             maintext().do_open(filename)
         except FileNotFoundError:
-            messagebox.showerror(
-                title="File Not Found", message=f"Unable to open {filename}"
-            )
+            logger.error(f"Unable to open {filename}")
             self.remove_recent_file(filename)
             self.filename = ""
             return
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -1,15 +1,18 @@
 """Handle file operations"""
 
 import json
+import logging
 import os.path
 import re
 import tkinter as tk
 from tkinter import filedialog, messagebox, simpledialog
 from typing import Any, Callable, Final, TypedDict, Literal
 
-from guiguts.mainwindow import maintext, sound_bell, logger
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: 

1. A good (only?) way to generate an error from the application code is to try to load a recent file that no longer exists.
2. Mostly to help with testing, two info messages are output at the start: "Guiguts started" and "GUI initialized". These will only be seen on the console in debug mode. Info messages are logged to the message log in debug and non-debug modes, but only the second message will appear in the message log because the first precedes the creation of the GUI.
3. Which messages are directed to which output, and the format of those messages, are easily alterable, either now or later as we see how the messaging system gets used.

src/guiguts/mainwindow.py
```diff
@@ -876,7 +887,4 @@ def statusbar() -> StatusBar:
     return MainWindow.statusbar
 
 
-def messagelog() -> MessageLog:
-    """Return the single MessageLog object"""
-    assert MainWindow.messagelog is not None
-    return MainWindow.messagelog
+logger = logging.getLogger(__package__)
```
cpeel, line 890: The usual convention is to have this at the top of the file after the imports and before the class definitions. I'd probably put it before the constants but both before and after make sense to me.

src/guiguts/file.py
```diff
@@ -7,7 +7,7 @@
 from tkinter import filedialog, messagebox, simpledialog
 from typing import Any, Callable, Final, TypedDict, Literal
 
-from guiguts.mainwindow import maintext, sound_bell
+from guiguts.mainwindow import maintext, sound_bell, logger
```
cpeel, line 10: What I've usually seen -- not saying this is correct -- is for the logger to be defined in every file, not imported from elsewhere. That way anything could freely log without needing to import other code besides `logging`.

cpeel (approved):
Code looks good but I didn't run it myself.

src/guiguts/file.py
windymilla: Thanks - all files that need access to the logger now use `getLogger` at the top of the file.

srjfoo (approved):

PR #98, opened 2024-01-11
Page labels

windymilla:
1. Store page label information in project json file
2. Show page label in status bar
3. Left click page label to go to a page using its page number
4. Allow user to configure page labels by right-clicking label in status bar
5. In config dialog, click style (Arabic/Roman/ditto) or number (number/increment/no count) to cycle between values. This is similar but more streamlined than GG1 version of dialog.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ Sphinx = "^7.2"
 pytest = "^7.4"
 mypy = "^1.8"
 types-Pillow = "^10.1"
+roman = "^4.1"
 build = "^1.0.3"
 
 [tool.poetry.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,7 @@ packages = [{include = "guiguts", from = "src"}]
 [tool.poetry.dependencies]
 python = "^3.11"
 Pillow = "^10.1"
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -17,6 +17,7 @@ Sphinx = "^7.2"
 pytest = "^7.4"
 mypy = "^1.8"
 types-Pillow = "^10.1"
+roman = "^4.1"
```
cpeel, line 20: You want this as a regular dependency (along with Pillow above) rather than a dev dependency.

src/guiguts/dialogs.py
```diff
@@ -0,0 +1,49 @@
+"""Common code/classes relating to dialogs."""
+
+import tkinter as tk
+from tkinter import simpledialog, ttk
+
+
+class OkCancelDialog(simpledialog.Dialog):
```
cpeel, line 7: Nice.

pyproject.toml
windymilla: Oops - so I do! Thanks

cpeel (approved):
This looks good to me caveat that I didn't run it.

srjfoo (approved):

PR #99, opened 2024-01-12
Trap JSON decode errors more cleanly

### LINKED ISSUE #71: Catch `JSONDecodeError` when loading JSON file
Indicates (hopefully) that user has manually edited JSON file badly, and needs to fix it.

windymilla:
Instead of an unhandled exception being reported, log an error if a JSON file cannot be decoded, e.g. if the user has edited it.

Code commonized and used in two places: loading `GGprefs.json` and loading the project json file (the file formerly known as "bin").

It also became apparent that `GGprefs.json` was being loaded too early (when the preferences module was imported into `application.py`, rather than being deliberately initialized at the correct time.)

Order is now: parse args, start logger, load prefs, initialize GUI. This means the logger is initialized correctly in debug mode (set via command line), and provides future opportunity for command line args to suppress loading prefs, which was useful for testing purposes in GG1.

Fixes #71

### DIFF
--- a/src/guiguts/utilities.py
+++ b/src/guiguts/utilities.py
@@ -1,6 +1,12 @@
 """Handy utility functions"""
 
+import json
 import platform
+import logging
+import os.path
+from typing import Any, Optional
+
+logger = logging.getLogger(__package__)
 
 # Flag so application code can detect if within a pytest run - only use if really needed
 # See: https://pytest.org/en/7.4.x/example/simple.html#detect-if-running-from-within-a-pytest-run
@@ -38,3 +44,21 @@ def _is_system(system: str) -> bool:
         if _is_system.system not in ["Darwin", "Linux", "Windows"]:
             raise Exception("Unknown windowing system")
         return _is_system.system == system
+
+
+def load_dict_from_json(filename: str) -> Optional[dict[str, Any]]:
+    """If file exists, attempt to load into dict.
+
+    Args:
+        filename: Name of JSON file to load.
+
+    Returns:
+        Dictionary if loaded successfully, or None.
+    """
+    if os.path.isfile(filename):
+        with open(filename, "r") as fp:
+            try:
+                return json.load(fp)
+            except json.decoder.JSONDecodeError as exc:
+                logger.error(f"Unable to load {filename}\n" + str(exc))
+    return None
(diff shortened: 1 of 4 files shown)


src/guiguts/utilities.py
```diff
@@ -38,3 +44,21 @@ def _is_system(system: str) -> bool:
...
+
+    Returns:
+        Dictionary if loaded successfully, or None.
+    """
+    if os.path.isfile(filename):
+        with open(filename, "r") as fp:
+            try:
+                return json.load(fp)
+            except json.decoder.JSONDecodeError as exc:
+                logger.error(f"Unable to load {filename}\n" + str(exc))
```
cpeel, line 63: Consider `Unable to parse file {filename}` or `Unable to load {filename} -- not valid JSON format` or something similar to indicate to the user that the problem wasn't with _accessing_ the file (doesn't exist, can't read it) but with the parsing (exists and readable but not formatted correctly).

windymilla: Thanks - that would be clearer. The Exception which is output on the line below gives details, e.g. "Line 3, column 4: expected colon" so if they have edited the file, there's a chance they will be able to correct it.

cpeel (approved):

srjfoo:
Tested by editing the project bin-json (removed a comma page separator). The following message resulted (in popup, command line, and message log).

`01:52:17: ERROR - utilities.py:63 - Unable to load /Users/sharonjoiner/Desktop/guiguts-testing/_complete_projects_for_testing/coachhorn/projectID49869a80942c1/projectID49869a80942c1.txt.json -- not valid JSON format
Expecting ',' delimiter: line 7 column 5 (char 106)`

Is that what I should be seeing?

windymilla:
> I do have a question, though. The file was opened anyway. Is that something that should be expected, or should the load be aborted?

Good question. This error should only happen if the user has manually edited the json file, so in one sense, they get what they deserve :) More seriously, given that they will get an error, I think it is clear enough that something has gone wrong, and they have the choice of pressing on regardless, or quitting and trying to fix it. So, I think it's OK for the file to open.
Happy for you to approve if you agree and I'll merge it, or if you feel strongly against it, I could cause the load to be aborted if the json file is corrupt.

srjfoo (approved):
> Happy for you to approve if you agree and I'll merge it, or if you feel strongly against it, I could cause the load to be aborted if the json file is corrupt.

I don't have strong feelings one way or another, but perhaps consider in a later PR giving users the option of aborting with a message like

```Corrupted JSON file```

with "Continue" and "Abort" buttons.

I know that sort of thing is possible, but I don't know if, in this case, it's worth doing, or just documenting in the user manual. Of course they can just quit without changes, and maybe that's the right approach.

PR #102, opened 2024-01-21
Add checksum to bin file to trap mismatches

windymilla:
User might have edited main file in an editor other than GG. This might mean page boundary positions stored in the bin file no longer match the positions in the text file. By storing an MD5 checksum in the bin file, check that the bin file and text file match, and warn if not.

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -1,5 +1,6 @@
 """Handle file operations"""
 
+import hashlib
 import json
 import logging
 import os.path
@@ -21,12 +22,14 @@ NUM_RECENT_FILES = 9
 PAGEMARK_PREFIX = "Pg"
 BINFILE_SUFFIX = ".json"
 
+BINFILE_KEY_MD5CHECKSUM: Final = "md5checksum"
 BINFILE_KEY_PAGEDETAILS: Final = "pagedetails"
 BINFILE_KEY_INSERTPOS: Final = "insertpos"
 BINFILE_KEY_IMAGEDIR: Final = "imagedir"
 
 
 class BinDict(TypedDict):
+    md5checksum: str
     pagedetails: PageDetails
     insertpos: str
     imagedir: str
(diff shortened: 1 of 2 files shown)


src/guiguts/file.py
```diff
@@ -214,6 +217,14 @@ def interpret_bin(self, bin_dict: BinDict) -> None:
         Args:
             bin_dict: Dictionary loaded from bin file
         """
+        md5checksum = bin_dict.get(BINFILE_KEY_MD5CHECKSUM)
+        if md5checksum and md5checksum != self.get_md5_checksum():
+            logger.error(
+                """Main file and bin (.json) file do not match.
```
cpeel, line 223: It's really annoying that python isn't going to remove the line indents of this multiline string like php does with its HEREDOCs to allow us to indent this with the rest of the code easily. And it's cumbersome to use multiple strings to force the code layout (although it works, I think -- not sure what black will do with it though):
```python
logger.error(
    "Main file and bin (.json) file do not match.\n"
    "File may have been edited in a different editor.\n"
    "You may continue, but page boundary positions\n"
    "may not be accurate."
)

File: src/guiguts/mainwindow.py [hunk 2]
```diff
@@ -363,7 +363,9 @@ def do_save(self, fname: str) -> None:
             fname: Name of file to save text to.
         """
         with open(fname, "w", encoding="utf-8") as fh:
-            fh.write(self.get(1.0, tk.END))
+            fh.write(
+                self.get(1.0, f"{tk.END}-1c")
```
cpeel, line 367: Feels a bit weird to have this "get all the text" not match the `get_md5_checksum()`'s version. May not be worth it, but consider having a function in this class that can be used to fetch the window's text contents so we don't need to implement that in different places.

It might be clearer to `.rstrip()` the text of trailing newlines.

cpeel (approved):

src/guiguts/file.py
windymilla: Good news - black accepts that format/layout quite happily.

windymilla:
@srjfoo testing notes: Open a file in GG, save and close it. Open the file in a different editor, make a change and save it. Re-open the file in GG - should get error popup.

To see the "problem" this PR "solves", in `master`, open a project file in GG & save it. In your other editor delete several lines (not page separator lines) and save. Re-load in GG and look at where the page label in the status bar changes as you use arrow keys to scroll down through the file - you should see that page number change no longer matches where the page separator lines are. 

Of course, you would see the same problem in this PR, but you would have been warned :)

srjfoo (approved):

PR #104, opened 2024-01-22
Fix Save As recent files bug

### LINKED ISSUE #103: Recent files bug
Start a new file, type in some text, Save As filename.
Look in Recent Files, and your new filename isn't there.

windymilla:
Save As did not update the recent files list, only Load did.

Fixes #103


windymilla:
Testing notes: Load a file, then use File-->Save As to save it as a new name. Then check File-->Recent Files to see if new name is in the recent files list.

cpeel (approved):

srjfoo (approved):

PR #106, opened 2024-01-23
Add ability to run python tools, beginning with pptxt

windymilla:
Plan is that a Python tool could be run standalone (by a dev in a poetry environment, or a user if they have manually installed the required modules), or it can be imported as a module and run within Guiguts.

First attempt at this is with pptxt, translated/adapted from the ppwb Go version by @G4OEU. Integration into GG by @windymilla 

Unlike GG1, this version has color output, so that characters/words can be highlighted.
This version does not yet contain the ability to click on an error, and be taken to the correct place in the main file.
However, Tools->PPtxt should generate output when run in most project files.


windymilla:
Split into two PRs

PR #110, opened 2024-01-30
Add column (block) Cut, Copy & Paste

windymilla:
1. Add classes to handle Tk Text indexes. GG1 has many occurrences of combining a row/column to get a string Tk Text index, or split an index into the row & column. To reduce this, store in a class which can be constructed with either form, and from which either form can be extracted. Also add a class to handle a single range of such indexes, i.e. a start and end.
2. Add Columnize Selection, which takes the current selection and replaces it with multiple selections in order to form a block from the top-left to the bottom-right of the current selection. This feature may not be required in practice, but is required internally to do Column Cut/Copy/Paste, so I plan to leave it in the menu for now.
3. Make Shift+Mouse-click & drag do column selection. Press Shift, then mouse-click, then drag. Should also generally work to click & drag to start normal select, then press Shift and continue dragging to switch to column select - but not vice versa, i.e. starting column selecting then releasing Shift to do normal select is not supported.
4. Add Column Copy (shortcut Shift+Ctrl+C/Shift+Cmd+C) which either copies a column selection, or columnizes a normal selection and then copies it. If there is a column selection, then normal Copy (Ctrl/Cmd+C) actually does a Column Copy. Format in clipboard is multiple lines delimited by newline characters.
5. Add Column Cut (shortcut Shift+Ctrl+X/Shift+Cmd+X) similar to Column Copy but cuts instead of copying. Normal Cut does Column Cut if there is a column selection.
6. Add Column Paste (shortcut Shift+Ctrl+V/Shift+Cmd+V) which pastes a previously copied/cut column selection. If there is no selected text in the main window the block will be inserted at the insert cursor position. If there is selected text, then selected lines will be replaced with pasted lines. If the clipboard text has fewer lines than the selected text, the clipboard is repeated to fill the space of the selected text. If the clipboard text has more lines than the selected text, then after all the selected text lines have been replaced, the remainder of the clipboard text is pasted in the lines below. Additional lines and/or spaces are added if necessary to make the column paste work as a block if near the end of a file or pasting over variable length lines. Normal Paste does Column Paste if there is already a column selection in the text.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -537,11 +537,14 @@ class MainText(tk.Text):
         min_col = min(col_range.start.col, col_range.end.col)
         max_col = max(col_range.start.col, col_range.end.col)
         for line in range(min_row, max_row + 1):
-            self.tag_add(
-                "sel",
-                IndexRowCol(line, min_col).index(),
-                IndexRowCol(line, max_col).index(),
-            )
+            beg = IndexRowCol(line, min_col).index()
+            end = IndexRowCol(line, max_col).index()
+            # If line is too short for any text to be selected, select to
+            # beginning of next line (just captures the newline). This
+            # is then dealt with in column_copy_cut.
+            if self.get(beg, end) == "":
+                end += "+ 1l linestart"
+            self.tag_add("sel", beg, end)
 
     def selected_ranges(self) -> list[IndexRange]:
         """Get the ranges of text marked with the `sel` tag.
@@ -570,6 +573,11 @@ class MainText(tk.Text):
         for range in ranges:
             start = range.start.index()
             end = range.end.index()
+            # Trap case where line was too short to select any text
+            # (in do_column_select) and so selection-end was adjusted to
+            # start of next line. Adjust it back again.
+            if range.end.row > range.start.row and range.end.col == 0:
+                end += "- 1l lineend"
             string = self.get(start, end)
             if cut:
                 self.delete(start, end)
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -160,6 +160,71 @@ def _process_accel(accel: str) -> tuple[str, str]:
     return (accel, f"<{keyevent}>")
 
 
+class IndexRowCol:
+    """Class to store/manipulate Tk Text indexes.
+
+    Attributes:
+        row: Row or line number.
+        col: Column number.
+    """
+
+    row: int
+    col: int
+
+    def __init__(self, index_or_row: str | int, col: Optional[int] = None) -> None:
+        """Construct index either from string or two ints.
+
+        Args:
+            index_or_row: String index, or int row
+            col: Optional column - needed if index_or_row is an int
+        """
+        if isinstance(index_or_row, str):
+            assert col is None
+            rr, cc = index_or_row.split(".", 1)
+            self.row = int(rr)
+            self.col = int(cc)
+        else:
+            assert isinstance(index_or_row, int) and isinstance(col, int)
+            self.row = index_or_row
+            self.col = col
+
+    def index(self) -> str:
+        """Return string index from object's row/col attributes."""
+        return f"{self.row}.{self.col}"
+
+    def rowcol(self) -> tuple[int, int]:
+        """Return row, col tuple."""
+        return self.row, self.col
+
+
+class IndexRange:
+    """Class to store/manipulate a Text range defined by two indexes.
+
+    Attributes:
+        start: Start index.
+        end: End index.
+    """
+
+    def __init__(self, start: str | IndexRowCol, end: str | IndexRowCol) -> None:
+        """Initialize IndexRange with two strings or IndexRowCol objects.
+
+
+        Args:
+            start: Index of start of range - either string or IndexRowCol
+            end: Index of end of range - either string or IndexRowCol
+        """
+        if isinstance(start, str):
+            self.start = IndexRowCol(start)
+        else:
+            assert isinstance(start, IndexRowCol)
+            self.start = start
+        if isinstance(end, str):
+            self.end = IndexRowCol(end)
+        else:
+            assert isinstance(start, IndexRowCol)
+            self.end = end
+
+
 # TextLineNumbers widget adapted from answer at
 # https://stackoverflow.com/questions/16369470/tkinter-adding-line-number-to-text-widget
 class TextLineNumbers(tk.Canvas):
@@ -179,7 +244,7 @@ class TextLineNumbers(tk.Canvas):
             dline = self.textwidget.dlineinfo(index)
             if dline is None:
                 break
-            linenum = str(index).split(".")[0]
+            linenum = IndexRowCol(index).row
             self.create_text(text_pos, dline[1], anchor="ne", text=linenum)
             index = self.textwidget.index("%s+1line" % index)
 
@@ -232,6 +297,18 @@ class MainText(tk.Text):
         # Set up response to text being modified
         self.modifiedCallbacks: list[Callable[[], None]] = []
         self.bind("<<Modified>>", self.modify_flag_changed_callback)
+
+        self.bind("<<Cut>>", self.smart_cut)
+        self.bind("<<Copy>>", self.smart_copy)
+        self.bind("<<Paste>>", self.smart_paste)
+
+        # Column selection
+        self.column_selecting = False
+        self.bind("<Shift-ButtonPress-1>", self.column_select_click)
+        self.bind("<Shift-B1-Motion>", self.column_select_motion)
+        self.bind("<Shift-ButtonRelease-1>", self.column_select_release)
+        self.bind("<KeyRelease-Shift_L>", self.column_select_stop)
+        self.bind("<KeyRelease-Shift_R>", self.column_select_stop)
 
     def _proxy(self, *args: Any) -> Any:
         """Proxy to intercept commands sent to widget and generate a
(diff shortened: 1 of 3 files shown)


srjfoo:
> 2. Add Columnize Selection, which takes the current selection and replaces it with multiple selections in order to form a block from the top-left to the bottom-right of the current selection. This feature may not be required in practice, but is required internally to do Column Cut/Copy/Paste, so I plan to leave it in the menu for now.

I think there's value in leaving it. If you're really paranoid, you want to see the actual column selection. At least on macOS, if you start by clicking and dragging, and only then press the shift key, you don't see the columnized selection, even though it is columnized.

> 3. Make Shift+Mouse-click & drag do column selection. Press Shift, then mouse-click, then drag. Should also generally work to click & drag to start normal select, then press Shift and continue dragging to switch to column select - but not vice versa, i.e. starting column selecting then releasing Shift to do normal select is not supported.

Works perfectly, though see note above about click&drag and then pressing shift.

A note about standardization: column select was something that I first learned about in TextWrangler/BBEdit -- before DP, I really wasn't using straight text editors that much. In BBEdit, Sublime Text, and the macOS Terminal program `option`+click&drag does the column select (click&drag and then pressing `option` does not switch to column select, but continues as a regular selection). In iTerm2, it's `option`+`cmd`+click&drag. In 3 of the 4 (all but Sublime Text), the cursor changes to `＋` (or at least what looks like a full-width plus sign) while column select is active. I don't know if that's realistic for GG2, but it would be a nice-to-have, maybe.

Another thing that I think would be very useful, if it can be managed, is a zero-width column selection. I tried it, and while I could select a one-character-width selection, and do a column paste overwriting that one character per line, I couldn't do a zero-width selection, to allow me to insert a selected column in between two (am I missing something here?).

> 4. Add Column Copy (shortcut Shift+Ctrl+C/Shift+Cmd+C) which either copies a column selection, or columnizes a normal selection and then copies it. If there is a column selection, then normal Copy (Ctrl/Cmd+C) actually does a Column Copy. Format in clipboard is multiple lines delimited by newline characters.

The menu functions work fine. The shortcuts do not (except for the standard copy/cut/paste -- this is pretty much in line with what we've seen with other shortcuts that are apparently not system global).

> 5. Add Column Cut (shortcut Shift+Ctrl+X/Shift+Cmd+X) similar to Column Copy but cuts instead of copying. Normal Cut does Column Cut if there is a column selection.

See above

> 6. Add Column Paste (shortcut Shift+Ctrl+V/Shift+Cmd+V) which pastes a previously copied/cut column selection. If there is no selected text in the main window the block will be inserted at the insert cursor position. If there is selected text, then selected lines will be replaced with pasted lines. If the clipboard text has fewer lines than the selected text, the clipboard is repeated to fill the space of the selected text. If the clipboard text has more lines than the selected text, then after all the selected text lines have been replaced, the remainder of the clipboard text is pasted in the lines below. Additional lines and/or spaces are added if necessary to make the column paste work as a block if near the end of a file or pasting over variable length lines. Normal Paste does Column Paste if there is already a column selection in the text.

I'm not entirely sure I've tested this part adequately, but I can take another bash at it tomorrow. I think my brain is ready for something else right now 😁 . But as with the others, what I've tested, the menu entries work fine, the shortcuts do not.

-----

So, it all boils down to -- yes, the menu items work, but no, the special shorcuts for macOS do not.

windymilla:
> > 2. Add Columnize Selection, which takes the current selection and replaces it with multiple selections in order to form a block from the top-left to the bottom-right of the current selection. This feature may not be required in practice, but is required internally to do Column Cut/Copy/Paste, so I plan to leave it in the menu for now.
> 
> I think there's value in leaving it. If you're really paranoid, you want to see the actual column selection. At least on macOS, if you start by clicking and dragging, and only then press the shift key, you don't see the columnized selection, even though it is columnized.

OK - will keep Columnize Selection for now. 

> 
> > 3. Make Shift+Mouse-click & drag do column selection. Press Shift, then mouse-click, then drag. Should also generally work to click & drag to start normal select, then press Shift and continue dragging to switch to column select - but not vice versa, i.e. starting column selecting then releasing Shift to do normal select is not supported.
> 
> Works perfectly, though see note above about click&drag and then pressing shift.

Just to check if it's macOS-only or the same as I see (and expected behavior for the code I've written)...
Click & drag, then press the shift key, **then continue dragging**. It should only change to column selecting when you continue to drag with the Shift down, not as soon as you press Shift. This is the same behavior as Notepad++. If you don't get column selecting even when you continue to drag, then something is going wrong. I'm not saying that it can't be changed to switch to column select as soon as you press Shift, but reverting to normal select when Shift is released would be require more internal delving than I want to do at the moment, so I'd prefer to keep it like I've intended.

> A note about standardization: column select was something that I first learned about in TextWrangler/BBEdit -- before DP, I really wasn't using straight text editors that much. In BBEdit, Sublime Text, and the macOS Terminal program `option`+click&drag does the column select (click&drag and then pressing `option` does not switch to column select, but continues as a regular selection). In iTerm2, it's `option`+`cmd`+click&drag. 

I could (almost certainly) switch to make it use `option` (and `Alt` on non-MacOS systems). Apparently MS Word uses Alt, so it sounds like that would make sense for macOS and Windows if I change it. Thanks.

> In 3 of the 4 (all but Sublime Text), the cursor changes to `＋` (or at least what looks like a full-width plus sign) while column select is active. I don't know if that's realistic for GG2, but it would be a nice-to-have, maybe.

May be possible to change the cursor to something different while Option/Alt-dragging - I'll have a look

> Another thing that I think would be very useful, if it can be managed, is a zero-width column selection. I tried it, and while I could select a one-character-width selection, and do a column paste overwriting that one character per line, I couldn't do a zero-width selection, to allow me to insert a selected column in between two (am I missing something here?).

Unfortunately (I did try!) you can't mark a zero-width region as a selection in the Text widget. However, if you just position the insert cursor at the top of the zero-width column, and do column paste, I think it should do what you want.

> > 4. Add Column Copy (shortcut Shift+Ctrl+C/Shift+Cmd+C) which either copies a column selection, or columnizes a normal selection and then copies it. If there is a column selection, then normal Copy (Ctrl/Cmd+C) actually does a Column Copy. Format in clipboard is multiple lines delimited by newline characters.
> 
> The menu functions work fine. The shortcuts do not (except for the standard copy/cut/paste -- this is pretty much in line with what we've seen with other shortcuts that are apparently not system global).

I'll look into the shortcuts in another PR. There must be something I've done wrong, hopefully in my "clever" code to deal with Window/Mac shortcut keys, because if it's not a bug of mine, then it must be some other worse bug to chase down.


> > 6. Add Column Paste (shortcut Shift+Ctrl+V/Shift+Cmd+V) which pastes a previously copied/cut column selection. If there is no selected text in the main window the block will be inserted at the insert cursor position. If there is selected text, then selected lines will be replaced with pasted lines. If the clipboard text has fewer lines than the selected text, the clipboard is repeated to fill the space of the selected text. If the clipboard text has more lines than the selected text, then after all the selected text lines have been replaced, the remainder of the clipboard text is pasted in the lines below. Additional lines and/or spaces are added if necessary to make the column paste work as a block if near the end of a file or pasting over variable length lines. Normal Paste does Column Paste if there is already a column selection in the text.
> 
> I'm not entirely sure I've tested this part adequately, but I can take another bash at it tomorrow. I think my brain is ready for something else right now 😁 . But as with the others, what I've tested, the menu entries work fine, the shortcuts do not.

Hopefully the sentence "If there is no selected text in the main window the block will be inserted at the insert cursor position" in point 6 above is what will cover your earlier query about a zero width selection.

srjfoo:
I've re-tested, with the most recent changes, and:

* Column select works (and shows the cross-hair) whether you press `option`/`alt` before or after starting to drag (as long as you continue dragging afterwards in the latter case).
* Column paste in the middle of a row without overwriting what's there works if you use the edit->column paste command, since there is apparently not a way to do a zero-width column selection.

I don't know (in light of the fact that the highlighting does change to column select in both cases) whether or not you want to keep the "Columnize Selection" command. I think it's totally fair to document the constraints and remove the command, since there are ways to get it to work, and eventually the menus will get very full, anyway. Probably better to document and remove it now before people come to depend on it.

> ... I'm not saying that it can't be changed to switch to column select as soon as you press Shift, but reverting to normal select when Shift is released would be require more internal delving than I want to do at the moment, so I'd prefer to keep it like I've intended.

I honestly don't see a problem with your approach. BBEdit does switch to column select as soon as the option key is pressed but it does not switch back to normal select mode even if you lift your finger off the option key -- unless you move the cursor. And it does not revert to a standard select if you lift your finger off the option key, even if you continue moving the cursor. It will cancel column select only if you cancel the selection.

> I could (almost certainly) switch to make it use `option` (and `Alt` on non-MacOS systems). Apparently MS Word uses Alt, so it sounds like that would make sense for macOS and Windows if I change it. Thanks.

> May be possible to change the cursor to something different while Option/Alt-dragging - I'll have a look

Many thanks for option/alt instead of shift to start a column selection, and also for the cross-hair cursor.

> > Another thing that I think would be very useful, if it can be managed, is a zero-width column selection. <snip>
> 
> Unfortunately (I did try!) you can't mark a zero-width region as a selection in the Text widget. <snip>

tsk!

I thought this was ready to go (punting the Mac shortcuts off to the future), and then I tried a column select across a paragraph break. FWIW, it doesn't work in Sublime Text or Terminal, either, but it does in iTerm2 and in BBEdit.

If you want to punt fixing this off to the future, or just decide not to fix it (and document it as "that's just the way it is"), let me know, and I'll approve it (or just take this post as my approval). I just wanted to mention it in case you thought it was worth fixing at this point.

srjfoo (approved):

src/guiguts/mainwindow.py
```diff
@@ -407,13 +503,232 @@ def get_text(self) -> str:
...
+
+        If more lines in clipboard than selected ranges, remaining lines will be
+        inserted in the same column in lines below. If more selected ranges than
+        clipboard lines, clipboard will be repeated until all selected ranges are
+        replaced.
+        """
+        # Trap empty clipboard or no STRING representation
+        try:
+            clipboard = self.clipboard_get()
+        except tk.TclError:
```
cpeel, line 597: Do we know what conditions this happens in? Does it only happen when the clipboard is empty? Feels like we're swallowing an error and going to silently fail at the paste.

src/guiguts/mainwindow.py
```diff
@@ -160,6 +161,71 @@ def _process_accel(accel: str) -> tuple[str, str]:
     return (accel, f"<{keyevent}>")
 
 
+class IndexRowCol:
```
cpeel, line 164: I really like this abstraction 👍

cpeel (approved):

src/guiguts/mainwindow.py
windymilla: It fails if the clipboard is empty, or if the clipboard does not have a "STRING" format, e.g. it's an image. There doesn't appear to be a way to query the clipboard that won't generate a `tk.TclError`. There's a similar check in the pasting function in GG1:
```
    eval { $clipboard_text = $w->SelectionGet( -selection => "CLIPBOARD" ); };
    return unless ( defined $clipboard_text and length $clipboard_text );
    ...
    go on to paste clipboard contents
```

PR #111, opened 2024-02-02
Handle attempts to undo/redo when stack is empty

windymilla:
If there's nothing to undo/redo, then attempting to undo/redo raises a Tk exception. Trap that case and sound bell instead.


windymilla:
Closing, because this really needs some internal changes in the MainText and/or TextLineNumbers classes. Once those are done, a different method of fixing the undo/redo issue will be used.

PR #112, opened 2024-02-03
Fix macOS shortcut key bindings

### LINKED ISSUE #64: Per-platform key bindings
Something that bothers me in Guiguts is the use of Windows-style keymaps even when run on Macs. Probably 99% of the difference between the platforms would be swapping `Ctrl` for `Cmd`. In other words, like this example:

| Function | Windows | Linux | Mac |
|-----|-----|-----|-----|
| Find | `Ctrl-F` | `Ctrl-F` | `Cmd-F` |
| Save | `Ctrl-S` | `Ctrl-S` | `Cmd-S` |

In most Mac apps, you can use Emacs-style movement keys in text areas / text input boxes. Many Mac users are accustomed to doing so. In the current incarnation of guiguts-py, this works (because no other maps are overriding it, yet). If `Cmd` bindings are used for the application commands / menus, then this should continue to work and it would be an improvement over the prior generation (for Mac users, at least).

Even for users who don't use these keys, using `Cmd` bindings would be more in line with the other apps on the system and it would be much more natural on the platform.

When I use Guiguts, I often hit one of these keys out of habit and trigger a command unintentionally.

For those not familar, Emacs-style mappings possibly relevant to this issue are below. I didn't check which of these overlap with Guiguts - at least several of them do, they may not all overlap.

- `Ctrl-F` `Ctrl-B` move cursor right (forward) or left (backward)
- `Ctrl-P` `Ctrl-N` move cursor to previous / next line (up, down)
- `Ctrl-A` `Ctrl-E` move cursor to beginning / end of current line
- `Ctrl-D` `Ctrl-H` delete character to right / left of cursor
- `Ctrl-K` delete from cursor to end of line
- `Ctrl-O` split line into two lines, at current cursor position
- `Ctrl-T` transpose characters on either side of cursor
- `Ctrl-V` page down

windymilla:
Primary problem is to do with case on the Mac. Having `Ctrl+Shift+S` and `Ctrl+Shift+s` appeared to stop them from working, whereas that is useful on other platforms due to behavior when Caps Lock is on. More details in code comments.

Couple of minor fixes:

When called via shortcut, functions
get called with an event arg, but not when called via menu. Such functions therefore need optional arg.

Also, after `Cmd+o` and `Cmd+Shift+s` which pop file selection dialogs, on macOS focus was not returned to the main window, which can confuse users.

Fixes #64

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -222,7 +222,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         if self.file.open_file(filename):
             self.mainwindow.clear_image()
 
-    def close_file(self) -> None:
+    def close_file(self, *args: Any) -> None:
         """Close currently loaded file and associated image."""
         self.file.close_file()
         self.mainwindow.clear_image()
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -13,7 +13,7 @@ from guiguts.mainwindow import maintext, sound_bell, IndexRowCol, root
 import guiguts.page_details as page_details
 from guiguts.page_details import PageDetail, PageDetails, PAGE_LABEL_PREFIX
 from guiguts.preferences import preferences
-from guiguts.utilities import is_windows, load_dict_from_json, is_mac
+from guiguts.utilities import is_windows, load_dict_from_json
 
 logger = logging.getLogger(__package__)
 
@@ -104,8 +104,7 @@ class File:
                 )
             if filename:
                 self.load_file(filename)
-            if is_mac():
-                root().grab_focus()
+            root().grab_focus()
         return filename
 
     def close_file(self) -> None:
@@ -165,8 +164,7 @@ class File:
             self.filename = fn
             maintext().do_save(fn)
             self.save_bin(fn)
-        if is_mac():
-            root().grab_focus()
+        root().grab_focus()
         return fn
 
     def check_save(self) -> bool:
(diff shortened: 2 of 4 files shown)


windymilla:
@srjfoo - this should be identical in behavior (on Macs) to my branch that you tested previously. It contains keyboard shortcut fixes, but doesn't correct the issue of undo-ing/redo-ing beyond the end of the undo/redo stack, which will be dealt with separately (after the internal line numbers/proxy work).

cpeel (approved):
Code LGTM but I didn't test the keys themselves.

windymilla:
@srjfoo : This should return focus to main text after open or save dialogs on Mac.
Additional test (to see if we can manage without minimize/restore) would be to edit `mainwindow.py` lines 48-50, commenting out 48 & 49, and changing 50 from `focus_set` to `focus_force`, i.e.
```
        #self.call("wm", "iconify", ".")
        #self.call("wm", "deiconify", ".")
        maintext().focus_force()
```

srjfoo:
> @srjfoo : This should return focus to main text after open or save dialogs on Mac. Additional test (to see if we can manage without minimize/restore) would be to edit `mainwindow.py` lines 48-50, commenting out 48 & 49, and changing 50 from `focus_set` to `focus_force`, i.e.
> 
> ```
>         #self.call("wm", "iconify", ".")
>         #self.call("wm", "deiconify", ".")
>         maintext().focus_force()
> ```

Yes! It does! 🎉 🎉 🎉

srjfoo (approved):

Issue #113, opened 2024-02-03
Bug in column copy/paste


windymilla:
Select several lines, copy, paste - it does column paste when it shouldn't

PR #114, opened 2024-02-03
Update line numbers without `_proxy`

windymilla:
The `_proxy` method caused problems with overriding methods, as well as catching tkinter exceptions.

Instead, bind relevant events, and override a handful of methods to update the line numbers.

Also, only update linenumbers once each time process becomes idle, even if several events or methods need them updating.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -236,7 +236,7 @@ class TextLineNumbers(tk.Canvas):
         self.textwidget = text_widget
 
     def redraw(self, *args: Any) -> None:
-        """redraw line numbers"""
+        """Redraw line numbers."""
         self.delete("all")
 
         text_pos = self.winfo_width() - 2
@@ -247,7 +247,7 @@ class TextLineNumbers(tk.Canvas):
                 break
             linenum = IndexRowCol(index).row
             self.create_text(text_pos, dline[1], anchor="ne", text=linenum)
-            index = self.textwidget.index("%s+1line" % index)
+            index = self.textwidget.index(index + "+1l")
 
 
 class MainText(tk.Text):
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: Should be no noticeable change in behavior. Try inserting & deleting text, scrolling up and down, etc., and check that line numbers get updated correctly.
Note that I discovered an unrelated pasting bug, so pasting multiline text behaves weirdly. Issue #113 entered to sort this.

cpeel (approved):

srjfoo:
> Testing notes: Should be no noticeable change in behavior. Try inserting & deleting text, scrolling up and down, etc., and check that line numbers get updated correctly.

Inserting and deleting text, scrolling up and down, all seem to work just fine, the main issue I see is that the line numbers don't align with the text:

image

* the last line number at the end of the file is cut off at the bottom
* the tops of the lower case letters in the body text appears to align with the tops of the line numbers.
* the font size for the numbers _appears_ to be different, but that may just be a font difference. It doesn't appear to affect the alignment (i.e. the line height, baseline to baseline appears to be the same); there's just an offset between line numbers and body text)

windymilla:
Fonts for text widget and line numbers widget should now be the same - hopefully this will mean they now align correctly.

I don't propose tackling the cut-off at the bottom at this point. It's not been introduced by this PR. Possibly related to a difference in the way borders are rendered, or maybe an artefact of the font size difference and will now magically go away with this latest commit.

srjfoo (approved):
> Fonts for text widget and line numbers widget should now be the same - hopefully this will mean they now align correctly.
> 
> I don't propose tackling the cut-off at the bottom at this point. It's not been introduced by this PR. Possibly related to a difference in the way borders are rendered, or maybe an artefact of the font size difference and will now magically go away with this latest commit.

The cut off text at the bottom was, indeed, a result of the font difference. In reviewing the code, I spotted:

```php
            linenum = IndexRowCol(index).row
            self.create_text(text_pos, dline[1], anchor="ne", font=font, text=linenum)
            index = self.textwidget.index(index + "+1l")
```

Would the `anchor="ne"` have anything to do with that?

windymilla:
Is the alignment better now? The NE means top right corner of course, and the coordinates should be the top of the line of text. So the top of the line numbers should align with top of the text.

PR #119, opened 2024-02-04
Fix copy/paste bug related to column select

### LINKED ISSUE #113: Bug in column copy/paste

windymilla:
Bad logic test in smart cut/copy/paste functions meant it sometimes wrongly did column paste instead of default paste.

Fixes #113


cpeel (approved):
Those pesky off-by-one errors are the worst!

srjfoo (approved):

PR #122, opened 2024-02-05
Minor reorg & regex library change

windymilla:
To prep for pptxt tool and for developing search/replace function,
1. Rename `dialogs.py` to `widgets.py` since non-dialog common widgets could be useful to put there.
2. Add a common top level dialog class used as a basis for the Message Log, and many future dialogs, including Search.
3. Use `regex` in place of `re` which doesn't have some some features such as Unicode character classes.

Should be no change in behavior, except you can now hit Escape to close the message log now that it uses the common top level dialog code.


srjfoo:
> Testing notes: most likely thing to have broken is focus returning to the main text window that was fixed recently. it should be doing the same calls, but from a different place, so might have gone wrong.

* When I close a text file that has an image open, the image doesn't go away (but that's true in `master` as well.
* When I undo as far as I can undo, that's to a blank screen, just before the file was opened. (Which does not appear to happen in `master`.)

windymilla:
Thanks for testing - I'll put in a separate issue about the image window, which does need other work too.

The problem of undoing back to a blank screen was the final PR just merged into master, so this branch didn't have it when it was created. I've just rebased, so that fix should be in this branch too now.

cpeel (approved):
Code-level review LGTM!

srjfoo (approved):

PR #126, opened 2024-02-10
Search

windymilla:
Add Search dialog:
1. Accessible via `Ctrl/Cmd+f` or from Search menu.
2. If some text is selected in the main window when the dialog is created, the search box will be prepopulated with that text (only up to the first newline in a multiline selection)
3. Search button searches for the next occurrence of the string. Holding Shift down while clicking searches in the opposite direction.
4. Pressing the Return/Enter key is the same as clicking the Search button.
5. Options Reverse, Case Insensitive, Wrap, Regex should be self-explanatory.
6. The "In selection" option only affects Count and Find All.
7. Count counts the number of occurrences of the string in the file/selection
8. Find All finds all occurrences of the string in the file/selection, and displays them in a "checker" dialog.
9. The checker dialog will in time handle output from all the checking tools, and have features similar to GG1, but for now, it just has the capability to click on a line and jump to the relevant place in the main text window.


windymilla:
@srjfoo - let me know if you need further information on testing beyond what's in the PR description above, thanks.

windymilla:
@cpeel - when you are reviewing the code, there's one area in particular I'd like to know if there is a better way to handle, please.
In summary, `ToplevelDialog` is intended to be the base class for most/all non-modal dialogs in GG2. The difficulty I had was that I needed some code to stop the user creating two Search dialogs, for example. I could do that in `SearchDialog`, but I'd end up doing it in all the dialogs inheriting from `ToplevelDialog`. However to put it in `ToplevelDialog` I ended up writing the function `show_toplevel_dialog` which creates the dialog of the type the user wants, but if a dialog of that type already exists, it just brings that to the front, rather than creating a second copy. I'm sure there must be a (potentially obscure) way of doing something similar, but allowing me to just instantiate the subclass, e.g. `dlg = SearchDialog(...)` instead of `dlg=show_toplevel_dialog(SearchDialog...)`, possibly intercepting the creating/initialization process using one or more of `__new__` or `__init__` as I've seen in example code for singleton classes.

srjfoo:
Mac testing.

> 1. Accessible via `Ctrl/Cmd+f` or from Search menu.

Yes.

> 2. If some text is selected in the main window when the dialog is created, the search box will be prepopulated with that text (only up to the first newline in a multiline selection)

Yes, but only if the previous search dialog has been closed. Will not replace what's in the search field already with the new selection.

> 3. Search button searches for the next occurrence of the string. Holding Shift down while clicking searches in the opposite direction.

Yes.

> 4. Pressing the Return/Enter key is the same as clicking the Search button.

Pressing Return/Enter does work for finding next/previous, but repeated usage of `Cmd`+`f` does nothing except flash the menu (equivalent to starting a new search?). (I think that the "search again" shortcut that I'm most familiar with is `Cmd`+`g`, but I know that's not true for all apps.)

> 5. Options Reverse, Case Insensitive, Wrap, Regex should be self-explanatory.

I thought I knew what `wrap` meant, but in order to have the search result show a selection that started at the end of one line and ended at the beginning of the next, I had to use a regex search and use the form: `first\slast`. But maybe I don't know how to use `wrap`. 😁 

> 6. The "In selection" option only affects Count and Find All.

Seem to work, except as stated above.

> 7. Count counts the number of occurrences of the string in the file/selection

First test, looks good.

> 8. Find All finds all occurrences of the string in the file/selection, and displays them in a "checker" dialog.

The simple search highlight's the search term as it is navigated to (possibly an OS function, as it seems to use the OS default highlighting?). Find all does not. It navigates to the search term, and places the cursor at the beginning of the search term, but unless you have a really painfully obvious cursor, it may be difficult to see. Would it be possible to add this?

windymilla:
> > 2. If some text is selected in the main window when the dialog is created, the search box will be prepopulated with that text (only up to the first newline in a multiline selection)
> 
> Yes, but only if the previous search dialog has been closed. Will not replace what's in the search field already with the new selection.

Thanks - I agree it would be ore helpful if it can do that.

> > 4. Pressing the Return/Enter key is the same as clicking the Search button.
> 
> Pressing Return/Enter does work for finding next/previous, but repeated usage of `Cmd`+`f` does nothing except flash the menu (equivalent to starting a new search?). (I think that the "search again" shortcut that I'm most familiar with is `Cmd`+`g`, but I know that's not true for all apps.)

That is as expected - my wording was a little imprecise. Pressing Return/Enter in the Search dialog is the same as clicking the Search button in the Search dialog. `Cmd+f` is the shortcut to bring up the Search dialog, so is a different action. 

On Windows, "Find Next" is typically the `F3` function button, and `Shift+F3` for "Find Previous". I'll see if I can implement those two, with `Cmd+g` and `Cmd+Shift+g` for Macs.

> > 5. Options Reverse, Case Insensitive, Wrap, Regex should be self-explanatory.
> 
> I thought I knew what `wrap` meant, but in order to have the search result show a selection that started at the end of one line and ended at the beginning of the next, I had to use a regex search and use the form: `first\slast`. But maybe I don't know how to use `wrap`. 😁

It means wrap round the end/start of the file and keep searching, instead of stopping at the end/start of file 😁. I notice that NP++ uses "Wrap around" for that feature, which would be a bit clearer, so I'll do that.

> > 8. Find All finds all occurrences of the string in the file/selection, and displays them in a "checker" dialog.
> 
> The simple search highlight's the search term as it is navigated to (possibly an OS function, as it seems to use the OS default highlighting?). Find all does not. It navigates to the search term, and places the cursor at the beginning of the search term, but unless you have a really painfully obvious cursor, it may be difficult to see. Would it be possible to add this?

The "default highlighting" (selection) of the term during simple search is something I coded, which matches NP++ behavior, but it would be easy to use a different type of highlighting. An advantage of selecting the match (rather than just coloring it in another color) is that the user can immediately do something to the match, like copy/cut/delete it, or type to replace it, or (when implemented) change case, or other similar feature.

Your suggestion about highlighting the relevant word when you click in the Find All and it jumps to the line in the main text is definitely how it will work. However, at the moment, I'm considering that to be a feature of the "checker" dialog, rather than a feature of "Search", so I haven't implemented it yet. 

You might also notice that (unlike GG1) it blindly jumps to the line/column number quoted at the start of the line in the checker dialog. If you keep the dialog up, and insert/delete some lines from the file, it will jump to the "wrong" line/column. Unlike Big Jule (if you're a Guys & Dolls fan) it doesn't "remember where the spots formerly was". 😁 Again, the plan is to fix that during work on the checker dialog, not Search.

windymilla:
@srjfoo - I think all your points (that I intend to do now) have been addressed. Plus, I realised Whole Word was missing, so I've added it.

src/guiguts/widgets.py
```diff
@@ -96,3 +134,32 @@ def grab_focus(
...
+    dlg_cls: type[TlDlg], root: tk.Tk, title: Optional[str] = None
+) -> TlDlg:
+    """Show the given dialog, or create it if it doesn't exist.
+
+    Args:
+        dlg_cls: Class of dialog to be created - subclass of ToplevelDialog.
+        root: Tk root.
+        title: Dialog title.
+    """
+    global _toplevel_dialogs
```
cpeel, line 156: Rather then make this global, make it the python-equiv of a static variable. [The method here](https://stackoverflow.com/a/279597) with the `hasattr` check is the one I use so its entirely self-contained within the function.

windymilla: Thanks @cpeel. That is what I did previously for a similar circumstance, and I tried it again when developing this. I've just had another go after your review. 
The problem is that mypy seems to really struggle with function attributes. You get an error on every line the attribute is used, for example
`src\guiguts\widgets.py:156: error: "Callable[[type[TlDlg], Tk, str | None], TlDlg]" has no attribute "_toplevel_dialogs"  [attr-defined]`
Of course, you can add a suppression comment on each line to get rid of those errors, but it makes it messy. 
Do you think it's OK to keep the global with "_" to flag it as "private", or is there a different technique, or do you think I have misunderstood either the technique or mypy?

Here's the code that gives mypy errors - seven of them:
```
TlDlg = TypeVar("TlDlg", bound=ToplevelDialog)


def show_toplevel_dialog(
    dlg_cls: type[TlDlg], root: tk.Tk, title: Optional[str] = None
) -> TlDlg:
    """Show the given dialog, or create it if it doesn't exist.

    Args:
        dlg_cls: Class of dialog to be created - subclass of ToplevelDialog.
        root: Tk root.
        title: Dialog title.
    """
    if not hasattr(show_toplevel_dialog, "_toplevel_dialogs"):
        show_toplevel_dialog._toplevel_dialogs = {}
    dlg_name = dlg_cls.__name__
    if (
        dlg_name in show_toplevel_dialog._toplevel_dialogs
        and show_toplevel_dialog._toplevel_dialogs[dlg_name].winfo_exists()
    ):
        show_toplevel_dialog._toplevel_dialogs[dlg_name].deiconify()
    else:
        if title is not None:
            show_toplevel_dialog._toplevel_dialogs[dlg_name] = dlg_cls(root, title)  # type: ignore[call-arg]
        else:
            show_toplevel_dialog._toplevel_dialogs[dlg_name] = dlg_cls(root)  # type: ignore[call-arg]
    return show_toplevel_dialog._toplevel_dialogs[dlg_name]  # type: ignore[return-value]
```

src/guiguts/widgets.py
```diff
@@ -96,3 +134,32 @@ def grab_focus(
...
+
+
+# Dictionary of ToplevelDialog objects, keyed by class name.
+# Used to ensure only one instance of any dialog is created.
+_toplevel_dialogs: dict[str, ToplevelDialog] = {}
+
+TlDlg = TypeVar("TlDlg", bound=ToplevelDialog)
+
+
+def show_toplevel_dialog(
```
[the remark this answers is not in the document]
windymilla: Thanks - good to know I wasn't too far off a sensible solution

src/guiguts/widgets.py
windymilla: I looked to see if you could do the equivalent of `# fmt: off` and `# fmt: on` in black to disable/re-enable that specific error for those lines. It doesn't appear possible at the moment - you have to disable the error for the whole module, or you can turn off all checking in a function, neither of which are very satisfactory. [This issue](https://github.com/python/mypy/issues/9483) is requesting the same feature.

Do you think it would be best if I just put the 7 suppression comments in, and not worry about it any more? It's just annoying to have to do that whenever I want a "static", but perhaps that's the price of using mypy?

windymilla: I pushed another commit to show what the code is like with the comments. Happy to go with whichever you think is best.

srjfoo:
> > > 2. If some text is selected in the main window when the dialog is created, the search box will be prepopulated with that text (only up to the first newline in a multiline selection)
> > 
> > 
> > Yes, but only if the previous search dialog has been closed. Will not replace what's in the search field already with the new selection.
> 
> Thanks - I agree it would be more helpful if it can do that.

Works splendidly!

> > > 4. Pressing the Return/Enter key is the same as clicking the Search button.
> > 
> > Pressing Return/Enter does work for finding next/previous, but repeated usage of `Cmd`+`f` does nothing except flash the menu (equivalent to starting a new search?). (I think that the "search again" shortcut that I'm most familiar with is `Cmd`+`g`, but I know that's not true for all apps.)
> 
> That is as expected - my wording was a little imprecise. Pressing Return/Enter in the Search dialog is the same as clicking the Search button in the Search dialog. `Cmd+f` is the shortcut to bring up the Search dialog, so is a different action.
> 
> On Windows, "Find Next" is typically the `F3` function button, and `Shift+F3` for "Find Previous". I'll see if I can implement those two, with `Cmd+g` and `Cmd+Shift+g` for Macs.

Nice, thank you!

> > > 5. Options Reverse, Case Insensitive, Wrap, Regex should be self-explanatory.
> > 
> > I thought I knew what `wrap` meant, but in order to have the search result show a selection that started at the end of one line and ended at the beginning of the next, I had to use a regex search and use the form: `first\slast`. But maybe I don't know how to use `wrap`. 😁
> 
> It means wrap round the end/start of the file and keep searching, instead of stopping at the end/start of file 😁. I notice that NP++ uses "Wrap around" for that feature, which would be a bit clearer, so I'll do that.

<cough> I notice that line ls getting kind of crowded in the dialog. Should it be "wrapped"? 😁 

> > > 8. Find All finds all occurrences of the string in the file/selection, and displays them in a "checker" dialog.
> > 
> > The simple search highlight's the search term as it is navigated to (possibly an OS function, as it seems to use the OS default highlighting?). Find all does not. It navigates to the search term, and places the cursor at the beginning of the search term, but unless you have a really painfully obvious cursor, it may be difficult to see. Would it be possible to add this?
> 
> The "default highlighting" (selection) of the term during simple search is something I coded, which matches NP++ behavior, but it would be easy to use a different type of highlighting. An advantage of selecting the match (rather than just coloring it in another color) is that the user can immediately do something to the match, like copy/cut/delete it, or type to replace it, or (when implemented) change case, or other similar feature.
> 
> Your suggestion about highlighting the relevant word when you click in the Find All and it jumps to the line in the main text is definitely how it will work. However, at the moment, I'm considering that to be a feature of the "checker" dialog, rather than a feature of "Search", so I haven't implemented it yet.
> 
> You might also notice that (unlike GG1) it blindly jumps to the line/column number quoted at the start of the line in the checker dialog. If you keep the dialog up, and insert/delete some lines from the file, it will jump to the "wrong" line/column. Unlike Big Jule (if you're a Guys & Dolls fan) it doesn't "remember where the spots formerly was". 😁 Again, the plan is to fix that during work on the checker dialog, not Search.

Sadly (considering the number of Broadway musicals' soundtracks I was exposed to when I was growing up) Guys and Dolls isn't one I'm familiar with.

I'm fine with the default highlighting. Note that (at least on macOS) the focus stays with the dialog, so I have to switch windows before I can type a replacement in place. Once I've switched focus to the active text window, replacing the selection works just fine. `Cmd`+`g` works regardless of whether or not the main text window has focus.

I guess I'm used to seeing the highlighting for Find All in the search results in BBEdit, so that's what I associate it with. Sounds like you have a plan for it, though, so that's good.

As for the disjoint line numbers -- I've seen that in BBEdit if I remove or add lines going from front to back, so if I know I'm doing that, I try to go back to front -- or just redo the search. So, I probably wouldn't even have noticed it if you hadn't made a point of mentioning it.

srjfoo (approved):

src/guiguts/widgets.py
cpeel: I really hate it when tooling makes you jump through hoops and just make your life harder.

What about if we made the function a class method on TopleveDialog itself? Make `_toplevel_dialogs = {}` a class attribute and then the function becomes something like:
```python
@classmethod
def show_dialog(cls, dlg_cls: type[TlDlg], title: Optional[str] = None) -> TlDlg:
    """Show the given dialog, or create it if it doesn't exist.
    Args:
        dlg_cls: Class of dialog to be created - subclass of ToplevelDialog.
        title: Dialog title.
    """
    # root should be available in a higher-level parent, right?
    dlg_name = dlg_cls.__name__
    if (
        dlg_name in _toplevel_dialogs
        and _toplevel_dialogs[dlg_name].winfo_exists()
    ):
        _toplevel_dialogs[dlg_name].deiconify()
    else:
        if title is not None:
            _toplevel_dialogs[dlg_name] = dlg_cls(root, title)  # type: ignore[call-arg]
        else:
            _toplevel_dialogs[dlg_name] = dlg_cls(root)  # type: ignore[call-arg]
    return _toplevel_dialogs[dlg_name]  # type: ignore[return-value]
```

Which becomes effectively a factory method.

windymilla: Thanks - I've had a bit of a look at that, and can have a closer look tomorrow.
I also experimented with keeping it as a function and function attribute, but using try/except, which resulted in fewer uses of the attribute, so fewer suppression comments (though I could probably do similar with the `hasattr` version by using a local `dlg` variable like I did in this version). 

Please could you let me know which version you prefer: the hasattr (maybe with local variable), the class method, or the try/except version below. I have a slight inclination towards the hasattr version, but would be OK with any of them tbh. Really appreciate all the advice, and learning different ways to do things - thanks.

```python
def show_toplevel_dialog(
    dlg_cls: type[TlDlg], root: tk.Tk, title: Optional[str] = None
) -> TlDlg:
    """Show the given dialog, or create it if it doesn't exist.

    Args:
        dlg_cls: Class of dialog to be created - subclass of ToplevelDialog.
        root: Tk root.
        title: Dialog title.

    Attributes:
        _toplevel_dialogs: Dictionary of ToplevelDialog objects, keyed by class name.
            Used to ensure only one instance of any dialog is created.
    """
    dlg_name = dlg_cls.__name__
    try:
        dlg = show_toplevel_dialog._toplevel_dialogs[dlg_name]  # type: ignore[attr-defined]
    except (AttributeError, KeyError) as exc:
        # Dialog not found in dict
        if isinstance(exc, AttributeError):
            # First time - create empty dict
            show_toplevel_dialog._toplevel_dialogs = {}  # type: ignore[attr-defined]
        dlg = None
    if dlg and dlg.winfo_exists():
        # Dialog already exists, so show it
        dlg.deiconify()
    else:
        # Dialog doesn't currently exist, so create it
        dlg = dlg_cls(root) if title is None else dlg_cls(root, title)  # type: ignore[call-arg]
        show_toplevel_dialog._toplevel_dialogs[dlg_name] = dlg  # type: ignore[attr-defined]
    return dlg
```

cpeel: > Please could you let me know which version you prefer: the hasattr (maybe with local variable), the class method, or the try/except version below. I have a slight inclination towards the hasattr version

The `hasattr` version is great with me and if that's the one you prefer I say go for it! I would prefer a non-global method but am largely ambivalent about the other options.

windymilla: @cpeel - many thanks. After looking in more detail at your class method solution, I think that makes most sense. In [these docs](https://docs.python.org/3/howto/descriptor.html#class-methods), it says "One use for class methods is to create alternate class constructors. For example, the classmethod [dict.fromkeys()](https://docs.python.org/3/library/stdtypes.html#dict.fromkeys) creates a new dictionary from a list of keys."
That seems similar to what we're doing here.
So, I've pushed a commit more or less with your code plus a couple of type declaration edits.

Regarding "root", TIL you don't need to give root as an argument when creating a Toplevel widget, which means I've been able to remove use of it from several places. So, I've pushed another commit for that. Thanks for prompting me to investigate that.

Hopefully ready for a final check of the code by you now, thanks.

Also, @srjfoo, if you could do one more quick check that Search is basically working, especially that you can't get two search dialogs up at once, that would be great, thank you.

cpeel: Oh yeah, the class method looks super clean. 👍

cpeel (approved):

PR #128, opened 2024-02-11
Trap attempt to load beyond list of recent files

### LINKED ISSUE #127: Using `-r1` on command line crashes if no prefs file
Tries to load most recent file, but there is no list of recent files.

windymilla:
Trying to load the 3 most recent file when there are only 2 files in the list in the prefs file, or similar attempts caused traceback.

Fixes #127

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -112,7 +112,10 @@ class Guiguts:
             self.file.load_file(self.args.filename)
         elif self.args.recent:
             index = self.args.recent - 1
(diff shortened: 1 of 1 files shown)


cpeel (approved):

srjfoo:
Just to make sure -- the correct response is to open an empty file with no error message.

windymilla:
I think that's good enough - extremely unlikely to happen to a "normal" user, ie. one who doesn't type "guiguts -r1" and who doesn't edit or delete their GGprefs file :)

PR #129, opened 2024-02-13
Add Mac shortcuts for beginning/end of file

windymilla:
1. Cmd with up/down arrow goes to beginning/end of file
2. Shift+Cmd+Up/Down selects to beginning/end of file
3. New `bind_event` method to avoid having many routines needing to return "break" if they are bound to events
4. By default on Windows word characters were any non-whitespace characters, and on Unix alphanumerics and underscores were word characters. Changed on both platforms to be alphanumerics plus apostrophes (straight & curly).

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -109,8 +109,8 @@ class MainText(tk.Text):
         # events that might change which line numbers should be displayed
         self.linenumbers = TextLineNumbers(self.frame, self)
         self.linenumbers.grid(column=0, row=0, sticky="NSEW")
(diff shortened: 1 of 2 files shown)


cpeel (approved):

srjfoo:
> 4\. By default on Windows word characters were any non-whitespace characters, and on Unix alphanumerics and underscores were word characters. Changed on both platforms to be alphanumerics plus apostrophes (straight & curly).

What about hyphenated words -- or is that an ambiguous situation that PPers are expected to handle themselves? (I agree that it's good to have regexes function the same on all platforms.

srjfoo (approved):

PR #132, opened 2024-02-15
Store geometry of root and toplevel dialogs

windymilla:
Currently implemented for the main (root) window and for dialogs inheriting from ToplevelDialog.

Width & Height and position on screen are stored in the GGprefs file. If a dialog is set not to be resizable, then the width and height are ignored when loaded from the GGprefs file, and the dialog is created at its default size. Currently there are no non-resizable Toplevel dialogs.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -42,12 +42,14 @@ class Root(tk.Tk):
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
-        self.geometry("800x400")
+        self.geometry(preferences.get("RootGeometry"))
         self.option_add("*tearOff", False)
         self.rowconfigure(0, weight=1)
         self.columnconfigure(0, weight=1)
         self.after_idle(lambda: grab_focus(self, maintext(), True))
         self.set_tcl_word_characters()
+        self.save_config = False
+        self.bind("<Configure>", self._handle_config)
 
     def report_callback_exception(
         self, exc: type[BaseException], val: BaseException, tb: TracebackType | None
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: although there are no non-resizable dialogs yet, you can temporarily make the Search dialog non-resizable to see the behavior. Add `resizable=False` as a parameter to line 52 of search.py:
```python
        super().__init__("Search & Replace", resizable = False, *args, **kwargs)
```

src/guiguts/mainwindow.py
```diff
@@ -72,6 +74,24 @@ def set_tcl_word_characters(self) -> None:
         self.tk.call("set", "tcl_wordchars", r"[[:alnum:]'’]")
         self.tk.call("set", "tcl_nonwordchars", r"[^[:alnum:]'’]")
 
+    def _handle_config(self, event: tk.Event) -> None:
+        """Callback from root dialog <Configure> event.
+
+        By setting flag now, and queuing calls to _save_config,
+        we ensure the flag will be true for the first call to
+        _save_config when pricess becomes idle."""
```
srjfoo, line 82: `pricess` -\> `process`?

src/guiguts/widgets.py
```diff
@@ -107,6 +116,39 @@ def show_dialog(cls, dlg_cls: type[TlDlg], title: Optional[str] = None) -> TlDlg
...
+            self.geometry(geometry)
+        except KeyError:
+            pass  # OK if no stored geometry for this dialog
+
+    def _handle_config(self, event: tk.Event) -> None:
+        """Callback from dialog <Configure> event.
+
+        By setting flag now, and queuing calls to _save_config,
+        we ensure the flag will be true for the first call to
+        _save_config when pricess becomes idle."""
```
srjfoo, line 135: `pricess` -\> `process`?

srjfoo (approved):
Seems to work as advertised. Just curious, though -- what sorts of dialogs should *not* be resizable?

windymilla:
> Seems to work as advertised. Just curious, though -- what sorts of dialogs should _not_ be resizable?

A dialog that just has some buttons, for example, or some small text fields that don't need to be made larger.
In GG1, about half the dialogs just have position stored, and half have position & size (see [lines 871-928 in Utilities.pm](https://github.com/DistributedProofreaders/guiguts/blob/master/src/lib/Guiguts/Utilities.pm#L871-L928) for a list where `geometryhash` stores size & position and `positionhash` just stores position) In fact the user can physically resize some of those dialogs, but they just end up with some whitespace around the buttons, and next time they pop the dialog it will be back to its default size. Try Tools->Fixup Page Separators in GG1 to see what I mean.

src/guiguts/widgets.py
```diff
@@ -107,6 +116,39 @@ def show_dialog(cls, dlg_cls: type[TlDlg], title: Optional[str] = None) -> TlDlg
...
 
+    def _do_config(self) -> None:
+        config_dict = preferences.get("DialogGeometry")
+        try:
+            geometry = config_dict[self.__class__.__name__]
+            x_resize, y_resize = self.resizable()
+            if not (x_resize or y_resize):
+                geometry = re.sub(r"^\d+x\d+", "", geometry)
+            self.geometry(geometry)
+        except KeyError:
```
cpeel, line 127: I think it's generally better to handle known conditions directly via a conditional than with an exception -- or to have the exception narrowly focused -- to prevent something _else_ in the try block raising the exception making hard-to-track-down bugs.

cpeel (approved):

src/guiguts/widgets.py
windymilla: Thanks @cpeel - is this better? Still trying to learn the balance/benefits between EAFP & LBYL styles.

PR #133, opened 2024-02-16
Add case-change features

### LINKED ISSUE #35: Add case-change features
As in GG1's edit menu:
- lowercase selection
- UPPERCASE selection
- Sentence case selection
- Title Case selection

tangledhelix:
Implements the following in the Edit menu:

* UPPERCASE SELECTION
* lowercase selection
* Sentence case selection
* Title Case Selection

If no text is selected, the commands / hotkeys should be ignored. Hotkeys were taken from GG1.

Tested on macOS. Testing on Windows / Linux would be appreciated. See comments in the code for some notes on particular issues.

1. Part of the title-case code should only happen if the text is English. GG2 has no way to determine the language at the moment, so far as I can tell, so we should revisit it later and add an English check to this code.
2. Potentially the sentence-case behavior could be improved. However, the current behavior is directly taken from GG1, so hopefully it isn't a problem for the users.
3. Some of the (Win/Linux) menu selection letters from GG1 are already in use, so I used alternatives, which are noted.

I'm completely open to changing things like the hotkeys, the menu letter-selectors (e.g. `~m`) in Win/Linux menus, if anyone wants to suggest different choices.

Fixes #35


src/guiguts/application.py
```diff
@@ -353,6 +353,33 @@ def init_edit_menu(self) -> None:
...
+            "",
+        )
+        menu_edit.add_button(
+            "~Title Case Selection",
+            lambda *args: maintext().transform_selection(
+                maintext().title_case_transformer
+            ),
+            "Ctrl+t",
+        )
+        # NOTE: this was "~u" in GG1; but "~u" is now used by Undo, using "~p"
```
windymilla, line 377: I haven't particularly been trying to preserve the GG1 "~" letters, so I'm fine with this.

src/guiguts/maintext.py
```diff
@@ -703,6 +703,102 @@ def find_matches(
...
+
+        Args:
+            fn: Reference to a function or method
+        """
+        sel = self.selected_ranges()
+        if not sel:
+            # nothing is currently selected
+            return
+
+        p1 = self.index("sel.first")
```
windymilla, line 717: At the moment, this doesn't cope well if the user does a "column" (or block) selection (`Option+drag mouse` on Mac).
For each feature we will need to decide whether (or when) to implement support for column selections. 
Internally, column selections consist of multiple selections (normally between column M and N on multiple rows to form a rectangle). If we decide not to support column selections for a feature, then the best  option is probably to just work on the first selection, i.e. the first line of the block. That is what I have done for the "Count" feature in the Search dialog for now (though I would envisage adding column selection support at some point). If we do want to support column selection (like GG1 does for case-changing) then we need to loop on the list of selections that `selected_ranges()` returns. An example of this is in `column_copy_cut`.

src/guiguts/application.py
```diff
@@ -353,6 +353,33 @@ def init_edit_menu(self) -> None:
         menu_edit.add_button(
             "Colu~mn Paste", maintext().columnize_paste, "Cmd/Ctrl+Shift+V"
         )
+        menu_edit.add_separator()
+        # NOTE: this was "~l" in GG1; but "~l" is now used by Column Cut, using "~w"
+        menu_edit.add_button(
```
windymilla, line 358: To avoid large menus, we could consider putting these in a submenu. I'm not sure of the balance between large menus and deep menus, so happy to go with whatever you think.

src/guiguts/maintext.py
```diff
@@ -703,6 +703,102 @@ def find_matches(
...
+            "to",
+        )
+
+        def capitalize_first_letter(match):
+            word = match.group()
+
+            # TODO: At the time this method was implemented, GG2 was not aware
+            # of the document language (`::main_lang()` in GG1 terms). When
+            # such support appears in GG2, we should add a test so that
+            # `exception_words` is only checked when the language is `en`.
```
windymilla, line 788: Good point, thanks! I've added [an issue](https://github.com/DistributedProofreaders/guiguts-py/issues/134) to support a document-specific language

windymilla (changes requested):
Looks good on Windows too (apart from column selection as mentioned in detailed comments)
Thanks!

cpeel (commented):
Code LGTM (sans Nigel's comments about column selections) - thanks for implementing this!

src/guiguts/application.py
tangledhelix: I realized as I was falling asleep last night that I left this counter to my own hotkey advice, so I'll look at this. Maybe I'll do the platform one... or maybe this should grow a shift key or something...

src/guiguts/application.py
tangledhelix: I briefly wondered about this myself, and the answer might depend on how closely we are planning to hew to the shape of GG1. I placed these in the same place I found them in GG1, but maybe we plan to make a new shape? Have you thought through that aspect already?

src/guiguts/maintext.py
tangledhelix: Aha, so that's why it's possible to return multiple ranges... I'm not accustomed to using column selection so that was off my radar. I'll revisit this and see if I can loop over a set of them.

src/guiguts/application.py
tangledhelix: > I haven't particularly been trying to preserve the GG1 "~" letters, so I'm fine with this.

Just to clarify, what I read this to mean is: you are fine with the `~` letter changing, but each menu item still ought to *have* one. Is that right?

By the way, I have not been testing those, and I don't think I can. The menu behavior is different in macOS. When a menu is open, you can press letters to jump to items, but it's based on the text label, not on the `~` letters. Example: in the current Edit menu, I can press `c o l` and it will jump as I type to `Cut` (c), then `Copy` (co), and finally `Column Cut` (col). In GG1, I could use the `~` letter because the menu handling was done by XQuartz, not macOS. This is all fine and normal, it's just information for you in case you weren't already aware.

Also for developer notes: if two menu items have the same `~` letter, Python doesn't complain about it (at least, on macOS it doesn't - maybe it does on other platforms?)

tangledhelix:
Since I don't seem to be done with this, I tried to convert to draft, but when I press the button Github just smiles politely yet ignores me... I probably don't have enough permission to do it. If marking as draft is normal for this project and someone wants to press the button for me, go ahead and do that, apparently I can't.

tangledhelix:
Never mind... it did so, it just didn't update until I hard-refreshed the page... 🙄 

> Since I don't seem to be done with this, I tried to convert to draft, but when I press the button Github just smiles politely yet ignores me... I probably don't have enough permission to do it. If marking as draft is normal for this project and someone wants to press the button for me, go ahead and do that, apparently I can't.

src/guiguts/application.py
windymilla: Sorry - on re-reading it, you're right, I wasn't clear. I am fine with the letter changing, but yes, let's keep a `~` letter for every menu item. If at any point in the future we decide not try not supporting them, it should be a simple fix in `process_label` to 
`    return (-1, label.replace("~", ""))`

I didn't know (or remember?) that they weren't supported on macOS - thanks for that info. As you described, I see in [TkDocs](https://tkdocs.com/tutorial/menus.html#:~:text=Underline) it says (my emphasis) 

> All platforms support keyboard traversal of the menubar via the arrow keys. On **_Windows and X11_**, you can also use other keys to jump to particular menus or menu items. The keys that trigger these jumps are indicated by an underlined letter in the menu item's label. 

One thing that may cause complaints from GG1 users is that (on the basis of stuff I read about modernising a Tk interface) I have disabled the "tear-off" functionality for menus in GG2, though it's entirely possible we could have that as a user preference - I confess I used it a lot in GG1. So, without tear-offs, using `Alt+E`, then `T` provides quick keyboard access to Title Case, for example, even if we don't have the Ctrl+t shortcut. On Macs, it sounds like a few arrow keys in addition would be needed. 

On Windows, if two menu items have the same `~` letter, the first one in the menu is selected by the keystroke.

src/guiguts/maintext.py
windymilla: I only became accustomed to it because of GG1 development, though I use it quite a lot in Notepad++ when editing code.

src/guiguts/application.py
windymilla: That's a really tricky question, and one I'm trying to put off a little longer, as far as it's possible to continue without making lots of extra work, either to match GG1 now, or the change to match in future. I'd imagined that once there were a few features in GG2, but well before it's actually a fully useful tool, we'd make a version available and engage existing PPers in a discussion about it. Closeness to GG1 would likely be a significant part of that. That said, I'm very happy to hear comments/suggestions about it at any point.

windymilla:
> ... If marking as draft is normal for this project ...

Happy either way - "Draft" does make it clearer to reviewers, so is probably better. An "I'll look into fixing this" comment is fine with me, though I've been guilty in the past of not making it clear when I've finished looking into it, so reviewers could be unsure when to retest.

src/guiguts/application.py
tangledhelix: @windymilla I'm curious if, in macOS, the tear-off menus are even possible. You say that you disabled the feature--can you point me to where that was? I'd like to try enabling it, to see what happens.

tangledhelix: > Are you happy that this will make `Ctrl+t` do Title Case on Macs? I know you were concerned previously about not overwriting shortcut Control keys unnecessarily.

I removed (for Mac only) the hotkeys for these menu items. We can always adjust later, but (as discussed in Slack) I'm thinking in the direction of hotkeys only for the more commonly used commands.

windymilla: AFK, but I think in Root class init constructor in Mainwindow. I think it is called tearoff or tear-off

windymilla: Agreed, let's keep to a minimum for now, then add them deliberately for common commands, maybe after some consultation with users. I'm happy for you to remove for other users too.

windymilla: Back at computer, @tangledhelix, the tearOff setting is in the Root constructor (about line 46)
```python
        self.option_add("*tearOff", False)
```

src/guiguts/application.py
tangledhelix: I support the "put it off" approach. If we mimic the GG1 menus as we add things, we don't have to think about it much. We can overhaul the menu structure later on when the app has a more mature shape. Restructuring the menus seems easy from the code I've seen so far...

src/guiguts/application.py
tangledhelix: @windymilla As I suspected, tearOff does nothing on macOS. Thanks for the pointer.

tangledhelix: > I'm happy for you to remove for other users too.

Removed entirely (not only for Mac)

src/guiguts/maintext.py
tangledhelix: I also changed from the prior behavior and not re-selecting the text after modification. First, because it wasn't working. And second because GG1 doesn't do that.

windymilla: > Added support for column selections. @windymilla, if you wouldn't mind a quick re-test...

Seems to work well, thanks.

I noticed that you had duplicated my code from `column_copy_cut` to handle the special "case where line was too short to select any text", and had two thoughts:
1. We probably don't want to have to do that everywhere we use column selections
2. In this specific case it doesn't matter since changing the case of a newline character does nothing anyway.

So I've added a separate issue #139 which one of us can pick up after this PR is merged. For now, you can either remove or leave the special case code in `transform_selection`.

tangledhelix:
Removed the draft flag, rebased/squashed.

src/guiguts/maintext.py
```diff
@@ -739,6 +739,99 @@ def find_matches(
...
+
+        The transformation is not aware of sentence structure; if the
+        input string consists of multiple sentences, the result will
+        likely not be what was desired. This behavior was ported as-is
+        from Guiguts 1.x, but could potentially be improved, at least
+        for languages structured like English.
+
+        To be more specific: if multiple sentences are selected, the
+        first character of the first sentence will be capitalized, and
+        the subsequent sentences will begin with a lowercase letter.
```
srjfoo, line 774: If `Sentence case` is applied to a columnar selection, the first letter of each line of the selection is capitalized. -- I have not yet verified that this same behavior is present in GG1 but am noting it (mind you, I don't know _**why**_ someone would want to apply sentence case to a columnar selection, but I test some weird cases just because). Anyway, should that behavior be noted in a comment?

Confirmed, as I expected, that the caveat about sentence case in a selection also applies to proper names (I'd have been surprised if it hadn't. :D 

Seriously, should Sentence case come with a warning that you may get unexpected results (the same is true of Title Case), so it's always a good idea to check the transformation to make sure. The current GG manual does not.

srjfoo (approved):
Looks good to me, too. My other comments are strictly things that occurred to me while testing.

src/guiguts/maintext.py
tangledhelix: I added a comment (in the docstring). I'm not sure what else to do unless we want to note it in the manual? I don't think GG2 has a manual yet.

FWIW, I copied the behavior from GG1, which is why I was OK leaving it this way. If it hasn't been a problem so far, presumably it's OK?

srjfoo: I'm assuming it's okay. As far as I know, no GG2 manual yet. I'll leave it to Nigel to decide whether it needs to be documented or not, but I know that I've seen comments about the title case button in the proofreading interfaces, warning users who use it to check and make sure that it got everything correct.

windymilla: No, there's no GG2 manual yet ;)
I'm happy that it does what makes sense, if the user does a column selection. I don't think there is anything else they would expect it to do, and in a real example, rather than our tests, it would do what you want. For example, a table in a text file with sentences in the middle column. 
It could be argued that sentence case as it is in GG1 (and hence in GG2 atm) scarcely warrants taking up an entry in the menu, but let's keep it as-is for now. In the future it could either be enhanced, dropped or kept.

PR #141, opened 2024-02-18
Add Replace to Search dialog

windymilla:
1. Replace replaces string and finds the next occurrence.
2. Replace All replaces all occurrences in file, or just in selection if `In selection` is checked.
3. Tk Text widget doesn't support regex replacement, so matched string is "replaced" in the code, then inserted back in file.
4. It remains to be seen if there will be compatibility problems between Tcl regex and Python regex syntaxes.
5. Note current difference from GG1: to include a capture group in the replacement string, use `\1`, not `$1`
6. Another difference from GG1 is that a match at the start of a range (when using Replace All on selection), or a match immediately after the insert cursor will be found, but will not in GG1.
7. Column selection for "In selection" is not yet supported for Count, Find All or Replace All


srjfoo:
I'm not sure where to add this request, but since it's related to search and replace, I'll start here. I'd really like to see separate "replace (only)" and "replace & find next" operations. If you replace & find next, you'll likely be several pages further along, and you can't eye-ball the change to make sure it's correct. (Regardless of whether GG1 has it or not, I think it would be a good addition.)

windymilla:
Notes of things to do:
1. Shorten labels of a couple check buttons (Nocase & Wrap?)
2. Add padding between check buttons to keep the next checkbox away from the previous label
3. Add Replace (but don't search for next) button (like GG1 has)
4. Look at `\m`/`\M`/`\b` compatibility fix
5. Make dialog non-resizable
6. Test Posix character classes - expected to fail

windymilla:
Changes discussed with @srjfoo are now done.

@cpeel, I put a possibly controversial edit in as the [final commit](https://github.com/DistributedProofreaders/guiguts-py/pull/141/commits/ca935a49e71d3e085b94a70b3b0c74a6d5a6cf29). I wanted to add another class method to ToplevelDialog like `show_dialog` we discussed previously. As I did it, I wondered whether I could omit the class argument, and just use the class that was being used to call the method. So, in the final commit, I did it for the new method `get_dialog` as well as `show_dialog`. It does seem to work, but I didn't know if it was poor practise. If it's not acceptable, I can just revert that commit. Thanks

windymilla:
Minor padding adjustments will hopefully look OK on all platforms.
Reverse "Replace & Search" bug fixed
Search dialog options (like "Regex" and "Wrap around", but not "In selection") now stored in Prefs file.

cpeel (approved):
> @cpeel, I put a possibly controversial edit in as the final commit.

That change looks sound to me. Had to wrap my head around how that got inherited but it seems fine to me!

srjfoo (approved):
LGTM

PR #146, opened 2024-02-21
Facilitate use of external editor

windymilla:
Allow user to add "Page Marker Flags" (e.g. `[Pg001]`) before they save a file and edit it in a different editor. When file is loaded back into GG2, location of Page Marker Flags is used to update the page break locations that were loaded from the bin (.json) file.
Since the bin file and main text file will not match, soften the consequent error message to a warning if the above situation is detected.
The "modified" flag on the file is set if any page boundary locations are updated, because although the text file is not modified, it needs resaving to save the updated bin file.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -304,6 +304,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         self.menu_file.add_button(
             "~Close", self.close_file, "Cmd+W" if is_mac() else ""
         )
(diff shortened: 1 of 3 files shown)


cpeel (approved):

srjfoo:
I meant to get this tested tonight, but somehow I managed to corrupt my testing installation on the M1. It took me a while to think to check on the Intel to see if I was seeing the same problems before I concluded that I need to reinstall, so unfortunately, no further testing tonight.

srjfoo (approved):

PR #147, opened 2024-02-21
Move special case code to common routine

windymilla:
Instead of routines that use `selected_ranges()` needing to
correct for short lines, put the correction directly into
`selected_ranges`.


windymilla:
Also a couple type hints added.

Testing notes: Should have no noticeable effect, but if it did cause a problem anywhere, it would be doing an action on a column selection, e.g. column select then cut/copy/paste or change case.

cpeel (approved):

srjfoo (approved):

PR #148, opened 2024-02-22
Add ability to set language(s) for a book

windymilla:
1. Lang button added to status bar
2. User can click button and enter one or more languages,
e.g. "en" or "en+la" or "fr,de" or "en es cz"
3. Languages are stored in the projects bin file.
4. If a file doesn't have a bin file (and when GG first starts) the
language defaults to the most recently used language, which
is stored in the Prefs file.

Also make Title Case Selection language-dependent:
List of exception words (such as "a" and "the") should only be
used in English texts.

Also resolved two duplicate keyboard navigation letters in the
Edit menu.

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -94,6 +94,7 @@ class File:
     def languages(self) -> Any:
         """Languages of currently loaded file.
 
+        Multiple languages are separated by "+"
         When assigned to, executes callback function to update interface"""
         return self._languages
 
@@ -102,6 +103,8 @@ class File:
         self._languages = value
         self._languages_callback()
         preferences.set("DefaultLanguages", value)
+        # Inform maintext, so text manipulation algorithms there can check languages
+        maintext().set_languages(value)
 
     def reset(self) -> None:
         """Reset file internals to defaults, e.g. filename, page markers, etc"""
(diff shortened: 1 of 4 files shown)


src/guiguts/file.py
```diff
@@ -447,6 +475,21 @@ def get_current_page_label(self) -> str:
...
+        """Allow the user to set languages.
+
+        Multiple languages may be separated by any non-word character
+        which will be converted to "+"
+        """
+        languages = simpledialog.askstring(
+            "Set Language(s)", "Language(s)", parent=maintext()
+        )
+        if languages:
+            lang_list = re.split(r"\W", languages)
```
cpeel, line 488: This could do interesting things if someone inputs a locale (`en_US`) here instead of a language.

cpeel (approved):
I don't know if this level of detail is necessary, but this is the time to think about it if it is. DP uses locales, not languages, because while `pt_BR` and `pt_PT` may both be Portuguese but they're different enough that it matters sometimes.

The reason I bring this up is not that I think we should _require_ users to specify a locale, but if we want to _enable_ them to, we need to account for that here.

And of course this adds additional complexity because for many cases -- like the list of English words to exclude -- you probably don't care if it's `en_US` or `en_UK`.

So you might want to just punt on it, but it seemed worthwhile to bring it up here.

src/guiguts/file.py
windymilla: I think it's OK, because `\W` means not `\w` (alphanumeric plus underscore character). I did think about that when I used `\W` and thought that it would be OK for locales.

However, I didn't take that into account when I checked if English was one of the languages in the Title Case check, so I'll do something about that now.

srjfoo (approved):

tangledhelix:
Sorry I didn't get to this before it was merged; but since you requested a review, I ran it through the paces (especially the case change bits) and it all looks good to me, FWIW.

PR #149, opened 2024-02-22
Remove Prefs dialog

windymilla:
Early experiment with creating a dialog - unusable (in fact dangerous) in current form.

All relevant prefs are set by some other action. At some future point, there might be a prefs dialog with things like font changes, perhaps with tabs using `ttk.Notebook`, or maybe several separate dialogs, e.g. Appearance, Wrap settings, etc.


cpeel (approved):

srjfoo (approved):
Having just apparently been bitten by this one, I decided to bump it to the top of my reviews! 🤣

PR #151, opened 2024-02-24
Fix bugs in column select

### LINKED ISSUE #150: Improve column select
To avoid full lines of selection on short lines.

Either set "short" lines to select to line end without the current "correction" and "uncorrection" elsewhere
Or change the handling of "short" lines to omit them from the selection completely?

Also:

When index of mouse position ends up to the left of the starting click, when the mouse is actually to the right, the selection suddenly flips as you pass over that line. Do something like make the end column equal the column that the mouse's x position would have given on the longest line between start and current point?

Also, delete/backspace do not delete the column selections correctly

windymilla:
1. Stop selection "flipping" when the mouse passes over a short line.
2. Don't highlight whole line when column select includes blank line.
3. Delete & backspace now delete a column selection correctly (like cut, but doesn't add to clipboard).

FIxes #150

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -143,6 +143,8 @@ class MainText(tk.Text):
         self.bind_event("<<Cut>>", self.smart_cut, force_break=False)
         self.bind_event("<<Copy>>", self.smart_copy, force_break=False)
         self.bind_event("<<Paste>>", self.smart_paste, force_break=False)
+        self.bind_event("<BackSpace>", self.smart_delete, force_break=False)
+        self.bind_event("<Delete>", self.smart_delete, force_break=False)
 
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

src/guiguts/maintext.py
```diff
@@ -432,8 +434,8 @@ def do_column_select(self, col_range: IndexRange) -> None:
             # If line is too short for any text to be selected, select to
             # beginning of next line (just captures the newline). This
             # is then dealt with in column_copy_cut.
-            if self.get(beg, end) == "":
-                end += "+ 1l linestart"
+            # if self.get(beg, end) == "":
```
cpeel, line 437: Leftover debugging or should this be deleted?

windymilla: Deleted - thanks

PR #153, opened 2024-02-25
Enhance Checker Dialog

windymilla:
Improve `CheckerDialog.add_entry` method

1. Swap argument order and make the range indicating the area of interest in the main text an optional argument. This makes it possible to add general text to the dialog, e.g. header line.
2. If a range is given, automatically add the start of the range as a prefix to the line of text. Otherwise this code would probably be duplicated in every tool that wants to add lines to the dialog.
3. Stop using "reserved" word `range` as a variable name. 
5. Update Search dialog to use the new checker features, in particular a header and footer to the Search All results.

Add default header widgets to CheckerDialogs

1. Add label showing number of "errors"
2. Add "Re-run" button to re-run the same check
3. Use this for Search Dialog's Find All button.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -17,18 +17,18 @@ class CheckerEntry:
 
     Attributes:
         text: Single line of text to display in checker dialog.
-        range: Start and end of point of interest in main text widget.
+        text_range: Start and end of point of interest in main text widget.
     """
 
-    def __init__(self, text: str, range: IndexRange) -> None:
+    def __init__(self, text: str, text_range: Optional[IndexRange]) -> None:
         """Initialize CheckerEntry object.
 
         Args:
             text: Single line of text to display in checker dialog.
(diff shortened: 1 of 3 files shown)


cpeel (approved):

windymilla:
Testing notes:
1. Do Find All from Search Dialog. 
2. Check the label at the top (NN entries) has the correct number, e.g. if there are 5 occurrences of string, it says "5 entries"
3. Look at the header line at the top of the list of found strings - it should say what you searched for, with some extra bits if appropriate, like "whole word".
4. Check that re-run is equivalent to clicking Find All again. For example, delete one of the words that was a match, then click the Re-run button - it should find one fewer matches than last time. Note that you can also change settings, e.g. "Match case" and it will Re-run with the new settings. Although this doesn't seem that useful for Find All, it will be more so for tools like Spell Check that don't pop from another dialog, and where there will be more settings in the top of the checker dialog, e.g. threshold, verbose, etc.

PR #155, opened 2024-03-01
Implement basic spell checker

windymilla:
1. Doesn't handle project good words/bad words lists
2. Doesn't have threshold value
3. Can't add words to project/user dictionaries

Include English, French & German dictionaries.


windymilla:
Testing notes: should behave the same as GG1 Spell Query if you set the threshold very high in GG1 in the Spell Query dialog, because GG2 doesn't have the threshold feature yet.

In GG2, Spell Checker is in the Tools Menu. In GG1, Spell Query is in the Tools menu.

srjfoo (approved):

src/guiguts/spell.py
```diff
@@ -0,0 +1,325 @@
...
+import regex as re
+
+from guiguts.checkers import CheckerDialog
+from guiguts.maintext import maintext, FindMatch
+from guiguts.preferences import preferences
+from guiguts.utilities import IndexRowCol, IndexRange
+
+logger = logging.getLogger(__package__)
+
+DEFAULT_DICTIONARY_DIR = os.path.join(
```
cpeel, line 14: This may not work if GG is installed as a package. See https://stackoverflow.com/a/20885799 for a way that should work for both methods, specifically https://docs.python.org/3.11/library/importlib.resources.html

We should see if there is anything we need to do with the poetry config such that these files are included in the final package too.

windymilla: By the way, do you think I should be testing out `poetry publish` at this stage or can I put that off for now if `build` appears to be working?

src/guiguts/spell.py
```diff
@@ -0,0 +1,326 @@
...
+        """Add dictionary words for given language.
+
+        Attempts to load default and user dictionaries.
+        Outputs warning if default dictionary doesn't exits.
+        Raises DictionaryNotFoundError exception if neither dictionary exists.
+
+        Args:
+            lang: Language to be loaded.
+        """
+        filename = str(DEFAULT_DICTIONARY_DIR.joinpath(f"dict_{lang}_default.txt"))
```
cpeel, line 278: The casting as a string and then using a standard open forces this back into a filesystem, not package, access method. The `DEFAULT_DIRECTORY_DIR` should be a Traversable object that we can `.open()` directly. This way if the package is being read and used from within a `.tar.gz` this still works.

windymilla: Ahhhh. That makes sense. I did that because the routine I used to load the default dictionary (bundled with the system) and the user's dictionary (in their GGprefs dir) took a filename string. I'll adapt it - I can probably get a Traversable for the user's dictionary too if I have a look round.

windymilla: I found that `Traversable` is like the `Path` class, so am using the latter for the user's dictionary. Then the method that opens and loads the dictionary can take `Traversable|Path` as its input argument type, as they both support `.open()`

cpeel (approved):

src/guiguts/spell.py
```diff
@@ -0,0 +1,326 @@
+"""Search/Replace functionality"""
+
+import importlib.resources
+import logging
+import os
+import regex as re
+
+from guiguts.data import dictionaries
```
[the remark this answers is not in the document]
windymilla: Yes - that's how it was done in the example of using `importlib.resources`. When we come to actually test packaging/publishing Guiguts, we'll see if I've done it correctly :)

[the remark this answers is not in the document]
tangledhelix: Ok, thanks for the explanation. The reason I asked was looking forward to packaging. On Mac, and quite possibly on any platform requiring code signature, the app directory itself can't change (because it invalidates the signature). So when I saw "guiguts.data" I got a bit nervous and wanted to check what that data was. Static dictionaries would be fine. So long as later on we aren't storing anything user-modified there, like global dict or project dicts. But we have the user preferences directory for that now.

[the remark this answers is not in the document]
windymilla: Thanks for that reminder @tangledhelix. In fact it was your previous work on considering packaging for GG1 that alerted me to that situation, and led to the introduction of the GGprefs directory for user preferences and user "global" dictionaries in GG1. I'm planning to keep the same structure in GG2, at least initially, and everything "under the release" should be treated as read-only.

PR #157, opened 2024-03-04
Enhance the checker dialog features

windymilla:
1. Change underline to colored text in the dialog for highlighting the word(s) of interest
2. Highlight the chosen line in gray so user knows which one they clicked on
3. Add right-click to remove error from listing - can also remove other lines in the dialog (e.g. the header lines) because some (future) tools have headers partway through their output and a user  may want to get rid of the header when they have handled all the errors in that section.
4. Add Ctrl+left-click (Cmd+left-click on Macs) to "process" the clicked error if possible - see below
5. Add Ctrl+right-click to "process" then remove the error.

Some tools, like jeebies give errors where the correction that needs making is obvious, so if the user "processes" that error, it can make the correction for them, swapping he/be.

For testing only (I'll remove before merging) I've put in a dummy processing function for Spell Check - it just logs an info message with the range of interest and the error message. This is just to prove that it gets called with the appropriate mouse clicks. There will be no "process" function for Spell Check in fact.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -6,11 +6,12 @@ from typing import Any, Optional, Callable
 
 from guiguts.maintext import maintext
 from guiguts.mainwindow import ScrolledReadOnlyText
-from guiguts.utilities import IndexRowCol, IndexRange
+from guiguts.utilities import IndexRowCol, IndexRange, process_accel
 from guiguts.widgets import ToplevelDialog
 
 MARK_PREFIX = "chk"
 HILITE_TAG_NAME = "chk_hilite"
+SELECT_TAG_NAME = "chk_select"
 
 
 class CheckerEntry:
@@ -42,13 +43,19 @@ class CheckerDialog(ToplevelDialog):
     """
 
     def __init__(
-        self, title: str, rerun_command: Callable[[], None], *args: Any, **kwargs: Any
+        self,
+        title: str,
+        rerun_command: Callable[[], None],
+        process_command: Optional[Callable[[CheckerEntry], None]] = None,
+        *args: Any,
+        **kwargs: Any,
     ) -> None:
         """Initialize the dialog.
 
         Args:
             title: Title for dialog.
             rerun_command: Function to call to re-run the check.
+            process_command: Function to call to "process" the current error, e.g. swap he/be
         """
         super().__init__(title, *args, **kwargs)
         self.top_frame.rowconfigure(0, weight=0)
@@ -62,10 +69,22 @@ class CheckerDialog(ToplevelDialog):
         )
         self.rerun_button.grid(column=1, row=0, sticky="NSE", padx=20)
         self.top_frame.rowconfigure(1, weight=1)
-        self.text = ScrolledReadOnlyText(self.top_frame, wrap=tk.NONE)
+        self.text = ScrolledReadOnlyText(
+            self.top_frame, context_menu=False, wrap=tk.NONE
+        )
         self.text.grid(column=0, row=1, sticky="NSEW")
-        self.text.bind("<ButtonRelease-1>", self.jump_to_rowcol)
-        self.text.tag_configure(HILITE_TAG_NAME, underline=True)
+        self.text.bind("<ButtonRelease-1>", self.select_entry_by_click)
+        self.text.bind("<ButtonRelease-2>", self.remove_entry_by_click)
+        self.text.bind("<ButtonRelease-3>", self.remove_entry_by_click)
+        _, event = process_accel("Cmd/Ctrl+ButtonRelease-1")
+        self.text.bind(event, self.process_entry_by_click)
+        _, event = process_accel("Cmd/Ctrl+ButtonRelease-2")
+        self.text.bind(event, self.process_remove_entry_by_click)
+        _, event = process_accel("Cmd/Ctrl+ButtonRelease-3")
+        self.text.bind(event, self.process_remove_entry_by_click)
+        self.process_command = process_command
+        self.text.tag_configure(HILITE_TAG_NAME, foreground="blue")
+        self.text.tag_configure(SELECT_TAG_NAME, background="gray")
         self.reset()
 
     def reset(self) -> None:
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes: Please read commit message first.
If you Cmd+click a message, it should get printed to the console (if you've used the -d command line option) and also to the Message Log.
If it's a left click it will show you the relevant place in the main text. If it's a right click, it will remove the error, and show you the next error.

cpeel (approved):

srjfoo:
> Testing notes: Please read commit message first. If you Cmd+click a message, it should get printed to the console (if you've used the -d command line option) and also to the Message Log. If it's a left click it will show you the relevant place in the main text. If it's a right click, it will remove the error, and show you the next error.

Unless I've really misunderstood, right-click (via two-finger tap on a trackpad, or right mouse click on both a USB mouse and a USB trackball), with or without the command key, just brings up the "select all/cut/copy/paste" context menu. If that needs to have a special meaning in gg2 (at least for Mac users), I'm guessing there's a way to override it somewhere? (Sidenote: I got it to do something once with the `ctrl` key, but couldn't repeat it, and nothing was logged.)

`Cmd` plus left click seems to work fine and events are logged.

windymilla:
Summary of mouse button use:

- View error: `Button 1` all platforms
- Process error: `Command Button 1` on Macs, and `Control Button 1` elsewhere
- Remove error: `Button 2` or `Control Button 1` on Macs, and `Button 3` elsewhere
- Process & remove: `Command Button 2` or `Command Control Button 1` on Macs, and `Control Button 3` elsewhere

srjfoo:
> Summary of mouse button use:
> 
> * View error: `Button 1` all platforms
> 
> * Process error: `Command Button 1` on Macs, and `Control Button 1` elsewhere
> 
> * Remove error: `Button 2` or `Control Button 1` on Macs, and `Button 3` elsewhere
> 
> * Process & remove: `Command Button 2` or `Command Control Button 1` on Macs, and `Control Button 3` elsewhere

I've gotten some funky results testing this, but I think the issue is our old friend window focus.

* `Button 1` works. 
* `Command Button 1` works.
* `Control Button 1` works (yes, control-left-click is an alternate for right-click on Macs, because that was their answer to having a right-hand button until the mice with them proliferated). `Command Button 2` works IFF the checker dialog window has focus. (Works with the two-finger-tap-on-the-touchpad-equivalent-for-right-click, also)
* See above. The commands work as described, and the`Command Button 2` works only when the checker dialog window has focus.
* Footnote. The mouse I was playing with had a scrollwheel-button-3. It's apparently not wired in for Macs, and I got some weird kind of paste that I'm not sure where the copy comes from.

@tangledhelix, if you have the time, could you re-test and see what results you get. You probably have a better grasp of what's going on than I do.

tangledhelix:
The gray used for the selected line is too dark (IMHO), affecting readability. It's even a bit too dark I think with the black text, but particularly with the blue highlight text it makes it harder to read.

Screenshot 2024-03-06 at 2 02 03 AM

I propose instead either `#dddddd` or `#eeeeee`.

Screenshot 2024-03-06 at 2 09 42 AM

Screenshot 2024-03-06 at 2 10 01 AM

tangledhelix:
My test results:

Button 1:
* changes the currently selected line in spellcheck
* this happens correctly regardless of current window focus
* focus shifts to the main text window

Button 2:
* If the spellcheck window has focus already:
    * the line I clicked is removed
    * whatever line is automatically selected in its place is now highlighted in spellcheck
    * focus shifts to the main text window
* If the spellcheck window does not have focus already:
    * I get the OS-native "cut, copy, paste" context menu and nothing else happens

Cmd-Button 1:
* Prints to console the info about the selected line
* Selects that line in the spellcheck window
* Shifts focus to the main window
* This happens correctly whether the spellcheck window had focus or not

Cmd-Button 2:
* If the spellcheck window did not have focus, I get the OS-native "cut, copy, paste" context menu
* If the spellcheck window did have focus, the line info is printed to console, and the line is removed
* Just like with button 2, the focus shifts back to the main window.

So I think this is working with a couple of nits to figure out:

1. I suspect the reason the right-click doesn't propagate through when the window isn't focused is that the OS-native context menu is "in the way". Perhaps there's a way to have Tkinter not trigger that menu, and pass the click through regardless of focus.

2. The window changes focus away from spellcheck in a way I didn't expect. I suspect that's because when a line is removed from the list (in `remove_entry_by_click`), a different line is automatically selected in its place (the next line, or the one below it, I believe) using `select_entry`. And that calls `maintext().do_select` which probably shifts focus to the main window as a side effect of setting the insert point to a new location. I didn't deeply debug that, but reading the code and observing the behavior that's my best guess as to why the focus is shifting.

The focus shifts are also strange and ... while regular and predictable, they don't make sense. For example this sequence.

- Click in spellcheck window when it has focus
    - new line is selected in spellcheck
    - the maintext window gets focus
    - cursor in maintext jumps to a new position matching the new spellcheck selection
- Click again in spellcheck, now that it doesn't have focus
    - all the same things happen except now spellcheck ends up with focus at the end
- This cycle repeats, the focus just goes back and forth every 2nd click

It's nearly 3am so I can't puzzle through why it's doing that right now - but it's consistent behavior, and no matter which state the spellcheck window started in, it does choose the new line, everything works correctly, just the focus doesn't always end up in the same state, and just... toggles.

windymilla:
> My test results:

That full list of results is really helpful @tangledhelix and I agree with your summary "nits", though the alternating focus thing is just...weird.

Nit 2 is accurate - that is what the code does (i.e. forces focus to the main text window when a new error line is selected in the dialog, which also happens when a line is removed). I believe that's the same behavior as in GG1 - at least on Windows. For a Windows user, the behavior doesn't seem strange, and I think that's how the Search/Replace dialog works too (shifting focus when a search is done). It would be interesting to know if GG1 has the same issue, that on a Mac you have to restore focus to the dialog before you can right click to remove a message. In the latest version of GG2 code, right click (button2 or Ctrl+button1) is bound to the error removal function, and returns "break" to stop default behavior. From what you and Sharon are describing, that only seems to be effective when the window already has focus.

The current behavior may be acceptable for Windows, but weird for Mac users who are more used to focus staying where they put it, and having to click to explicitly change focus. I remember Sharon saying on GG1 she sometimes had to explicitly change focus to a dialog before being able to do something, so needing an extra click. I'll have a bit of a look around to see if I can find any info on focus issues & Python/tkinter, particularly to do with Macs.

windymilla:
Another option of course is to change the behavior from GG1, and say that focus remains with the dialog that is clicked, rather than forcing it to the main text widget. Or, as another alternative, just making that change on macOS.

windymilla:
The latest commit uses @tangledhelix 's suggestion for the "selected" background color, but after @srjfoo 's test in dark mode, it also sets the text color to black so it stands out against the light gray background.
I think I've removed the forcing of focus that was happening from the checker dialog, and also when doing **reverse** searching from the S/R dialog.
If this works to your satisfaction on macOS, I think we'll merge it, and worry at a later date about whether Windows users (or all users via a preference) get focus-forcing or not.

srjfoo (approved):

PR #158, opened 2024-03-07
Add pptxt tool

windymilla:
New version of pptxt, based on ppwb version, then enhanced and adapted to work with GG.


windymilla:
Hoping that this might be getting close to merging as a basic version of the tool - further work to improve its functionality and/or layout of messages could happen in future PRs

windymilla:
Added Quentin as a reviewer to make it easier for him to see any review messages.

srjfoo (approved):
One possible issue reported to @windymilla in Slack, but for the most part, everything seemed to work as expected.

src/guiguts/tools/pptxt.py
```diff
@@ -0,0 +1,2591 @@
...
+        for n in range(len(T)):
+            # For efficiency!
+            Tn = T[n]
+            Tn = Tn.replace("①", "-")
+            Tn = Tn.replace("②", "’")
+            Tn = Tn.replace("③", "‘")
+            Tn = Tn.replace("④", "'")
+            Tn = Tn.replace("⑤", ".")
+            Tn = Tn.replace("⑥", "^")
+            L.append(Tn)
```
windymilla, line 223: I may have misunderstood this little section, but would it have the same effect and possibly be more efficient and clear to omit the temporary list?
```
s.replace("①", "-")
s.replace("②", "’")
s.replace("③", "’")
s.replace("④", "’")
s.replace("⑤", "’")
s.replace("⑥", "^")
return s.split()
```

src/guiguts/tools/pptxt.py
```diff
@@ -0,0 +1,2591 @@
...
+        line = new_line
+
+        # Replace HTML entities (e.g. &amp;) with a space.
+        new_line = ""
+        while res := re.search(r"(&[a-z0-9]+?;)", line):
+            new_line += " " + line[0 : res.start(0)]
+            line = line[res.start(0) + len(res.group(0)) :]
+        if len(line) != 0:
+            new_line += " " + line
+        line = new_line
```
windymilla, line 102: If these sections are replacing occurrences of a regular expression with a space, could you use a call to re.sub instead of each block? Something like the following for the HTML entities block (untested)
```
re.sub(r"(&[a-z0-9]+?;)", " ", line)
```

src/guiguts/tools/pptxt.py
```diff
@@ -0,0 +1,2591 @@
...
+            scannos_list = s[i]
+            for j in range(len(scannos_list)):
+                scannos_misspelled.append(scannos_list[j])
+        s = {}
+
+        # Can add new entries above. Maintain the same code structure with no line longer
+        # than 88 characters and no dictionary index greater than 99. Note the 's = {}'
+        # line above is superfluous there but is REQUIRED if new dictionary entries are
+        # added after loop-code (see above) that appends dictionary entries to the list
+        # misspelled_scannos.
```
windymilla, line 2493: @G4OEU, could you explain the reasoning behind building the dictionary in this way via the small temporary dictionaries, to help decide if it's the best way to do that.

windymilla (commented):
Behavior of the tool looks good to me as a first robust version - a few queries on the code

src/guiguts/tools/pptxt.py
```diff
@@ -0,0 +1,2591 @@
...
+        if len(a_hy) > 0:
+            if first_header:
+                first_header = False
+            else:
+                checker_dialog.add_entry("")
+            checker_dialog.add_entry("Hyphen")
+            for record in a_hy:
+                line_number = record[0]
+                line = record[1]
+                report_all_occurrences_on_line(ch_hy, line, line_number)
```
windymilla, line 1797: For a line such as
`This contains--a double hyphen and a single-hyphen`
each hyphen in the double hyphen is reported (unexpected), whereas for
`This only contains--a double hyphen`
the hyphens in the double hyphen are not reported (as expected).
This seems to be independent of whether the earlier check said "Book seems to use '--' as em-dash so not reporting these further"

windymilla: Here are a couple of screenshots of the issue. There should be just one message about line 1335, rather than three, because the first two are part of a double-hyphen, as in this screenshot:
image
If I delete just the single hyphen from before "P2" on that line and re-run pptxt, then I get no messages at all about L1335, as in this screenshot:
image

So, it seems that if there is a single hyphen, then every hyphen on the line gets reported (including those in double hyphens), but if there is no single hyphen, then hyphens within double hyphens are not reported.

Here's a zip of the text file the screenshots came from:
[projectID60849c915cbd1.zip](https://github.com/DistributedProofreaders/guiguts-py/files/14551099/projectID60849c915cbd1.zip)

src/guiguts/tools/pptxt.py
G4OEU: This code extract is from the 'get_words_on_line' function. It parses the text of a line and returns a list of tokens that PPtxt considers as 'words' for the purposes of the various checks it performs. I rewrote this function immediately after PPtxt was added to the master GG. I did this to clarify and extend the range of tokens it provides, simplify the code that does it and improve its efficiency. As part of that effort the section above was coincidentally rewritten exactly in the form you suggest!

windymilla: Thanks @G4OEU - I know you have said there was additional tidying up of that nature to be done. It would be good to get a version of `pptxt` merged into `master`, so I'm happy with additional tidying going into a new PR at a later stage. I'll give @cpeel a chance to comment, but IMHO, there's nothing here that needs to stop us merging this, and allowing you to continue work in another branch.

src/guiguts/tools/pptxt.py
G4OEU: The use of small temporary dictionaries is no longer needed. I used them because I misunderstood a _flake8_ error message. I had originally arranged four scannos on each line so that they were aligned vertically into four neat columns. This meant multiple spaces after commas but _flake8_ objected to that with an "E241 multiple spaces after ','" message for each line but the last. I later discovered there is a way to tell _flake8_ to ignore that issue. However it is better to simply list the entries with just one space after each comma as in the code extract below.

The code extract shows how four scannos/per line should be arranged to avoid the _flake8_ error. The new code will, of course, have more scannos per line up to the maximum line length that flake8 allows.

Note that the list assigment has to be bracketed by "# fmt: on/# fmt: off" lines to prevent _black_ reformatting the list entries to one per line.
```
# fmt: off
scannos_misspelled = [
    "1gth",  "1lth", "1oth", "1Oth", 
    "2gth", "2ist", "2lst", "oa", 
    "po", "2Oth", "2OTH", "2oth", 
    "3lst",  "3Oth", "3oth", "abead",
    "ablc", "Abont", "abovc", "abscnt",
    "awav"
]
# fmt: on
print(scannos_misspelled)
```

src/guiguts/tools/pptxt.py
G4OEU: Yes. It now does that.

src/guiguts/tools/pptxt.py
G4OEU: This is fixed in the latest version of PPtxt.

src/guiguts/tools/pptxt.py
```diff
@@ -0,0 +1,2591 @@
...
+########################################################
+# pptxt.py
+# Perl author: Roger Franks (DP:rfrank) - 2009
+# Go author: Roger Franks (DP:rfrank) - 2020
+# Python author: Quentin Campbell (DP:qgc) - 2024
+# Last edit: 7-mar-2024
+########################################################
+
+
+def pptxt() -> None:
```
cpeel, line 20: We really shouldn't just put everything in the file in one function -- there's no need and it serves no purpose.

It would be better to put functions and globals at the top and have a function that is the "main" that does the work.

Integration-wise, I would expect guiguts to have a function inside `application.py` (or similar) that is the glue between Guiguts (which knows about the text) and pptxt as the tool. `pptxt.py` should not know, or need to know, anything about CheckerDialog or maintext or any of the other GG-specific bits -- treat it as a library with non-GG-specific interfaces.

src/guiguts/tools/pptxt.py
cpeel: Yes, please, let's use the ^^ format and just have black ignore the lines. The way it is now is going to be error prone I think.

G4OEU: It's fixed now in my version of PPtxt. The 3,451 scanno words are assigned in the conventional way (see above) to the scannos_misspelled list in just 390 or so lines. Extending the list should be less prone to error and it doesn't matter if duplicates are present in the list. The few lines of code that build the scannos dictionary from that list ignores any duplicates.

src/guiguts/tools/pptxt.py
windymilla: Sorry for not communicating properly with everyone (especially @cpeel) about dropping the requirement for pptxt to be a standalone tool. I don't want to end up making extra work for everyone, so I'll liaise directly with @G4OEU about the additional changes he has already made that cover some of the points above, and agree any other tweaks before we get this first major tool merged in :tada:
From that point, we can have smaller changes in commits/PRs rather than the large iterations that Quentin has kindly hurdled over as I moved the goalposts.

cpeel: Ah, the integration of GG objects in the class makes sense in that framework 👍 . My only request is that we try and move the contents of the one function up a level so they are in the top-level scope.

Thanks for all your work on this @G4OEU! It will be great to have pptxt directly usable within GG2!

G4OEU: Will do. Nigel has already asked for that in direct correspondence with me.

Thank you for your kind remarks.

PR #160, opened 2024-03-10
Add VS Code debugger configurations

tangledhelix:
There are three configs included:

1. Run with debug output
2. Run with debug output and open most recent file
3. Run without debug output

Also updates README with notes on how to choose the Python intepreter, as well as general notes on the available debug configurations.

This has been tested on macOS and Linux (Ubuntu). It **should** work on Windows as well, as there are no OS-specific paths, etc., but it's not been tested there.


srjfoo (approved):

README.md
```diff
@@ -204,6 +204,26 @@ Developers are encouraged to add tests (as appropriate) when new code is added t
...
+    - Run Guiguts with debug output enabled
+    - Open the most recently opened file
+- "Guiguts (no debug output)"
+    - Run Guiguts without debug output
+
+Requirement: [Python Debugger][vsc_debugpy] extension
+
+[vsc_debugpy]: https://marketplace.visualstudio.com/items?itemName=ms-python.debugpy
+
+Use the "Python: Select Interpreter" command to choose the appropriate Python environment. Your Poetry config should be detected and available to choose.
```
windymilla, line 225: @dlowe - Despite my VScode naivety, I managed to get GG2 working in the debugger on Windows - thanks!!!

The problem I had was that the python from the poetry config did not appear to be in the "Select Interpreter" list. I don't know if it's related to the fact that I am also using pyenv to use/select multiple python versions, which needs an additional `poetry config` before pyenv & poetry play nicely. Or maybe I was/am just doing something wrong. 

What I had to do was to run
`poetry config virtualenvs.path`
to tell me where poetry has cached the virtualenvs - in my case:
`C:\Users\<username>\AppData\Local\pypoetry\Cache\virtualenvs`
Then I used "Enter interpreter path", then "Browse your system to find an interpreter" within VScode to point it to the particular version of python I'm using with poetry.
After that, all was fine.

I also saw a [comment online](https://stackoverflow.com/questions/59882884/vscode-doesnt-show-poetry-virtualenvs-in-select-interpreter-option) about [telling poetry to put the virtualenv in the project](https://python-poetry.org/docs/configuration/#virtualenvsin-project) which said that VScode would then be able to detect it. So, I don't know if you or @cpeel have an opinion on whether I should try that out? Or should we just summarize what I've mentioned above that if VScode doesn't find the correct interpreter, you can find it from poetry, then manually browse to find it. As mentioned above, I don't know if this is Windows-specific or relates to the use of pyenv, or just that on your system poetry does put the virtual env in the project?

cpeel: I expect that it will be pretty common and for now suggest that we just document the `poetry config virtualenvs.path` + enter interpreter path approach. That way there's one less large set of local-only directories we'd need to add to `.gitignore` and get in the way of grep and other things.

.vscode/launch.json
```diff
@@ -0,0 +1,31 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
cpeel, line 2: Comments aren't a part of the JSON standard (even if VSCode supports them in it for some reason), I'd prefer we didn't include them here so we don't have invalid JSON files floating around.

cpeel (approved):

README.md
tangledhelix: @windymilla I'm not sure whether it's specific to Windows, or to using `pyenv` (which I'm not using and didn't test). I can say two things:

1. It auto-detects the Poetry environment on both Mac and Ubuntu for me
2. On neither Mac nor Ubuntu are the Poetry environments located in the project directory.

I think Casey has the right idea, probably best if you document what worked for you on Windows, and if someone comes along and it doesn't work for them, we can figure that out if needed.

## Linux:
```
ubuntu@ubuntu:~$ cd guiguts-py
ubuntu@ubuntu:~/guiguts-py$ poetry env use $(which python3.11)
Creating virtualenv guiguts-IGN9NGEH-py3.11 in /home/ubuntu/.cache/pypoetry/virtualenvs
Using virtualenv: /home/ubuntu/.cache/pypoetry/virtualenvs/guiguts-IGN9NGEH-py3.11
```

Screenshot 2024-03-10 at 11 30 02 PM

## Mac:
```
❯ poetry env info -p
/Users/dan/Library/Caches/pypoetry/virtualenvs/guiguts-6B18dp7I-py3.11
```

Screenshot 2024-03-10 at 9 18 41 PM

tangledhelix: @windymilla Perhaps `poetry env info -e` would be a shortcut in the process you described above. That should show the full path to the Python interpreter from Poetry's point of view.

.vscode/launch.json
tangledhelix: OK, removed the comments.

README.md
srjfoo: I have three Macs that I can run a variety of tests on, one each on Monterey (mid-2015 MBP (Intel)), Ventura (Early? 2019 MBP (Intel)) and Sonoma (Late 2021 MBP (M1)), but I'm running with just the [basic install](https://github.com/DistributedProofreaders/guiguts-py/blob/master/README.md#macos-development-setup). No pyenv.

What do we want to figure out (at least from a macOS point of view)? Should I try installs both with pyenv and without? Is one approach better than another? I'm happy to experiment in the name of clarifying instructions, if that's desirable.

tangledhelix: I'll defer to Nigel on that question, but my 2¢ is that it's probably not a valuable use of time to exhaustively test and document every variation of this particular setup. If this were meant for GG end-users, maybe; but it's for developers only, and only thes ones using this editor, and only those wanting to use the debugger...

PR #161, opened 2024-03-10
Refactor: remove `range` as a variable name

### LINKED ISSUE #137: Builtin class "range" is being masked by reassignment
Python has a default class `range` ([ref](https://docs.python.org/3/library/stdtypes.html#range)), but in various places we are reassigning to that name and overwriting it with another value. But in other places, we use `range()` also. Possibly they haven't come into conflict so far, but it may be best to use another name so we don't run into a problem later.

```
❯ python3
Python 3.11.6 (main, Oct  2 2023, 13:45:54) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> type(range)
<class 'type'>
>>> range = 3
>>> type(range)
<class 'int'>
```

```
src/guiguts/maintext.py:394:    def do_select(self, range: IndexRange) -> None:
src/guiguts/maintext.py:400:        self.tag_add("sel", range.start.index(), range.end.index())
src/guiguts/maintext.py:438:        for range in ranges:
src/guiguts/maintext.py:439:            start = range.start.index()
src/guiguts/maintext.py:440:            end = range.end.index()
src/guiguts/maintext.py:444:            if range.end.row > range.start.row and range.end.col == 0:
src/guiguts/maintext.py:666:        range: IndexRange,
src/guiguts/maintext.py:684:        start_index = range.start.index()
src/guiguts/maintext.py:685:        stop_index = range.end.index()
src/guiguts/search.py:240:        range = None
src/guiguts/search.py:245:                range = sel_ranges[0]
src/guiguts/search.py:248:            range = IndexRange(maintext().start(), maintext().end())
src/guiguts/search.py:249:        if range:
src/guiguts/search.py:252:                range,
src/guiguts/search.py:271:        range = None
src/guiguts/search.py:274:                range = sel_ranges[0]
src/guiguts/search.py:276:            range = IndexRange(maintext().start(), maintext().end())
src/guiguts/search.py:277:        if range:
src/guiguts/search.py:280:                range,
```

tangledhelix:
Previously we were using `range` as a variable name or an argument in some places. `range()` is a global function. This didn't cause any problems with the Guiguts logic, but it did trip up linters and is not best practice.

Fixes #137

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -190,7 +190,7 @@ class Guiguts:
 
     def help_about(self, *args: Any) -> None:
         """Display a 'Help About' dialog."""
-        help = """Guiguts - an application to support creation of ebooks for PG
+        help_message = """Guiguts - an application to support creation of ebooks for PG
 
 Copyright Contributors to the Guiguts-py project.
 
@@ -211,7 +211,7 @@ License along with this program; if not, write to the
 Free Software Foundation, Inc., 51 Franklin Street,
 Fifth Floor, Boston, MA 02110-1301 USA."""
 
-        messagebox.showinfo(title="About Guiguts", message=help)
+        messagebox.showinfo(title="About Guiguts", message=help_message)
 
     def show_page_details_dialog(self, *args: Any) -> None:
         """Show the page details display/edit dialog."""
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -434,13 +434,13 @@ class MainText(tk.Text):
         """Clear any current text selection."""
         self.tag_remove("sel", "1.0", tk.END)
 
-    def do_select(self, range: IndexRange) -> None:
+    def do_select(self, _range: IndexRange) -> None:
         """Select the given range of text.
 
         Args:
             IndexRange containing start and end of text to be selected."""
         self.clear_selection()
-        self.tag_add("sel", range.start.index(), range.end.index())
+        self.tag_add("sel", _range.start.index(), _range.end.index())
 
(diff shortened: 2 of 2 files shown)


tangledhelix:
Added one more change, the same thing was happening with `help` in one place. Pylint is currently happy, at least as far as built-in names are concerned.

src/guiguts/application.py
```diff
@@ -190,7 +190,7 @@ def quit_program(self, *args: Any) -> None:
 
     def help_about(self, *args: Any) -> None:
         """Display a 'Help About' dialog."""
-        help = """Guiguts - an application to support creation of ebooks for PG
+        help_message = """Guiguts - an application to support creation of ebooks for PG
```
srjfoo, line 193: I realize this was already this way, so this is as much a question for @windymilla as anything: Should PG be spelled out. _**We**_ all know what PG is, but "Project Gutenberg" sounds a bit more formal.

Instead of `help_message`, the variable could be "help_about" since that's how it's described in the comment, and more specific.

windymilla: Probably would be better to say Project Gutenberg, but maybe not in this PR. When making a "Help About" dialog, I just took that line from the first line of GG2's [README](https://github.com/DistributedProofreaders/guiguts-py/blob/master/README.md), so at some point we should probably revisit what is written in both places.

windymilla (approved):
Thanks for catching & fixing these @tangledhelix

srjfoo (approved):

PR #163, opened 2024-03-11
Add Project Dictionary support

windymilla:
1. Project dictionary is a JSON file with two lists (good & bad words).
2. Project dictionary is loaded when file loaded
3. File->Project->Add Good/Bad Words to Project dictionary menu entry does what it says, i.e. it looks for good_words.txt and bad_words.txt, adds them to the project dictionary and re-saves it.
4. Ctrl-clicking (Cmd-click on Mac) on spelling error reported by the Spell Checker to add it as a good word to the project dictionary. 
5. Ctrl-right-clicking (or Cmd-Ctrl-click on Mac) adds the spelling and removes that spelling error message from the list of messages
6. Shift-Ctrl-right-clicking (or Shift-Cmd-Ctrl-click on Mac) adds the spelling and removes all error messages about that spelling.
7. Can also select the spelling error message and click "Add to Project Dict" button in dialog as an alternative to Shift-Ctrl-right-click, i.e. it adds to the dictionary and removes all occurrences of that message.
8. Right-click (or Ctrl-click on Mac) just removes the clicked error message, or use the Skip button
9. Shift-right-click (or Shift-Ctrl-click on Mac) removes all occurrences of the clicked message, or use the Skip All button

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -56,3 +56,13 @@ Widgets
 =======
     
 .. automodule:: widgets
+ 
+Root
+====
+    
+.. automodule:: root
+ 
+Project Dict
+============
+    
+.. automodule:: project_dict
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -394,7 +394,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_edit = Menu(menubar(), "~Tools")
         menu_edit.add_button(
             "~Spelling Check",
-            lambda: spell_check(self.file.add_good_word_to_project_dictionary),
+            lambda: spell_check(
+                self.file.project_dict, self.file.add_good_word_to_project_dictionary
+            ),
         )
 
     def init_view_menu(self) -> None:
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -17,7 +17,6 @@ if __name__ == "__main__" and __package__ is None:
 from guiguts.file import File, NUM_RECENT_FILES
 from guiguts.maintext import maintext
 from guiguts.mainwindow import (
-    root,
     MainWindow,
     Menu,
     menubar,
@@ -25,9 +24,9 @@ from guiguts.mainwindow import (
     statusbar,
     ErrorHandler,
 )
-
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences
+from guiguts.root import root
 from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
 from guiguts.utilities import is_mac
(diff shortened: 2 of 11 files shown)


cpeel (approved):
Nice.

srjfoo:
I needed a slightly more graphic representation of the shortcuts above and put this together (please check and make sure I got it right):

```
   Windows:      | Mac:          | Action
-----------------+---------------+------------------------
4. Ctrl+click    | Cmd+click     | add spelling
-----------------+---------------+------------------------
5. Ctrl+         | Cmd+          | add spelling and remove
     right-click |   Ctrl-click* |   instance from results
-----------------+---------------+------------------------
6. Shift+Ctrl+   | Shift+Cmd+    | add spelling and remove
     right-click |   Ctrl-click* |   all instances
-----------------+---------------+------------------------
7. Click on word and use "Add to Project Dict" button, same as 6.
-----------------+---------------+------------------------
8. Right-click   | Ctrl-click*   | remove instance
                 |               |   (click, use Skip button)
-----------------+---------------+------------------------
9. Shift+        | Shift+        | removes all instances
     right-click |   Ctrl-click* |   (click, Skip all)

* Ctrl-click is a Mac shortcut for a right-click, but
  many Mac users use two-button mice.
```

I looked in the wiki to see if these map directly from GG1, but, either they don't or I found the wrong place in the GG manual. I'm going to stick my neck out and ask, from a PP standpoint, what are the most important actions? Do they have the simplest key/mouse combos? (Should that be Windows & Linux?)

srjfoo (approved):

windymilla:
> I needed a slightly more graphic representation of the shortcuts above and put this together (please check and make sure I got it right):
> 
> ```
>    Windows:      | Mac:          | Action
> <snip>
> ```
> 

I think that is an accurate summary, yes - thank you.

> I looked in the wiki to see if these map directly from GG1, but, either they don't or I found the wrong place in the GG manual. 

They don't exactly - I'm hoping to make things a bit more consistent in GG2 for all the tools:

1. To select, use primary button = mouse 1 on all OS
2. To remove, use secondary button = mouse 3 on Windows/Linux, mouse 2 (or Ctrl-1) on macOS
3. To "process" a message, hold Ctrl (or Cmd) while clicking
4. To affect all similar messages, hold Shift while clicking

So

- To process, it's 1 & 3: Ctrl+primary or Cmd+primary
- To remove all similar messages, it's 2 & 4: Shift+secondary
- To process & remove, it's 2 & 3: Ctrl+secondary or Cmd+secondary
- To process & remove all, it's 2, 3 & 4: Shift+Ctrl+secondary or Shift+Cmd+secondary

> I'm going to stick my neck out and ask, from a PP standpoint, what are the most important actions? Do they have the simplest key/mouse combos?

In trying to get the above-mentioned consistency, I don't know if it will always also be possible to have the simplest combo as the most common/important action. In the spell check case, "Skip", "Skip all" (Shift+secondary) and "Add to Dict & Remove All" (Shift+Ctrl/Cmd+secondary) are the most common. Because I've provided buttons, another way of working (that I would probably use) is you select the first message, then just use the buttons - each time you click a button, it would then go onto the next word. It would be easy to provide keyboard shortcuts to those buttons too if we want to.

The way I've structured the code, if we decide in future (before the first release!) that actually, Shift should be the modifier for "process" and Ctrl for "remove", or even another modifier like "Alt", it would be very simple to make that change, and all the tools that display results in the Checker dialog would get the change instantly.

>  (Should that be Windows & Linux?)

I believe so, yes

PR #166, opened 2024-03-12
Removed some lint

windymilla:
Did some of the easy/non-controversial ones.

Rating has changed from 9.38 to 9.68, I think


src/guiguts/application.py
```diff
@@ -45,21 +41,24 @@ def __init__(self) -> None:
 
         Creates windows and sets default preferences."""
 
-        self.parse_args()
+        self._parse_args()
```
cpeel, line 44: What is pylint complaining about that's causing these renames? Is it being opinionated about "private" functions  being prefaced by `_` or something?

tangledhelix: I'm curious about that too. My current master branch has this particular line of code at line 48 of application.py, and pylint isn't complaining about anything on that line (or very near there either) for me.

(I didn't review the overall PR - I just checked this line quickly because Casey called it out.)

windymilla: Pylint was complaining about too many public methods in the class, and those methods are only called internally in the class, so I tried adding the underscore to flag them as "protected". There are similar complaints about numbers of methods/attributes elsewhere that I didn't tackle. It would probably be good to consider if we think whether we want to do anything about that style of complaint, or if we'd rather suppress them. Any thoughts?

windymilla: Added-underscore changes reverted now. I'm happy to go with your thoughts @tangledhelix and @cpeel on a the best way to handle the "too many/too few" pylint messages

cpeel: I would disable those checks rather than doing gymnastics for the sake of appeasing it. I don't think those types of checks are going to make the code inherently better.

windymilla: I've disabled those, but left the ones about too many arguments and too many instance attributes, in case we want to decide that they are reasonable warnings that we will generally try to abide by (possibly adjusting the threshold number that causes the warning). 
If you'd rather these were not disabled for now, @tangledhelix, I don't mind reverting this last commit, merging this PR, and then having a discussion about which of the remaining warnings we want to disable/adjust/fix.

src/guiguts/application.py
```diff
@@ -182,12 +181,12 @@ def update_title(self) -> None:
         filetitle = " - " + self.file.filename if self.file.filename else ""
         root().title("Guiguts 2.0" + modtitle + filetitle)
 
-    def quit_program(self, *args: Any) -> None:
```
tangledhelix, line 185: These confused me a bit, but digging into why you made this change, I presume it's because of `unsued-argument`, e.g.:

```
src/guiguts/application.py:185:0: W0613: Unused argument 'args' (unused-argument)
```

And in fact, `args` was not being used in the method. But then, neither is the new form `_args` ... pylint must be happy enough to ignore that if you use a `_` prefix?

My 2¢ is that unused arguments should just be dropped; if the method changes in the future in a way that requires an argument, then an argument (including something like `*args` and/or `**kwargs`) can be added, but otherwise I would rather remove them than rename them to mollify the linter. The argument is still unused, regardless of the name.

I am making an assumption here, that you were aiming for a standard interface to most functions. (I could be wrong, of course.) While I would agree that `*args` and/or `**kwargs` is better than a whole slew of positional arguments, having them here when they aren't used seems (IMHO) confusing to a future developer, who may be unsure what they're doing there when they aren't being used. In this very short 2-line function, it's really obvious it's not used, but in a longer one it may be unclear. So I'd just remove it.

"The only bug-free code is no code" as they say...

src/guiguts/application.py
tangledhelix: I agree with Casey.

src/guiguts/application.py
windymilla: @tangledhelix, the reason the args is there is not because I was aiming for a standard interface. 

The problem arises because several of these functions are run via two routes: either through a menu button, or through a shortcut key-combination, like `Cmd+w` to call `close_file`. When called via a menu button, Tk doesn't pass in any arguments, but when called via a shortcut, it passes in a `Tk. event`. Then, without the `*args`, an exception is generated that an extra argument was given. 

Another way of "solving" it would be to have a lambda function which similarly has args that it ignores and calls `close_file` without any args. I did that at line 298
```python
        self.menu_file.add_button(
            "~Open...", lambda *args: self.open_file(), "Cmd/Ctrl+O"
        )
```
Although pylint doesn't complain about that one 🤷 it's only moving the problem somewhere else.

Another solution would be to have two functions, one which takes an arg and one which doesn't. The former would be called in the "event" case, and would do nothing but ignore its arg and call the latter. The latter would be called directly in the "menu" case. That is again just moving the problem.

As I type this, I wonder if I could hide the "arg ignoring" inside the `add_button` method, with a comment explaining. Then it would only be in one place. An advantage of that is it would avoid the situation (which happened to me a few times in the early stages) where you bind a function but there's no shortcut for it, so you don't have an unused arg, then later you add a shortcut, and it suddenly fails. 

I'll put up another commit, and see what you think.

windymilla: OK @tangledhelix - I've made the argument deliberately ignored in the common binding routine where possible, rather than in lots of different callbacks. I think that approach makes much more sense. I did also find a couple of functions that had the optional arg but were never called with an arg - probably copy/paste errors.
All in all, I think it's neater now. Let me know what you think, and I'll either merge, or do some more tweaking if you think it's necessary.
Thanks for helping puzzle through it.

tangledhelix: The PR review view still shows a lot of `_args` ... confusing ... but I looked at your actual branch and searched for all of them and the `*args` or `*_args` that I did find mostly make sense. Looks good to me! (I did not test anything, just read code.)

windymilla: Thanks - I'll get it merged shortly, and we can always tweak further when we see what it looks like, and either work in a specific file, or tackle one or two more of the lint messages.

PR #167, opened 2024-03-18
Checker dialog fixes

### LINKED ISSUE #159: Checker dialog issues
1. Running pptxt, and keeping the checker dialog visible and running S&R Find All re-uses the Checker dialog, which is sort of correct, except it's still the pptxt version of it, so its Re-run button runs pptxt, rather than Find All even though the messages in the dialog are Find All messages. I suspect if there were custom widgets, like Spell Query threshold, they would be left behind too. Probably easiest to either always delete the Checker dialog and create it from scratch, or possibly just do that if the type of dialog is different, e.g. SpellCheck to pptxt.
2.  Focus can end up on the Re-run button or the label(?) - anyway, not in the text widget. It's useful to be able to focus in the text widget, so selection can be done, e.g. Ctrl+a, Ctrl+c to copy all messages to clipboard. Make the intercepted left/right clicks set focus in the checker dialog text widget at the click point? Or possibly at the start of the clicked line?
3. Number of entries displayed at top of dialog is not updated when entries are removed.

windymilla:
Fix #159 - several checker dialog issues

1. Stop confusion between different checker dialogs if a new one is popped before the old is closed, e.g. Spell Check and S&R FInd All dialogs. Checker dialogs are used for a variety of tools, and vary in their appearance and controls. They therefore need destroying and recreating, not just deiconifying, if they already exist. This could be solved by inheriting different dialogs from CheckerDialog, e.g. SpellCheckDialog, FindAllDialog, etc. However, it would still be necessary to deal with things like the language changing or the project dictionary changing while a dialog exists, in which case some functions in the dialog would not work correctly. This seems the best fix for now, with very little cost. Other dialogs that inherit from TopLevelDialog, such as MessageLog, do not need to be destroyed & re-created.

2. Position insert cursor at start of current line in CheckerDialog when line is selected. This makes it easier to use Ctrl+a, Ctrl+c to copy all messages, or use Shift+arrow keys to select specific messages for copying.

3. Keep "Entries" count up-to-date. When entry is removed from CheckerDialog, reduce the count.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -342,6 +342,9 @@ class CheckerDialog(ToplevelDialog):
             event: Event object containing mouse click position.
         """
         self.highlight_entry(entry_index)
+        self.text.mark_set(tk.INSERT, f"{entry_index+1}.0")
+        self.text.see(tk.INSERT)
+        self.text.focus_set()
         entry = self.entries[entry_index]
         if entry.text_range is not None:
             start = maintext().index(self._mark_from_rowcol(entry.text_range.start))
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -129,6 +129,7 @@ class CheckerDialog(ToplevelDialog):
         """Reset dialog and associated structures & marks."""
         self.entries: list[CheckerEntry] = []
         self.count_linked_entries = 0  # Not the same as len(self.entries)
+        self.update_count_label()
         self.text.delete("1.0", tk.END)
         for mark in maintext().mark_names():
             if mark.startswith(MARK_PREFIX):
@@ -174,6 +175,12 @@ class CheckerDialog(ToplevelDialog):
             maintext().mark_set(
                 self._mark_from_rowcol(text_range.end), text_range.end.index()
             )
+            # If none selected, select the first message with a text range
+            if self.current_entry_index() is None:
+                for index, entry in enumerate(self.entries):
+                    if entry.text_range:
+                        self.select_entry(index)
+                        break
         self.update_count_label()
 
     def update_count_label(self) -> None:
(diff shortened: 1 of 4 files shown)


cpeel (approved):

srjfoo (approved):
Just to make sure that I'm understanding the first point correctly: certain dialog results are not allowed to co-exist? Is it a situation that doesn't occur in GG1 (since Find All is not present there)? And is there a situation where it would be appropriate for search results and the contents of find all available at the same time?

That's just curiosity speaking -- it looks good to me!

windymilla:
> Just to make sure that I'm understanding the first point correctly: certain dialog results are not allowed to co-exist? Is it a situation that doesn't occur in GG1 (since Find All is not present there)? And is there a situation where it would be appropriate for search results and the contents of find all available at the same time?
> 
The situation does arise in GG1 too. All the "checks" share the same output dialog, e.g. pptxt, jeebies, HTML validator, etc. If we decide that it would be better for Find All, for example, to have it's own instance of the dialog, and the others to share, or for all the others to have their own, the code change is very easy, essentially just inheriting a new FindAllCheck dialog from the base CheckerDialog. The code stores one set of info, and only allows one instance, for each dialog class, so at the moment they all use the CheckerDialog class, but could easily have subclasses of that in a way that makes them share/not share as we desire. An alternative would be to have a "dialog_name" or "dialog_subtype" attribute, that would also be checked. GG1 has something like that (and GG2 may also need it) so that when you press F1 for help on the dialog, it takes you to the correct manual page, i.e. pptxt, jeebies, HTML validator, etc., even though they share a dialog. I suspect it will be better to keep those two functions separate, so probably using the subclass idea for where we want a separate dialog, and using the attribute idea for things like links to the manual page.

PR #171, opened 2024-03-20
More lint fixes

windymilla:
1. Disabled some messages completely, mostly "too-many..." code-style messages, but also "global-statement" warnings - see below for discussion.
2. Retained "too-many-boolean-expressions" as I thought it was a good thing to avoid, and also not likely to happen often, so ok to disable on a per-line basis.
3. Several warnings about improved way to access dictionaries, e.g. using `.items()` acted on.
4. More `*args` arguments removed.
5. Enforced encoding when opening files - good to find, since .json file wasn't being written with utf-8 encoding, so failed to read with utf-8 encoding, depending on contents.

For discussion and advice, please:

- Is disabling the "global-statement" warnings ok with you. We use globals in a few places, and IMO, our code review process should be sufficient to pick up on cases where a developer uses globals inappropriately, rather than warnings from pylint, but I'm happy to be guided by more experienced python/pylint users.
- This brings the list of lint errors down to 9, consisting of two types: complaints about global "constants" not being all UPPER_CASE, and suggestions about re-raising exceptions differently.
    - For the first, they aren't uppercase because they aren't actually constant - do you think it best to disable that warning, or to keep the warning and make the "constants" uppercase, or add comments to disable the warning on those 5 specific lines?
    - For the second, I don't understand the difference (if any) or advantage in the suggestion to re-raise those Exceptions differently, so would value advice on the best way to tackle those: do what pylint suggests, or disable the warning, or disable on a per-line basis. Similarly for the one "too general exception" warning.

For reference, here's the output from pylint:
```
************* Module maintext
src\guiguts\maintext.py:787:16: W0707: Consider explicitly re-raising using 'raise TclRegexCompileError(str(exc)) from exc' (raise-missing-from)
src\guiguts\maintext.py:837:20: W0707: Consider explicitly re-raising using 'raise TclRegexCompileError(str(exc)) from exc' (raise-missing-from)
src\guiguts\maintext.py:956:0: C0103: Constant name "_single_widget" doesn't conform to UPPER_CASE naming style (invalid-name)
************* Module root
src\guiguts\root.py:17:0: C0103: Constant name "_the_root" doesn't conform to UPPER_CASE naming style (invalid-name)
************* Module spell
src\guiguts\spell.py:25:0: C0103: Constant name "_the_spell_checker" doesn't conform to UPPER_CASE naming style (invalid-name)
************* Module utilities
src\guiguts\utilities.py:20:0: C0103: Constant name "called_from_test" doesn't conform to UPPER_CASE naming style (invalid-name)
src\guiguts\utilities.py:52:12: W0707: Consider explicitly re-raising using 'except AttributeError as exc' and 'raise Exception('Unknown windowing system') from exc' (raise-missing-from)
src\guiguts\utilities.py:52:12: W0719: Raising too general exception: Exception (broad-exception-raised)
src\guiguts\utilities.py:179:0: C0103: Constant name "_bell_callback" doesn't conform to UPPER_CASE naming style (invalid-name)

```

Regarding the last few pylint messages, I'm very happy to make suggested changes, or for someone else to add a commit, or to leave them for a future PR, possibly one which also makes a clean pylint compulsory instead of advisory as it is at present, as @tangledhelix suggested we could when all the lint was fixed.

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -294,7 +294,7 @@ class File:
         """
         binfile_name = bin_name(basename)
         bin_dict = self.create_bin()
-        with open(binfile_name, "w") as fp:
+        with open(binfile_name, "w", encoding="utf-8") as fp:
             json.dump(bin_dict, fp, indent=2, ensure_ascii=False)
 
     def interpret_bin(self, bin_dict: BinDict) -> bool:
@@ -563,7 +563,7 @@ class File:
         if image_num is not None:
             try:
                 index = maintext().rowcol(page_mark_from_img(image_num))
-            except tk._tkinter.TclError:  # type: ignore[attr-defined]
+            except tk.TclError:
                 # Bad image number
                 return
             maintext().set_insert_index(index)
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -32,7 +32,6 @@ jobs:
     - name: Lint with pylint
       run: |
         poetry run pylint --recursive y .
-      continue-on-error: true
     - name: Check format with black
       run: |
         poetry run black --check --verbose ./src
--- a/README.md
+++ b/README.md
@@ -144,14 +144,8 @@ All of the above tools will be installed via `poetry` as described above.
 
 `poetry run black .` will reformat all `src` & `tests` python files where necessary.
 
-This project uses Github Actions to ensure neither flake8 nor black report any
+This project uses Github Actions to ensure none of the above tools report any
 error.
-
-> **NOTE:** At this time, `pylint` is running in GitHub Actions in an advisory
-mode only.  That means that any warning or error reported by `pylint` will not
-cause the Action to fail. Over time, as we make changes in Guiguts and/or the
-`pylint` configuration, the aim is to clear all of `pylint`'s warnings. At that
-point perhaps it can run in an enforcement mode.
 
 Naming conventions from [PEP8](https://pep8.org/#prescriptive-naming-conventions)
 are used. To summarize, class names use CapWords; constants are ALL_UPPERCASE;
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,6 +3,8 @@
 import os
 import sys
 
+# pylint: disable=invalid-name
+
 project = "Guiguts 2.0"
 
 sys.path.insert(0, os.path.abspath("../src/guiguts/"))
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -66,3 +66,9 @@ Project Dict
 ============
     
 .. automodule:: project_dict
+
+ 
+PPtxt
+============
+    
+.. automodule:: tools.pptxt
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,16 @@ disable = [
   "logging-not-lazy",
   "too-few-public-methods",
   "too-many-ancestors",
-  "too-many-public-methods"
+  "too-many-public-methods",
+  "too-many-statements",
+  "too-many-instance-attributes",
+  "too-many-locals",
+  "too-many-arguments",
+  "too-many-return-statements",
+  "too-many-lines",
+  "too-many-branches",
+  "too-many-nested-blocks",
+  "global-statement"
 ]
 
 enable = ["c-extension-no-member"]
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -105,6 +105,8 @@ class MainText(tk.Text):
         super().__init__(self.frame, **kwargs)
         tk.Text.grid(self, column=1, row=0, sticky="NSEW")
 
+        self.languages = ""
+
         # Create Line Numbers widget and bind update routine to all
         # events that might change which line numbers should be displayed
         self.linenumbers = TextLineNumbers(self.frame, self)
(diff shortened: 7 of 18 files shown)


windymilla:
Testing notes for @srjfoo: should be no change in functionality - a quick poke around the various features, and maybe a check that I haven't changed the output from pptxt would be good, thanks

src/guiguts/file.py
```diff
@@ -563,7 +563,7 @@ def do_goto_image(self, image_num: Optional[str]) -> None:
         if image_num is not None:
             try:
                 index = maintext().rowcol(page_mark_from_img(image_num))
-            except tk._tkinter.TclError:  # type: ignore[attr-defined]
```
tangledhelix, line 566: I believe this could be rewritten as `tk.TclError`; functionally the same, but it would stop the linter complaint.

```
>>> import tkinter as tk
>>> tk._tkinter.TclError is tk.TclError
True
```

tangledhelix (changes requested):
This looks good to me in general.

The purist in me doesn't like using globals, but they sometimes make things so much simpler. So I lean in both directions - it's likely okay to just ignore them in the linter and rely on code reviewers, but I'd be equally happy to see the linter complain and have any new globals marked as exceptions with the pylint ignore statements.

The only change I'd make is the protected-access one for TclError, see that comment.

src/guiguts/file.py
windymilla: Thanks - you're right - not sure why I ended up making it more complicated.

windymilla:
Thanks for the review, @tangledhelix. Do you have any thoughts on getting rid of the last 9 pylint messages?

srjfoo:
> Testing notes for @srjfoo: should be no change in functionality - a quick poke around the various features, and maybe a check that I haven't changed the output from pptxt would be good, thanks

Some questions I had during the first round of testing were reported in Slack. Second round of testing, as far as I can see, things look good (touching keyboard shortcuts, pptxt, s/r and spell check). If there's anything else I need to look at, though, let me know.

srjfoo (approved):

tangledhelix:
> Thanks for the review, @tangledhelix. Do you have any thoughts on getting rid of the last 9 pylint messages?

These three seem pretty straightforward ([raise-missing-from](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/raise-missing-from.html)):

```
src/guiguts/maintext.py:787:16: W0707: Consider explicitly re-raising using 'raise TclRegexCompileError(str(exc)) from exc' (raise-missing-from)
src/guiguts/maintext.py:837:20: W0707: Consider explicitly re-raising using 'raise TclRegexCompileError(str(exc)) from exc' (raise-missing-from)
src/guiguts/utilities.py:52:12: W0707: Consider explicitly re-raising using 'except AttributeError as exc' and 'raise Exception('Unknown windowing system') from exc' (raise-missing-from)
```

Add `from exc` at the end of the line, and pylint should be happy (and, probably improve the exception if/when it throws).

```python
        except tk.TclError as exc:
            if str(exc).startswith("couldn't compile regular expression pattern"):
                raise TclRegexCompileError(str(exc)) from exc
```

For [broad-exception-raised](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/broad-exception-raised.html), I think it wants something more specific than `Exception`, so perhaps use `RuntimeError` instead?

> exception RuntimeError
> Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

And then there are the [invalid-name](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html) lines...

```
src/guiguts/maintext.py:956:0: C0103: Constant name "_single_widget" doesn't conform to UPPER_CASE naming style (invalid-name)
src/guiguts/root.py:17:0: C0103: Constant name "_the_root" doesn't conform to UPPER_CASE naming style (invalid-name)
src/guiguts/spell.py:25:0: C0103: Constant name "_the_spell_checker" doesn't conform to UPPER_CASE naming style (invalid-name)
src/guiguts/utilities.py:20:0: C0103: Constant name "called_from_test" doesn't conform to UPPER_CASE naming style (invalid-name)
src/guiguts/utilities.py:179:0: C0103: Constant name "_bell_callback" doesn't conform to UPPER_CASE naming style (invalid-name)
```

I looked at the [pylint documentation](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html) on this one. Python doesn't have a const declaration so I was curious how it decides what it thinks is a const.

> `const`
> Module-level constants, any variable defined at module level that is not bound to a class object.

Simple enough... it's deciding these are constants simply due to the place they appear.

I think for these we can go either way... I didn't carefully survey all of them, but I don't think any of the names are magic values that can't be changed? So we could simply rename them to UPPER_CASE to mollify pylint. Or we could add ignore markers to make it shut up... or some of both, case by case...

I'll defer to you as to whether the current names should be left alone. I looked at a couple of them but I didn't carefully think through all of them, so I don't want to just knee-jerk and say "yeah rename 'em" ...

windymilla:
Squashed the various lint-removing commits together, which together remove all lint from the project.
Final commit to make the Github Action enforce pylint-compliant code & update the docs accordingly.

Thanks, @tangledhelix, if you wouldn't mind a quick final look, I'd be grateful.

src/guiguts/file.py
```diff
@@ -563,7 +563,8 @@ def do_goto_image(self, image_num: Optional[str]) -> None:
...
+            except tk.TclError:
+                print("here", flush=True)
                 # Bad image number
```
tangledhelix, line 568: left-over debugging statement?

tangledhelix (changes requested):
One debugging statement lingering, other than that it looks good to me!

PR #175, opened 2024-03-21
Further small checker dialog fixes

windymilla:
1. Make "destroy" True by default for checker dialogs - it is False by default for other Toplevel dialogs. The pptxt work was merged without the correct setting for "destroy", which is what prompted this edit. The bug which this fixes is that if spellcheck is run, and then pptxt, then pptxt re-uses the spellcheck dialog, with the consequence that clicking "re-run" re-runs spellcheck instead of pptxt.
2. Make FindAll independent of the other checker dialogs, so that it can be displayed at the same time. It is a more general tool, rather than a "checker", so deserves it's own life, otherwise you can't do Spellcheck, followed by Find All occurrences of word.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -398,6 +398,6 @@ class CheckerDialog(ToplevelDialog):
             rowcol: Location in text file to be marked.
 
         Returns:
-            Name for mark, e.g. "chk123.45"
+            Name for mark, e.g. "Checker123.45"
         """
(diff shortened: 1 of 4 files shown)


srjfoo (approved):

windymilla:
Rebased to include just-merged lint fixes, and fixed lint error from this PR. No change in functionality

srjfoo (approved):

cpeel (approved):

PR #176, opened 2024-03-22
Add ordinal & unicode char name to status bar

windymilla:
The current character's hex ordinal, and unicode name are displayed in a new status bar button.

In future, we may want to re-order the status bar buttons, to put this last because it changes width by quite large amounts for different characters. Alternatively, we may decide to make it a fixed width, or a minimum width that will cover most characters. Or, we may decide to put some buttons, like language on the right hand side of the status bar, so they are not affected by the length of this button.
I'm putting that decision off till later.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -7,6 +7,7 @@ import logging
 import os.path
 from tkinter import messagebox
 from typing import Optional
+import unicodedata
 import webbrowser
 
 
@@ -477,6 +478,19 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         the_statusbar.add_binding(
             "page label", "<ButtonRelease-3>", self.show_page_details_dialog
         )
+
+        def ordinal_str() -> str:
+            """Format ordinal of char at current insert index for statusbar."""
+            char = maintext().get(maintext().get_insert_index().index())
+            # unicodedata.name fails to return name for "control" characters
+            # but the only one we care about is line feed
+            try:
+                name = unicodedata.name(char)
+            except ValueError:
+                name = "LINE FEED" if char == "\n" else ""
+            return f"U+{ord(char):04x}: {name}"
+
+        the_statusbar.add("ordinal", update=ordinal_str)
 
         the_statusbar.add("languages label", text="Lang: ")
         the_statusbar.add_binding(


tangledhelix:
I could see this potentially confusing users sometimes. I'll explain some different scenarios:

1. Normally the character to the right of the cursor is displayed. So long as people know it's the one to the right, it makes sense, but it's not necessarily obvious it's the character to the right vs. to the left ... at least not without thinking about it.
2. If you select one character, you might expect to be shown that character. But what in fact happens is, it still shows whichever character is to the right of the cursor.
3. If you drag R-to-L to select that one character, the cursor ends up on the left side of the selection, so it would show you the selected character.
4. But if you drag L-to-R, the cursor ends up on the right side of the selection, so now it shows the character to the right of the cursor/selection.

The behavior is the same if multiple characters are selected, but in that case I would not expect it to be related to the selection. It only seems like it could be confusing when a single character is selected.

I'm not sure whether this is important and maybe isn't a blocker -- it's just something I noticed and wanted to point out.

srjfoo:
Dan's comments got me thinking, so I checked:

The information given in the status bar for gg1 is the same as in gg2. I.e. if you select a character, the position reported for the cursor depends on how the character was selected, and if the insertion cursor is at the beginning of a line, the column number reported is 0:
* double-click, the insertion cursor appears to the right, and the character following the selection (i.e. following the insertion cursor) is the one that is reported. If at end or beginning of line, or file, `000A` is reported.
* click and drag right, ditto above.
* click and drag left, the insertion cursor is displayed to the left of the character, and the correct unicode character is displayed in the status bar.

\<tb>

I looked also at Sublime Text and BBEdit. For both, if the insertion cursor is at the beginning of a line with no selection, the column number reported is column 1.

Sublime Text: If you select a character, cursor position is not reported, just the number of characters selected. And lines if more than one line. (I use Sublime Text more for occasional checking than anything else, so I haven't played around enough with the settings to know if there are other things that can be added to the status bar.

BBEdit: Whatever method is used to select a character, the cursor position is reported to be at the column position where the selected character is. I.e. the column position a typed character would appear. Another setting in the status bar reports either the number of characters/words/lines in the file, or the number of characters/words/lines selected if there's an active selection.

\<tb>

I don't know if it's possible, but I think that, regardless of how column numbers are reported
* if a column is reported for a selection, it should be consistent, whether double-clicked, or click-and-dragged rtl or ltr
* if a character is selected, the character selected should have the code reported for that character, regardless of which direction the character is selected in, or, a code should not be shown at all. (If more than one character is selected, perhaps it shouldn't be shown at all?)

tangledhelix:
For the sake of completeness, I tested a bit on Linux (Ubuntu), and the behavior was the same as what I wrote already.

windymilla:
As @srjfoo said, the behavior is (at least intended to be) identical to GG1 (which has behaved that way "for ever"). I haven't been aware of any GG1 users reporting confusion - the character reported is the one immediately following the cursor, regardless of whether you have selected one, many or no characters, and regardless of the direction of any selection. It's always the character immediately after the insert cursor.
It would be possible/easy to extend the current feature so that if just a single character is selected, then the ordinal/name of that character is shown instead of the one following the cursor. A decision would need to be made what to do if two or more characters were selected.
In any case, I don't think such a change from GG1 should be made at this point. If we thought it would be good as an future feature, it should probably be discussed with users first.

windymilla:
Rebased to include just-merged lint fixes

tangledhelix:
> As @srjfoo said, the behavior is (at least intended to be) identical to GG1 (which has behaved that way "for ever"). I haven't been aware of any GG1 users reporting confusion - the character reported is the one immediately following the cursor, regardless of whether you have selected one, many or no characters, and regardless of the direction of any selection. It's always the character immediately after the insert cursor.

Avoiding user confusion is important, so I can't imagine any reason to change this basic behavior.

> It would be possible/easy to extend the current feature so that if just a single character is selected, then the ordinal/name of that character is shown instead of the one following the cursor.

A decision would need to be made what to do if two or more characters were selected. In any case, I don't think such a change from GG1 should be made at this point. If we thought it would be good as an future feature, it should probably be discussed with users first. The notes on the behavior of other editors would be useful for that conversation. By the way, the same is true of the column number that is shown - it is the column number of the insert cursor, not of "the selection".

windymilla:
Added issue #177 to capture some of the above discussion with a view to considering a change of behavior in future.

tangledhelix:
## TL;DR

I propose that when a selection (including column select) is in effect, we should display nothing. That is, unless the selection is exactly one character, in which case, we should display the selected character.

Note: While I was composing and revising this, Nigel created #177 which is related.

## My observations

The character display is (currently) determined by the insertion point.

The insertion point can end up in one of **four** places when making a selection, depending on how you made the selection.

| Insertion point | How selection was made |
|----|----|
| Left edge | Drag with mouse, right-to-left |
| | Keyboard select, right-to-left |
| | Double-click on a word |
| | Triple-click on a line |
| | Shift-click to select from insertion point to HERE, where insertion point was later in the file |
| | Search using the Search & Replace tool (same result for forward or reverse direction) |
| Right edge | Drag with mouse, left-to-right |
| | Keyboard select, left-to-right |
| | Shift-click to select from insertion point to HERE, where insertion point was earlier in file |
| Somewhere in the middle! 😵 | Double-click a word, then move pointer before releasing click |
| | Triple-click a line, then move pointer before releasing click |
| Random, possibly even unrelated location | Column selection (see below) |

The behavior for column select seems confusing (IMHO). The insertion point doesn't move in that scenario, and is not even related to the selection area; they can be in two unrelated locations. Therefore, it seems even more confusing than a normal span-based selection -- the displayed character may not even be a part of the selected region!

(I tested all of the above on Mac and Ubuntu -- I would guess Windows has the same behaviors but I'm not certain.)

## Responses to previous comments

> As @srjfoo said, the behavior is (at least intended to be) identical to GG1 (which has behaved that way "for ever"). I haven't been aware of any GG1 users reporting confusion - the character reported is the one immediately following the cursor, regardless of whether you have selected one, many or no characters, and regardless of the direction of any selection. It's always the character immediately after the insert cursor.

Avoiding user confusion is important, of course. I can't imagine a reason to change the base behavior (where nothing is selected). However, I feel that changing the behavior when a selection is in effect would be a major UX improvement. 

> It would be possible/easy to extend the current feature so that if just a single character is selected, then the ordinal/name of that character is shown instead of the one following the cursor.

My 2¢ is that for the case where a single character is selected, the selected character should be displayed. It seems like the only intuitive behavior for that case.

> A decision would need to be made what to do if two or more characters were selected. In any case, I don't think such a change from GG1 should be made at this point. If we thought it would be good as an future feature, it should probably be discussed with users first. The notes on the behavior of other editors would be useful for that conversation. By the way, the same is true of the column number that is shown - it is the column number of the insert cursor, not of "the selection".

Asking the users makes sense, of course. There could be strong opinions about it. As I noted already, my opinion about it is: display nothing in that case.

If someone asks users in the forums, feel free to link to this post that lists all the different behaviors, perhaps it will help illustrate the situation?

windymilla:
Thanks for the detailed analysis @tangledhelix. Since #177 links here, we can use that to inform us when #177 is considered.
I agree that the column selection especially is not obvious and almost certainly not desirable. I'll add a specific mention of it to #177.

srjfoo (approved):
I realize this is for the future, but ...
> In future, we may want to re-order the status bar buttons, to put this last because it changes width by quite large amounts for different characters. ...

This would be my choice. If you make it a set length that is shorter than the descriptions for common accented characters, there'll be a lot of truncation; if you set it so that you can see all of `U+0101: LATIN SMALL LETTER A WITH MACRON`, there will be a lot of wasted space for most letters.

windymilla:
Added [issue #181](https://github.com/DistributedProofreaders/guiguts-py/issues/181) to consider ordering, probably once the remaining buttons have been added to the status bar.

cpeel (approved):

PR #187, opened 2024-04-04
Implement jeebies

G4OEU:
Based on the Go version at ppwb,
Uses 3-form algorithm, e.g. "w1-he-w2"


src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+        if rlim != len(para) - 1:
+            rlim -= 1
+
+        return llim, rlim + 1, para[llim : rlim + 1]
+
+    def find_in_dictionary(self, three_form: str) -> int:
+        """returns frequency in hebe phrase corpus or zero"""
+
+        words = three_form.split()
+        key = words[0] + "|" + words[1] + "|" + words[2]
```
cpeel, line 245: Minor suggestion: if `words` only has 3 things in it you could do `"|".join(words)`

src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+        freq = 0
+        if key in self.dictionary:
```
cpeel, line 247: Consider:
```python
freq = self.dictionary.get(key, 0)
```

src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+        either 2-form (|he|a:168) or 3-form (since|he|stopped:7) phrases.
+        """
+        file = "hebelist.txt"
+        path = DEFAULT_DICTIONARY_DIR.joinpath(file)
+        try:
+            with path.open("r", encoding="utf-8") as fp:
+                for line in fp:
+                    line = line.strip()
+                    if ":" in line:
+                        parts = line.split(":")
```
cpeel, line 280: You can also do:
```python
key, value = line.split(":")
```

That makes the variables a bit clearer when you use them later.

src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+        """Check for jeebies in the currently loaded file."""
+
+        # Create the checker dialog to show results
+        checker_dialog = CheckerDialog.show_dialog(
+            "Jeebies Results", rerun_command=jeebies_check
+        )
+        checker_dialog.reset()
+
+        # Get the whole of the file from the main text widget
+        text = maintext().get_text()
```
cpeel, line 44: Since you don't need `text` later, consider:
```python
input_lines = maintext().get_text().splitlines()
```

This makes it so we only have one copy of the text in memory -- the array form.

src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+        wbs = []
+        # Paragraphs as single string, all lower case
+        wbl = []
+
+        # for line, line_num in maintext().get_lines():
+        s = ""
+        be_cnt_in_file = 0
+        he_cnt_in_file = 0
+        for line in input_lines:
+            line = line.rstrip("\r\n")
```
cpeel, line 59: FYI I don't _think_ you need to specify `\r\n` because both of those are whitespace and by default `.rstrip()` removes all whitespace. What you have is fine too, clearly.

src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+            )
+        checker_dialog.add_entry("")
+
+        be_suspects_cnt = self.find_and_report_be_phrases(wbs, wbl, checker_dialog)
+        he_suspects_cnt = self.find_and_report_he_phrases(wbs, wbl, checker_dialog)
+
+        if be_suspects_cnt == 0 and he_suspects_cnt == 0:
+            checker_dialog.add_entry("    No suspect phrases found.")
+
+    def find_and_report_be_phrases(
```
cpeel, line 91: Since this and `find_and_report_he_phrases()` are identical except for the `he` and `be` strings consider merging them into one so we don't need to worry about them drifting. You could even make this the core function `_find_and_report_xx_phrases(self, phrase: string, ...)` and have named wrapper functions that call them with `he` or `be`.

cpeel (commented):
Looks great. Some minor suggestions / comments.

src/guiguts/tools/jeebies.py
```diff
@@ -0,0 +1,297 @@
...
+        freq = 0
+        if key in self.dictionary:
+            freq = self.dictionary[key]
+
+        return freq
+
+    def add_to_dialog(
+        self, header: str, para_slice: str, checker_dialog: CheckerDialog
+    ) -> None:
+        """helper function that abstracts repeated code
```
windymilla, line 255: Sentence case with a full stop for the first sentence of the docstring, please. Similarly for the description of the args. These strings get extracted into the docs by sphinx, and the HTML files look better with consistent capitalisation. Thanks.

windymilla (commented):
Thanks - I like the way it mimics the spell check structure - hopefully we can use similar for all the tools, and if/when appropriate consolidate any other significant bits of code that turn out to be common in the structure for all tools.
I'm happy with the behaviour in the tests I've done.

src/guiguts/tools/jeebies.py
```diff
@@ -16,6 +16,16 @@
...
 
 
+########################################################
+# jeebies.py
+# C author: Jim Tinsley (DP:jim) - 2005
+# Go author: Roger Franks (DP:rfrank) - 2020
+# Python author: Quentin Campbell (DP:qgc) - 2024
+########################################################
+# 2024.04.06 Used sentence case with full stop for first
+#            sentence of docstrings. Ditto for args.
```
windymilla, line 26: Although I'm not against having a record of the original authors of these tools, I don't think we want to add comments about all the other changes we make (like the docstring capitalization). Rather than looking in the code comments to see what changes have been made recently, Github provides us with several mechanisms to help do that, such as the PR comments/conversations, the commit comments, and "blame" view (e.g. see [this blame view](https://github.com/DistributedProofreaders/guiguts-py/blame/master/src/guiguts/maintext.py) of maintext.py where you can see who edited each line of code and why)
So, I'd prefer to see these two comment lines removed, please.

windymilla: The docstrings look great now, thank!

G4OEU:
No problem. Will do.

src/guiguts/tools/jeebies.py
windymilla: @G4OEU - I noticed that your new version still has two very similar copies of the code, although they are now both contained within the same function.

I interpreted @cpeel's suggestion as meaning combine them into one version of the code, but using variables in place of "he" and "be", so `hebe` for "he" or "be" as you have used, and `behe` for "be" or "he" respectively. 
So it could look something like this (untested - just grabbed and edited from your version - I may easily have hebe and behe the wrong way round in places):

```python
    def find_and_report_hebe_phrases(
        self, hebe: str, wbs: list, wbl: list, checker_dialog: CheckerDialog
    ) -> int:
        """Look for suspect "w1 be w2" or "w1 he w2" phrases in paragraphs.
        """

        suspects_count = 0
        # Get the opposite word to the one being reported on
        behe = "he" if hebe == "be" else "be"

        # Search for 3-form pattern "w1 he/be w2" in a lower-case copy of paragraph.
        for rec_num, para_lc in enumerate(wbl):
            for match_obj in re.finditer(fr"[a-z’]+ {hebe} [a-z’]+", para_lc):
                # Have a 3-word form here ("must be taken").
                # See if it's in the list.
                hebe_form = para_lc[match_obj.start(0) : match_obj.end(0)]
                hebe_count = self.find_in_dictionary(be_form)
                # Change 'be' to 'he' and lookup that 3-word form.
                behe_form = re.sub(hebe, behe, be_form)
                behe_count = self.find_in_dictionary(he_form)
                # At this point we have the two forms and how common they are in
                # hebe_count and behe_count.

                # Improved Golang/PPWB format and values calculation
                if behe_count > 0 and (
                    hebe_count == 0 or behe_count / hebe_count > PARANOID_LEVEL_3WORDS
                ):
                    suspects_count += 1
                    info = f"({behe_count}/{hebe_count} - '{behe}' is seen more frequently than '{hebe}' in this phrase)"

                    # Add this 3-form and its info to the dialog

                    header = f'Query phrase "{hebe_form}" {info}'
                    # Get whole-words slice from original para text that is approximately
                    # centered on the 3-form phrase.
                    _, _, centered_slice = self.get_para_slice(
                       match_obj.start(0), wbs[rec_num]
                    )
                    self.add_to_dialog(header, centered_slice, checker_dialog)
        return suspects_count
```

cpeel (approved):
Looks great!

srjfoo:
Sorry to be so late to the party -- and needing to ask for clarification, as well.

I brought up one of my test files (which was actually clean of he/be errors) and seeded it with some. Bear in mind that I'm a Mac user, not Windows, and that since I'm not a functional PPer at the moment (though I understand the overall process), I need rather explicit instructions of what I'm supposed to see.

What I actually experienced:
* Seeded he/be errors
* Ran jeebies
* Was presented with two queries
* * Both queries have two lines. One is descriptive, the other shows the questionable line
* * Control-click or track-pad-mimicking of right-click or right-click with a track-ball (I can dig out a real mouse if I need to), all with or without shift, removes one line from the list (should it be removing both lines of an entry instead of just one?).
* No other key combination that I could come up with that excluded the control-click group of things did anything other than selecting a single line.

Anyway help, please! Is this what I should be seeing, or is it Yet Another Difference Between Windows and Mac? (Has it been tested on Linux?)

windymilla:
> * * Both queries have two lines. One is descriptive, the other shows the questionable line
> * * Control-click or track-pad-mimicking of right-click or right-click with a track-ball (I can dig out a real mouse if I need to), all with or without shift, removes one line from the list (should it be removing both lines of an entry instead of just one?).
> * No other key combination that I could come up with that excluded the control-click group of things did anything other than selecting a single line.
> 
> Anyway help, please! Is this what I should be seeing, or is it Yet Another Difference Between Windows and Mac? (Has it been tested on Linux?)

That is what I would expect @srjfoo, though @G4OEU may have further comments
This first version of jeebies is based on Roger's ppwb Go code - the next update will use ideas from "ye olde original" C code version. This version doesn't have the info it needs to let the user click on an error and jump to it, but the next version will. 
If it is necessary (and I don't know that it will be) I will add a general feature to the checker dialog so that multiple-line messages can be treated as a single error for the purposes of Ctrl/Cmd-clicking and Right-clicking.
So, IMO, if it is reporting the he/be errors you expect on the Mac (the same as ppwb's pptxt tool, but not necessarily including all the queries that the C version catches) then I think it's OK.

G4OEU:
I can understand why @srjfoo raised that concern. The next version of Jeebies will avoid that issue as each queried he/be phrase will be presented as a single clickable line. The response from @windymilla correctly summarises the expected behaviour in the current version and the reasons behind it.

G4OEU:
Here is the report from the new Jeebies when the words 'he' and 'be' occur in the file but are not involved in any 'suspect' phrases.

jeebies_new_2

G4OEU:
A feature of the PPWB version of Jeebies is that if neither the 'word1 he/be word2' phrase nor its converse 'word1 be/he word2' phrase appears in the corpus of DP texts then its is never flagged. For example neither the 'be' phrase in the sentence “My name’s Cameron; it’ll maybe be for me.” nor its 'he' converse phrase appears in the corpus so it won't be queried. I think for that very reason it should be queried.

What do you think?

srjfoo:
> A feature of the PPWB version of Jeebies (and the current and new Python versions) is that if neither a 'word1 he/be word2' phrase nor its converse 'word1 be/he word2' phrase appears in the corpus of DP texts then its is never queried. For example neither the 'be' phrase in the sentence “My name’s Cameron; it’ll maybe _be_ for me.” nor its 'he' converse phrase appears in the corpus so the 'be' phrase won't be queried. I think for that very reason it should be queried in the report.
> 
> What do you think?

I agree. Just because it doesn't fit any of the available patterns is not a good reason for assuming it's correct and not querying it. (Assuming I'm correctly understanding your question 😁 )

windymilla:
> What do you think?

It would be necessary to check how many false positives this would cause, possibly a lot if it relies on the same 2 or 3 word phrase occurring in the DP texts that were used to gather stats. I don't know how that was done - would just one occurrence in the DP texts show up in the hebe.txt file? 

If we decided it was useful, I think it would almost certainly have to be controlled by a "paranoid" (or similar) tickbox.

Since this new version will be in a new PR after this one is merged, I'm happy to advise on how to implement such a tickbox - there are some in the (not yet merged) Word Frequency checker tool.

srjfoo:
> > * * Both queries have two lines. One is descriptive, the other shows the questionable line
> > * * Control-click or track-pad-mimicking of right-click or right-click with a track-ball (I can dig out a real mouse if I need to), all with or without shift, removes one line from the list (should it be removing both lines of an entry instead of just one?).
> > * No other key combination that I could come up with that excluded the control-click group of things did anything other than selecting a single line.
> > 
> > Anyway help, please! Is this what I should be seeing, or is it Yet Another Difference Between Windows and Mac? (Has it been tested on Linux?)
> 
> That is what I would expect @srjfoo, though @G4OEU may have further comments This first version of jeebies is based on Roger's ppwb Go code - the next update will use ideas from "ye olde original" C code version. This version doesn't have the info it needs to let the user click on an error and jump to it, but the next version will. If it is necessary (and I don't know that it will be) I will add a general feature to the checker dialog so that multiple-line messages can be treated as a single error for the purposes of Ctrl/Cmd-clicking and Right-clicking. So, IMO, if it is reporting the he/be errors you expect on the Mac (the same as ppwb's pptxt tool, but not necessarily including all the queries that the C version catches) then I think it's OK.

I picked a comment at random to reply to, but I'll try to respond to several, probably.

Thanks -- I thought that it was broken for me. If the current version doesn't implement the left-click doing anything, I won't worry about it. If the only thing that is supposed to work is the right-click/control-click to remove something from the list, it definitely does that.

I didn't understand that the ratio was a ratio, and even now, knowing that it's a ratio, I'm unclear where the values for the ratio come from? The project text? The jeebies data? It was pretty obvious that it wasn't line number data, so I'm guessing that if it's kept, it should be explained in the documentation.

From the screenshot of the next iteration, I like the he/be being colored (would it be better to highlight it as in other tools?). Bear in mind, when coloring things, that on Macs, the theme color follows the OS. Since I keep my system on the dark theme, that means I have white text on dark grey.

@G4OEU, from everything you and @windymilla have said about this iteration, I'm seeing the right things, so I'll go ahead and get this approved.

windymilla:
@G4OEU, it may be worth you adding a new issue for further discussion of the new jeebies, since this PR will be merged & closed fairly soon.
It would be worth discussing the points you have made and maybe others:

- whether to show the ratio in the error message
- whether to include phrases that aren't in the DP corpus
- whether that should be controlled by a tickbox
- whether we need the header that explains how jeebies works in the jeebies output (or just in the manual)
- etc
-

srjfoo (approved):

windymilla:
> From the screenshot of the next iteration, I like the he/be being colored (would it be better to highlight it as in other tools?). Bear in mind, when coloring things, that on Macs, the theme color follows the OS. Since I keep my system on the dark theme, that means I have white text on dark grey.

I think the he/be in the new version are being treated the same as the highlighting in other tools - I thought we'd already got colors that were acceptable in your dark theme via a previous PR

srjfoo:
> > From the screenshot of the next iteration, I like the he/be being colored (would it be better to highlight it as in other tools?). Bear in mind, when coloring things, that on Macs, the theme color follows the OS. Since I keep my system on the dark theme, that means I have white text on dark grey.
> 
> I think the he/be in the new version are being treated the same as the highlighting in other tools - I thought we'd already got colors that were acceptable in your dark theme via a previous PR

That shows how good my memory is -- I was remembering the text being highlighted in other tools, not the text, itself, being in color.

I mentioned it mostly because the screenshot has a white background and thought that @G4OEU might not be aware that the dark background might be an issue.

I'm finding the current highlighting okay, but not optimal; it's a light grey-blue (which follows my OS theme, and maybe should be changed. It doesn't really stand out well enough to be immediately visible. That being said, I don't know whether to advocate for having the highlight color customizable or not.

windymilla:
>  That being said, I don't know whether to advocate for having the highlight color customizable or not.

I'm gonna resist that as long as I can :)
Since the modern ttk (as opposed to tk) widgets are "themed" (hence the extra t) what we will probably end up with is some sort of theming where different colors can be customized, or at least where we offer a few different themes (which can include other differences, not just color).

srjfoo:
> I'm gonna resist that as long as I can :)

I know 🤣

Issue #179, opened 2024-03-24
Suggestion: Persistent range option for Search-and-Replace


charliehoward4dp:
In GG1, once a range has been selected and a Search or S&R has been performed, the range is de-selected. There are times when several operations within that range need to be done (e.g., for a long table), so a way of reusing it without having to go through the selection step repeatedly would save time and improve accuracy.

If the most recent range start-end is always stored temporarily, and a "re-use selected range" checkbox was added to the Dialog, that would suffice. That checkbox would be turned off automatically, for safety, after each S&R.

tangledhelix:
An alternate approach might be to add a way to re-select the most recent selection. As an example, Vim and its derivatives allow for this with the `gv` command.

windymilla:
GG1 actually has this feature (see the **bold sentence below**) though I had forgotten about it until @charliehoward4dp entered this issue.

In the status bar, the selected range is displayed. This is either the start & end point of the selection or the number of rows/cols selected (for a column/block selection) - you left-click the label to toggle between those.
If you double-left-click, it pops the selection dialog - I've never used it but maybe someone does.
**If you right-click, it restores the previous selection.**
If you shift-right-click, it restores a column/block selection.

charliehoward4dp:
Thank you, Nigel. Now that I look, that is explained in the manual, and always has been. Maybe I should be the manual sometime.

windymilla:
I'm labelling this "future feature" because it appears virtually no-one knew about its presence in GG1, so it isn't really core. We should do it though.

tangledhelix:
This same "previous selection" behavior is available in GG1 (as a button) in the "Highlight Character or Regex" dialog.

windymilla:
I've started work on this one

charliehoward4dp:
The implementation of this in GG1 does not always remember the most recent selection (I have not been able to find a pattern for this lapse of memory), and in addition, my suggestion is for something that will be specifically remembered by S&R. That way, if someone highlights a big table and must make a series of changes to it, but sees a small error within the table when part-way through that series, the error can be selected and fixed, without losing the S&R selection of the entire table.

windymilla:
> This same "previous selection" behavior is available in GG1 (as a button) in the "Highlight Character or Regex" dialog.

[...]

PR #190, opened 2024-04-07
Status bar improvements

windymilla:
1. Put "ordinal" at the end so its length-changes don't make everything else jump around.
2. Add "current selection" label to status bar
3. Show range, e.g. "12.5-13.8", for regular selection.
4. Show rows/cols, e.g. "R:3 C:6", for column selection.
5. Click on label to revert to previous selection.

Due to the issue of drag selection consisting of several selections of increasing length, and we don't want these to fill the previous selection slot, we only store previous selection if both the start and end of the selection are different to the previous ones. This will work in almost every real-life circumstance, and I can't find a reasonable way to avoid it. I investigated whether it was possible to detect if the mouse button was pressed (not easily without interfering with existing click/double-click functionality) and if it was possible to override every method of doing a selection (possible, but lots of code to write, attempting to mimic native widget behavior, which is not written in Python) and if it was possible to use the `<<Selection>>` event (no, because of drag selection generating several of these).

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -479,6 +479,11 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             "page label", "<ButtonRelease-3>", self.show_page_details_dialog
         )
 
+        the_statusbar.add("languages label", text="Lang: ")
+        the_statusbar.add_binding(
+            "languages label", "<ButtonRelease-1>", self.file.set_languages
+        )
+
         def ordinal_str() -> str:
             """Format ordinal of char at current insert index for statusbar."""
             char = maintext().get(maintext().get_insert_index().index())
@@ -491,11 +496,6 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             return f"U+{ord(char):04x}: {name}"
 
         the_statusbar.add("ordinal", update=ordinal_str)
-
-        the_statusbar.add("languages label", text="Lang: ")
-        the_statusbar.add_binding(
-            "languages label", "<ButtonRelease-1>", self.file.set_languages
-        )
 
     def logging_init(self) -> None:
         """Set up basic logger until GUI is ready."""
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -479,6 +479,27 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             "page label", "<ButtonRelease-3>", self.show_page_details_dialog
         )
 
+        def selection_str() -> str:
+            """Format current selection range for statusbar.
+
+            Returns:
+                "Start-End" for a regualar selection, and "R:rows C:cols" for column selection.
+            """
+            ranges = maintext().selected_ranges()
+            maintext().save_selection_ranges()
+            if not ranges:
+                return "No selection"
+            if len(ranges) == 1:
+                return f"{ranges[0].start.index()}-{ranges[-1].end.index()}"
+            row_diff = abs(ranges[-1].end.row - ranges[0].start.row) + 1
+            col_diff = abs(ranges[-1].end.col - ranges[0].start.col)
+            return f"R:{row_diff} C:{col_diff}"
+
+        the_statusbar.add("selection", update=selection_str, width=16)
+        the_statusbar.add_binding(
+            "selection", "<ButtonRelease-1>", maintext().restore_selection_ranges
+        )
+
         the_statusbar.add("languages label", text="Lang: ")
         the_statusbar.add_binding(
             "languages label", "<ButtonRelease-1>", self.file.set_languages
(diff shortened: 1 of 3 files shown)


cpeel (approved):
Nice improvements.

windymilla:
Fixes #179

windymilla:
> > 1. Put "ordinal" at the end so its length-changes don't make everything else jump around.
> > 2. Add "current selection" label to status bar
> 
> General comment. If a PPer chooses to un-dock the images, there's no way that the status bar is going to be long enough. It doesn't wrap and it doesn't scroll. In the book I have open right now, testing that, I can't see any content in the status bar past `Lang: en`.

I'm not sure I understand this. If the main window is too narrow, you won't see all the status bar (as in GG1, NP++ and Visual Studio). Are you saying that when you undock the image and have it floating to the right of your main window, your main window is narrower, so you can't see everything. I think that's just too bad. As far as I can see, the status bar buttons take up no more width than their GG1 equivalents. I don't think wrapping/scrolling is an option for the status bar.

> > 5. Click on label to revert to previous selection.
> 
> That is, it's a toggle. Clicking it again goes back to th

Yes - if you use it to go to the previous selection, then what was your selection is now the previous, so it toggles between the last two. In GG1, you can only go back to previous selection (right-click selection label in status bar) if you currently have no selection, not if you've made a new selection. If that would be clearer or more helpful, I could implement that instead for GG2. I thought what I've done is more flexible, but maybe you think more confusing?

> > Due to the issue of drag selection consisting of several selections of increasing length, and we don't want these to fill the previous selection slot, we only store previous selection if both the start and end of the selection are different to the previous ones. This will work in almost every real-life circumstance, and I can't find a reasonable way to avoid it. I investigated whether it was possible to detect if the mouse button was pressed (not easily without interfering with existing click/double-click functionality) and if it was possible to override every method of doing a selection (possible, but lots of code to write, attempting to mimic native widget behavior, which is not written in Python) and if it was possible to use the `<<Selection>>` event (no, because of drag selection generating several of these).
> 
> Does this have to do with making a selection and then dragging it to a different place in the document? If not, I'm not understanding. If so, it's not working on my Mac.

No, it's nothing to do with dragging a selection. Referring back to the previous point on "click to revert to previous selection", if I select "1.1-2.3", then select "4.5-4.8", then clicking the label toggles between those two selections. However, if I now select "4.5-4.10" (note that the start point of the new selection is still 4.5, like the previous one - only the end point has changed) it won't be stored as a "new" selection so you can't toggle between them. The key point is that for something to be considered a "new" selection, it has to have both the start and end different to the previous selection, rather than just being an extension or contraction of the previous selection by adjusting one end of it. I think it will turn out to be a choice between the GG1 behavior mentioned above (only revert if current selection is empty) or this restriction on what a "new" selection is. I think I prefer what we have, but could be persuaded otherwise. Hope that clarifies things.

srjfoo:
> ... Are you saying that when you undock the image and have it floating to the right of your main window, your main window is narrower, so you can't see everything. I think that's just too bad. As far as I can see, the status bar buttons take up no more width than their GG1 equivalents. I don't think wrapping/scrolling is an option for the status bar.

And a good reason to keep the image docked. No, I don't think it takes up any more room than it does in GG1, and to be fair, it's also an issue in BBEdit if the window is too narrow (many's the time I wish it would scroll, because I don't maximize my windows). As with several things, I mentioned it to make sure that was intentional, and something that PPers will just have to deal with.

> Yes - if you use it to go to the previous selection, then what was your selection is now the previous, so it toggles between the last two. In GG1, you can only go back to previous selection (right-click selection label in status bar) if you currently have no selection, not if you've made a new selection. If that would be clearer or more helpful, I could implement that instead for GG2. I thought what I've done is more flexible, but maybe you think more confusing?

No, I like it the way it is in GG2. I obviously didn't finish my thought before I submitted the comment 🤣 . It was an observation, not a complaint!


> No, it's nothing to do with dragging a selection. Referring back to the previous point on "click to revert to previous selection", if I select "1.1-2.3", then select "4.5-4.8", then clicking the label toggles between those two selections. However, if I now select "4.5-4.10" (note that the start point of the new selection is still 4.5, like the previous one - only the end point has changed) it won't be stored as a "new" selection so you can't toggle between them. The key point is that for something to be considered a "new" selection, it has to have both the start and end different to the previous selection, rather than just being an extension or contraction of the previous selection by adjusting one end of it. I think it will turn out to be a choice between the GG1 behavior mentioned above (only revert if current selection is empty) or this restriction on what a "new" selection is. I think I prefer what we have, but could be persuaded otherwise. Hope that clarifies things.

Yes, that clarifies things. And I'm fine with that. It's unfortunate about both ends having to be different, but so be it?

srjfoo (approved):

PR #191, opened 2024-04-08
Implement generic dictionary

G4OEU:
Three files changed:
application.py
project_dict.py
tools/pptxt.py


windymilla (approved):
Code changes look good to me - thanks for the small clarifications to the comments, as well as using `report_all_occurrences_on_line` to reduce code duplication. 
I think the generic dictionary name will be a definite improvement over GG1 (and what I had done in GG2).
Tested adding word ending in "tb" to project dictionary and confirmed that pptxt no longer reported it as suspicious.

cpeel (approved):

src/guiguts/tools/pptxt.py
```diff
@@ -1271,32 +1205,32 @@ def unicode_numeric_character_check() -> None:
     )
 
     # Courtesy limit if user uploads fpgen source, etc.
```
srjfoo, line 1207: Double-checking that `fpgen` is what's meant here. Is there a more general description?

src/guiguts/tools/pptxt.py
```diff
@@ -1271,32 +1205,32 @@ def unicode_numeric_character_check() -> None:
...
+    courtesy_limit = 5
     numeric_char_reference_count = 0
+    lines_with_numeric_char_references = 0
 
     line_number = 1
     regx = r"(&#[0-9]{1,4};|&#x[0-9a-fA-F]{1,4};)"
```
srjfoo, line 1213: Shouldn't the decimal digits be `{1,5}`? `9999` decimal is considerably less than `ffff` hexadecimal (which I think covers the BMP?). (Though maybe I'm misunderstanding what's intended here.)

windymilla:
I'll leave @G4OEU to answer your specific questions, @srjfoo.
I didn't give you enough testing notes, so here are some:

The first change this PR makes is that instead of using `my_file_dict.json` as the project dictionary for `my_file.txt`, it will use the generic name `project_dict.json` (in the same folder as the text file of course. This means that a project dictionary can be used with all the files in the project, regardless of their name, e.g. "my_file.txt", "my_file_spellchecked.txt", "my_html_version.txt", etc.
To test that, you can do a spell check then add words to the project dictionary, and check they appear in `project_dict.json`.

The second change is that some of pptxt's checks make exceptions for words that are in the project dict. The example I tested was I entered a word that ended in "tb", which got flagged by pptxt. Then I did a spellcheck, which also flagged my word, then I added it to the project dict.. Then I re-ran pptxt to check that it no longer flagged that word.

src/guiguts/tools/pptxt.py
G4OEU: It should be and I will fix that as well. Thanks for pointing it out.

Issue #180, opened 2024-03-25
Clicking on highlighted line in a report window does not bring GG main window to screen if it's minimised


G4OEU:
I'm not sure what should be the behaviour when the GG main window is minimised and you click on a highlighted line in a report.

If the main GG window is minimised (not displayed on the computer screen) then I would expect the main GG window to be at least restored and probably brought to the fore. 

I note that if both the report window and GG main window are displayed, clicking on a highlighted line in the report brings the report window to the fore and highlights the appropriate part of the file line in the main GG window. In this situation I expected that clicking on a highlighted report line would bring the main GG window to the fore with the file line highlighted.

windymilla:
I definitely agree with the first part - the main window should be un-minimized.

The second issue is complicated by the differing behavior of different platforms. @srjfoo, @tangledhelix and I discussed this after my first draft of the checkerdialog, which did force focus to the main window. See #157, particularly @tangledhelix comment beginning "My test results". One problem, as I understand it, is that on macOS if the focus transfers to the main window so the focus is not on the checker dialog, then a right click on the checker dialog (to remove that error) gets intercepted by the operating system.  Mac users have to put up with some similar issues in GG1 - the way that focus and mouse-clicks are processed is different from the way it happens on Windows.

I can think of some options:

1. Make the focus-switching behavior OS-specific
2. Make the focus-switching behavior a user preference (probably defaulting differently for different OSes)
3. Investigate using right-button release instead of press, in case that makes a difference. 
4. Make it work like GG1 (and say "tough luck" to Mac users)
5. Leave it like it is in GG2 (and say "touch luck" to WIndows users)
6. Investigate what happens on Linux, whichever of the above is chosen.

PR #192, opened 2024-04-08
Icon focus

windymilla:
If entry in checker dialog is selected, then ensure main window is deiconified (but not force-focused)

Because "deiconify" does not work on (some) X-based systems, and may not be needed on macOS, make the iconify/deiconify hack when the program starts, specific to Windows.

Also, when clicking on a checker dialog message, macOS users would expect focus to remain on the checker dialog, but other users would find it helpful to focus on the main window, so make that OS-specific too.

Fixes [#180](https://github.com/DistributedProofreaders/guiguts-py/issues/180)

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -382,7 +382,7 @@ class CheckerDialog(ToplevelDialog):
             start = maintext().index(self._mark_from_rowcol(entry.text_range.start))
             end = maintext().index(self._mark_from_rowcol(entry.text_range.end))
             maintext().do_select(IndexRange(start, end))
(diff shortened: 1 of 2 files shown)


cpeel (approved):

windymilla:
Testing notes for @srjfoo:
1. On startup, you should not see the screen briefly minimized then restored, as you've seen previously.
2. After running a checker tool, e.g. spell check, clicking on a query message in the checker dialog should take you to the correct location. 
3. The focus should not be forced to the main window on macOS.
4. If the main window was minimized, but the checker dialog was not, then the main window should be restored when the error message is clicked.

srjfoo (approved):

Issue #193, opened 2024-04-08
Checkers improvement - sort options


windymilla:
Generic sort options for (almost) all checker types...

1. Sort by row.col number (the default)
2. Sort by "type" - may mean alphabetical sort on error message, or on highlighted part of error message, or maybe on an additional, hidden, attribute of the error message. Meaning of "type" may depend on which checker it is.
3. Show only one "type" or all "types" - maybe a dropdown of checkboxes (or a dialog) for all the types available, (like bookloupe's display options dialog). Does it need to be persistent (in prefs) - if so, how to handle types that haven't occurred in a dialog yet? Maybe a dialog should know a list of the possible error types it might display?
4. Remove all entries that are the same type as the selected message? (May be just an extension of "Remove All")
5. Revisit "process & remove all" - it processes the current entry, then removes all matching entries - that makes sense for spelling, where processing means "add to dict", so you want to remove all identical entries. However, for jeebies, or Basic Fixup, you want it to loop for all matching items, processing and removing each one. Difference is between "(process & remove) all" and "process & (remove all)"
6. Some types of checker may not want "process & remove all"
7. There's going to have to be some cleverer idea of what "matching" means. Different types of checker will need different "match" functions. 

May be a better word than "type".

PR #194, opened 2024-04-09
Add Tools->Basic Fixup

windymilla:
Slightly different to GG1, this displays all the "errors", and user Ctrl/Cmd-clicks error to "process" it, i.e. fix it. In GG1, it just silently makes the changes (not necessarily desirable).

Future boosts to the checker dialog (see issue #193) may make this tool better.

Another possible future improvement would be a button to say "Make all suggested changes" after user has reviewed and removed (via right-click) any dodgy ones).


src/guiguts/misc_tools.py
```diff
@@ -0,0 +1,130 @@
...
+
+logger = logging.getLogger(__package__)
+
+BLOCK_TYPES = "[$*XxFf]"
+POEM_TYPES = "[Pp]"
+
+
+def process_fixup(checker_entry: CheckerEntry) -> None:
+    """Process the fixup error."""
+    if checker_entry.text_range:
```
cpeel, line 18: If I've put an entire function within an if I like to negate the condition and return, then fall through to the actual code so I don't have to indent the whole function. Clearly just a personal preference and what you have works great, but I noticed it and thought I'd mention it.

cpeel (approved):

src/guiguts/misc_tools.py
windymilla: Oh yes, thanks! I found a couple other small things while fixing that too!

windymilla:
Testing notes for @srjfoo 
Should be suitable for running on a CTF.
Checks things like trailing spaces, `multiple      spaces` between words, spaces before punctuation, etc.
Doesn't check lines inside block markup `/*`, `/$`, etc
Ignore poetry line numbers (inside `/P` markup) as far as spacing is concerned, e.g.
```
/P
    Here is the first line of my poem        1
P/
It covers most of what the GG1 version checks and fixes

srjfoo:
Not sure whether or not these need to be fixed, but the two things that jumped out at me were that it would be nice to

1. Omit proofers' notes from basic fixup, as the no-wrap is omitted. Apparently a some proofers like to put multiple sequential spaces inside their notes.
2. It would be good to have some way of indicating "I've seen this before, it's okay, please don't show it to me again." One of my test projects had about 20 instances of `whilst`, which basic fixup identifed as a possible `1/l` error. (Interestingly, the same word showed up once in another book, and was identified as a scanno.)

On macOS, `cmd`+left click "fixed" the error. `ctrl`+left-click, right-click on a mouse, and two-finger-right-click-alias all removed the line from the list.

Are all of these expected behavior?

windymilla:
> Not sure whether or not these need to be fixed, but the two things that jumped out at me were that it would be nice to
> 
> 1. Omit proofers' notes from basic fixup, as the no-wrap is omitted. Apparently a some proofers like to put multiple sequential spaces inside their notes.
> 2. It would be good to have some way of indicating "I've seen this before, it's okay, please don't show it to me again." One of my test projects had about 20 instances of `whilst`, which basic fixup identifed as a possible `1/l` error. (Interestingly, the same word showed up once in another book, and was identified as a scanno.)
> 
> On macOS, `cmd`+left click "fixed" the error. `ctrl`+left-click, right-click on a mouse, and two-finger-right-click-alias all removed the line from the list.
> 
> Are all of these expected behavior?

All of that is expected. I've added a bit more to issue #193, which should deal with point 2.

I'm not going to do anything about point 1 for now.

windymilla:
Latest commit should stop if reporting `whilst` - good catch @srjfoo

srjfoo (approved):

PR #202, opened 2024-04-14
Implement tooltips

windymilla:
1. Default tooltip position is centered below widget
2. If tooltip would extend off the left or right of screen, it is kept on screen by shifting it right or left.
3. If tooltip would extend below bottom of the mainwindow or dialog that contains the widget, then tooltip is centered above widget instead of below.
4. Accepts string which can contain newline characters for multi-line tooltip.
5. Added to status bar items and to the page details dialog which is accessed by right-clicking the "Lbl" button in the status bar.
6. Other tooltips will be added as new features are added.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -464,7 +464,11 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             row, col = maintext().get_insert_index().rowcol()
             return f"L:{row} C:{col}"
 
-        the_statusbar.add("rowcol", update=rowcol_str)
+        the_statusbar.add(
+            "rowcol",
+            tooltip="Left click: Go to line\nRight click: Toggle line numbers",
+            update=rowcol_str,
+        )
         the_statusbar.add_binding("rowcol", "ButtonRelease-1", self.file.goto_line)
         the_statusbar.add_binding(
             "rowcol", "ButtonRelease-3", maintext().toggle_line_numbers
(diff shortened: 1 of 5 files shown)


srjfoo:
Screenshots:

---

Bottom of main window, but not bottom of screen:

image

---

Main window shoved all the way to the right edge of the screen:

image

---

Main window shoved all the way to the bottom:

image

---

windymilla:
Why does literally nothing work on macOS the same as on Windows?😠

windymilla:
image

windymilla:
image
On right of screen - tooltip gets pushed left so it's all visible

windymilla:
For @srjfoo, I've added debug print statements when the tooltip is popped. Should be something like this each time a tooltip is popped:
```
tooltip: geom: +1394+742, size: 142x23
widget: pos: 1432,1432, size: 109x25
widget_tl: pos: 926,926, size: 800x393
screen: 1536x960
```
Please could you try to reproduce the first bug, and then copy/paste the debug output (in Slack if easier). Thanks

windymilla:
Possibly helpful stuff:

- https://stackoverflow.com/questions/24377988/tkinter-on-ubuntu-14-02-reports-width-as-sum-of-both-monitors (especially "root.winfo_width() equals the width of the current display on both X and windows.")
- https://stackoverflow.com/questions/3129322/how-do-i-get-monitor-resolution-in-python
- https://stackoverflow.com/questions/30312875/tkinter-winfo-screenwidth-when-used-with-dual-monitors (especially mention of https://pypi.org/project/screeninfo/)

srjfoo:
> Why does literally nothing work on macOS the same as on Windows?😠 (That may be a bit harsh - it's quite possible it's a dual monitor issue, rather than a macOS one - is the issue you're seeing only happening when using 2 monitors, @srjfoo ?)

Oh, good question. I've definitely been using two monitors recently -- it helps spread things out. But that's a distinct possibility. I'lll get back to you on that. That being said, it's also a good thing to be aware of, since a lot of people do use a second monitor.

windymilla:
Latest commit still has debug print in. @srjfoo - please could you check if tooltip size prints out sensibly (approx 200x25 not 1x1). Also changed tooltip in the page label config to be near the cursor instead of at the bottom, which was weird.
Assuming 1x1 issue above is fixed, tooltips should now remain within the main window, regardless of dual monitors - that doesn't apply to the new mouse-positioned ones, which (deliberately) could go outside the window.

srjfoo (approved):

windymilla:
Extended testing complete, and temporary debug prints remove, so hopefully ready for code review now - thanks @cpeel

windymilla:
Rebased

PR #207, opened 2024-04-18
Centralize most of the mouse binding

windymilla:
In particular this is to consolidate the need for Mouse-3 to be interpreted as Mouse-2 and Control-Mouse-1 on Macs. Previously, these were done in different places, and not all places did both bindings.

One place that may not have had support for Control+1 (one-button mouse equivalent of right-button) is popping the context menu in text widgets on Macs. May have got away with it through default Mac behavior.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -465,46 +465,42 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             return f"L:{row} C:{col}"
 
         the_statusbar.add("rowcol", update=rowcol_str)
-        the_statusbar.add_binding("rowcol", "<ButtonRelease-1>", self.file.goto_line)
+        the_statusbar.add_binding("rowcol", "ButtonRelease-1", self.file.goto_line)
         the_statusbar.add_binding(
-            "rowcol", "<ButtonRelease-3>", maintext().toggle_line_numbers
+            "rowcol", "ButtonRelease-3", maintext().toggle_line_numbers
         )
 
         the_statusbar.add(
             "img",
             update=lambda: "Img: " + self.file.get_current_image_name(),
         )
-        the_statusbar.add_binding("img", "<ButtonRelease-1>", self.file.goto_image)
+        the_statusbar.add_binding("img", "ButtonRelease-1", self.file.goto_image)
 
         the_statusbar.add("prev img", text="<", width=1)
-        the_statusbar.add_binding("prev img", "<ButtonRelease-1>", self.file.prev_page)
+        the_statusbar.add_binding("prev img", "ButtonRelease-1", self.file.prev_page)
 
         the_statusbar.add("see img", text="See Img", width=9)
         the_statusbar.add_binding(
             "see img",
-            "<ButtonRelease-1>",
+            "ButtonRelease-1",
             self.show_image,
         )
         the_statusbar.add_binding(
-            "see img", "<ButtonRelease-3>", self.file.choose_image_dir
+            "see img", "ButtonRelease-3", self.file.choose_image_dir
         )
-        the_statusbar.add_binding(
-            "see img", "<Double-Button-1>", self.toggle_auto_image
-        )
+        the_statusbar.add_binding("see img", "Double-Button-1", self.toggle_auto_image)
 
         the_statusbar.add("next img", text=">", width=1)
-        the_statusbar.add_binding("next img", "<ButtonRelease-1>", self.file.next_page)
+        the_statusbar.add_binding("next img", "ButtonRelease-1", self.file.next_page)
 
         the_statusbar.add(
             "page label",
             text="Lbl: ",
             update=lambda: "Lbl: " + self.file.get_current_page_label(),
         )
+        the_statusbar.add_binding("page label", "ButtonRelease-1", self.file.goto_page)
         the_statusbar.add_binding(
-            "page label", "<ButtonRelease-1>", self.file.goto_page
-        )
-        the_statusbar.add_binding(
-            "page label", "<ButtonRelease-3>", self.show_page_details_dialog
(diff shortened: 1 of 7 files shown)


windymilla:
Testing notes - check that right mouse (and control-left-mouse) does what is expected in various places, e.g. context menu in text widgets, right-click "See Img" in status bar, etc.

cpeel (approved):
Very nice centralization of that code 👍

windymilla:
Further testing notes:
1. With the consolidation in this commit, from now on whenever documentation or comments refer to "Right mouse", that will mean "right hand button on a 3-button mouse (Windows/Linux) or a 2-button mouse (Mac), and any alternative OS-specific or hardware-specific methods of getting a right mouse click, such as Ctrl-left-mouse on a Mac, 2-finger tap on trackpad, etc." Prior to this commit, sometimes when right-mouse was intended, one of the above alternatives might not have been implemented. 
2. Also, the phrase "Cmd/Ctrl" means "Cmd" on a Mac, "Ctrl" elsewhere. So, for example, "Cmd/Ctrl-left mouse" does **_not_** mean "Cmd" with the "Ctrl-left mouse" method of right clicking on a Mac - it means "Cmd-left mouse" on a Mac, and "Ctrl-left mouse" elsewhere. Since "right mouse" includes all methods of getting a right mouse click, then "Cmd" with "Ctrl-left mouse" would be included in the meaning of the phrase "Cmd/Ctrl-right mouse". Sorry if that's confusing - but "Cmd/Ctrl" seems a good way of shortening "Cmd on a Mac, Ctrl on Linux and Windows". So, if you're a Mac user, just replace "Cmd/Ctrl" with "Cmd" in any descriptions. I don't know what we'll do for the manual, i.e. whether to use a similar shorthand, or whether it needs to be spelt out longhand for every occurrence.
3. Uses of right-mouse in current version of GG:
  a. Main text window - open context menu (cut, copy, paste, select all)
  b. Some read-only text widgets (currently just the message log) - open context menu (copy, select all). A read-only text widget is also used for Word Frequency and the checker dialogs (Fixup, Spellcheck, PPtxt, Jeebies, Find All), but the context menu is disabled for all of those - see below for whether right mouse does anything in those.
  c. Status bar "L:1 C:0" button - toggle display of line numbers on the left.
  d. Status bar "See Img" button - choose dir that contains png scan files.
  e. Status bar "Lbl:" button - show page details (label config) dialog.
  f. All checker dialogs (Fixup, Spellcheck, PPtxt, Jeebies, Find All) - remove the clicked entry (note that on Macs, you may need ensure focus is on the dialog before right-clicking) and select the next entry.
  g. All checker dialogs (Fixup, Spellcheck, PPtxt, Jeebies, Find All) - Shift with right-mouse removes the clicked entry and all other entries that "match" it - at the moment "match" means the text must be identical, though the line.col may be different, so this is most likely useful in Spellcheck to remove (or skip) all occurrences of a "mis-spelling". The "Shift" means do it to "all".
  h. All checker dialogs (Fixup, Spellcheck, PPtxt, Jeebies, Find All) - Cmd/Ctrl with right mouse "processes" the clicked entry and removes it. Note that "Cmd/Ctrl" means on a Mac use "Cmd+right-mouse", on other platforms use "Ctrl+right-mouse". If you are on a Mac and using "Ctrl+left-mouse" to simulate "right-mouse", that means to do "Cmd+right-mouse", you will need to do "Cmd+Ctrl+left-mouse". The meaning of "process" is different across different checker dialogs. In Fixup, it should correct the reported error, e.g. replace multiple spaces with a single space, or change `lst` to `1st`. In Spellcheck, it adds the word to the project dictionary. In PPtxt, Jeebies and Find All, "process" does nothing, i.e. it's the same as a plain click without Cmd/Ctrl. For Jeebies at least, "process" will be implemented in a later version (change "he" to "be" or "be" to "he").
  i. Using Shift and Cmd/Ctrl with right mouse has an ambiguous meaning: either "process" this error and all similar errors and remove them, or "process and remove" this error and all similar errors, i.e. do the similar errors get processed or not? Currently, they don't, but I am planning to change it so they do (at least to trial that with users)

Summary of checker behavior:

- Left click to "go to" the error
- RIght click to "remove" the error (& go to next)
- Cmd/Ctrl with one of the above to also "process" the error
- Shift with one of the above to also do "all" similar errors (see above note `3i` about ambiguity of Shift with Cmd/Ctrl)

For this PR, I'm not really interested in whether we think right-mouse is the best button to use for particular uses, nor whether there's self-inconsistency in the intended behavior described above. I just want to be sure that the program does behave as described above. I've added issue #210 to suggest some further consistency work for the status bar click types, and issue #193 makes reference to resolving the ambiguity described in `3i` above.

windymilla:
Rebased

srjfoo:
On an M1 MBP running Sonoma 14.4.1, with second monitor plugged in:

|   |right-click|2-finger-tap|ctrl+left click|
|---|:---:|:---:|:---:|
|See Img--File->Open|✅|✅|✅|
|Lbl:--Config Pg Labels|✅|✅|✅|
|Basic Fixup--skip/remove from list|✅|✅|✅|
|Word Freq.--?|*** |*** |***|
|Spelling Check--skip/remove from list|✅|✅|✅|
|PPtxt--skip/remove from list|✅|✅|✅|
Jeebies--skip/remove from list|✅|✅|✅|

`***` I can't tell what it's supposed to do; right-click does nothing; two-finger-tap does nothing; ctrl + left-click works the same as left-click.

---

Global footnote: `esc` closes/cancels each widget window in reverse stacking order, plus macOS dialogs (does not close the root window). `Cmd`+`.` cancels only macOS-level dialogs.

---

Is there something else I should be testing?

srjfoo (approved):

PR #209, opened 2024-04-19
Make `Cmd+w` close dialog/file as expected

windymilla:
On a Mac, user would like `Cmd+w` to close the dialog with focus. If the focus is in the main window, then it should close the currently loaded file, and not close the main window.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -236,6 +236,18 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         if self.file.open_file(filename):
             self.mainwindow.clear_image()
 
+    def close_command(self) -> None:
+        """Close top-level dialog with focus, or close the file if it's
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: Hopefully this is what you had in mind @srjfoo. 
`Cmd+w` should close the dialog with focus. If focus is on the main window, then it continues to do "Close file" as it did previously. This only affects Mac users, since Windows users can only access the function through the menus (not keyboard shortcut), so the focus for them is always in the main window and it continues to close the file as previously.

cpeel (approved):
Code LGTM but I defer to @srjfoo on how it works :D

srjfoo:
> Code LGTM but I defer to @srjfoo on how it works :D

Seems to work splendidly!

srjfoo (approved):

Issue #196, opened 2024-04-11
Add Preferences dialog


windymilla:
The minimum is to be able to turn the audible bell on & off.

Possibly use a Notebook (tabbed) widget to separate the different types of preference (Appearance, Processing, etc).

windymilla:
Things to include:

- Toggle line numbers
- Auto image
- Ordinal name display

PR #215, opened 2024-04-21
Simplify status bar clicking

### LINKED ISSUE #210: Status bar mouse clicks
Try to make use of Cmd/Ctrl and Shift more consistent with use in checker dialogs.

1. Do we really want "double click" for  "See Img" to choose image dir? I think not - maybe "Cmd/Ctrl" to "process" the image dir might make sense, and "Shift" to toggle Auto Img (Shift meaning "all" when used in checker dialogs)
2. Similarly "Cmd/Ctrl" to "process" page labels, i.e. pop the page label config dialog.
3. Would Cmd/Ctrl-left mouse be better than using right mouse where possible, e.g. in "R:1 C:0"? (I think so)
4. Also, do the clicks need to be "button-release", or would a simple button press do?

windymilla:
1. Avoid using right button. Use left button for main feature, and Shift-left for additional feature if needed.
2. Move "Choose image scans dir" to File-->Project menu, so no longer available by clicking status bar. It is popped automatically anyway, if the user tries to look at a scan before it is set up. GG1 has a File-->Project menu button for it too.
3. Add Shift-click to the ordinal button to toggle the display of the character name. Some users find the changing length of the button distracting, and it would take up too much space to make it long enough to fit the longest common character names. Setting is stored in GGprefs file.

Fixes #210

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -29,7 +29,7 @@ from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
 from guiguts.tools.pptxt import pptxt
 from guiguts.tools.jeebies import jeebies_check
-from guiguts.utilities import is_mac
+from guiguts.utilities import is_mac, is_windows
 from guiguts.word_frequency import word_frequency, WFDisplayType, WFSortType
 
 logger = logging.getLogger(__package__)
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes:
I think this aligns with what we discussed: Avoid button 3 where possible, and prefer Shift over Cmd/Ctrl where possible. Also add the functionality into the menus, not have it just hidden away via some obscure click in the status bar. I've added Scan Dir selection to the File-->Project menu, and I've added a comment to #196 to put things like auto-img into the upcoming Preferences dialog.

cpeel (approved):

srjfoo (approved):

PR #216, opened 2024-04-22
Read ISO-8859-1 files from DP correctly

windymilla:
Some text files and good/bad words files may still be ISO-8859-1 rather than UTF-8. Try to read them as UTF-8, and if they fail, read them as ISO-8859-1.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -345,10 +345,14 @@ class MainText(tk.Text):
         Args:
             fname: Name of file to load text from.
         """
-        with open(fname, "r", encoding="utf-8") as fh:
-            self.delete("1.0", tk.END)
-            self.insert(tk.END, fh.read())
-            self.set_modified(False)
+        self.delete("1.0", tk.END)
+        try:
+            with open(fname, "r", encoding="utf-8") as fh:
+                self.insert(tk.END, fh.read())
+        except UnicodeDecodeError:
+            with open(fname, "r", encoding="iso-8859-1") as fh:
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes:
1. Load an ISO-8859-1 text file - should load correctly and characters like degree signs should display correctly.
2. Get a goodwords file, ensure it is ISO-8859-1, and that it has at least one character, like a degree sign in it. Use File-->Project-->Load good/bad words

cpeel (approved):

srjfoo (approved):
Works splendidly.

One question: If a file is detected as not UTF-8, should the user get a warning something like:
> File not UTF-8; ISO-8859-1 assumed: converting to UTF-8

It would probably never happen, but I'm imagining someone trying to open a Shift-JIS file in GG 😁 

(I had the felicity one time of sorting out a file that had a mixture of UTF-8 and Shift-JIS in it.)

windymilla:
> One question: If a file is detected as not UTF-8, should the user get a warning something like:
> 
> > File not UTF-8; ISO-8859-1 assumed: converting to UTF-8

I think it's sufficient that we are supporting utf8 and latin1. That warning message is only really needed for files that are neither utf8 nor latin1, as you say, but would be output for valid latin1 files. Balancing the queries from users about the message when they have a valid latin1 file, against the rarity of loading a Shift-JIS file, for example, I think I'm happy to convert silently. The other checks in GG will soon catch where [unusually-encoded characters get interpreted as Latin1](https://en.wikipedia.org/wiki/Mojibake).

PR #219, opened 2024-04-23
Add a Preferences dialog

windymilla:
1. Menu Edit-->Preferences pops the dialog on Windows/Linux, and the standard application menu Settings entry pops it on a Mac.
2. Currently has 2 tabs (one of which is currently empty)
3. Line numbers, Auto Img, and Bell settings implemented.
4. Existing Preference callback feature utilized and extended to be called whenever preference changes. This means preference can be used to set/query the value, rather than needing a separate variable and/or set/query functions.
5. Should be able to change tabs in the dialog by using Ctrl-tab (maybe Cmd-tab on Macs?) and by using Alt+first letter of tab name (probably not on a Mac).

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,7 +13,7 @@ Checkers
 .. automodule:: checkers
  
 File
-=========
+====
     
 .. automodule:: file
 
@@ -73,9 +73,14 @@ Word Frequency
 .. automodule:: word_frequency
 
 PPtxt
-============
+=====
     
 .. automodule:: tools.pptxt
+ 
+Misc. Dialogs
+=============
+    
+.. automodule:: misc_dialogs
  
 Misc. Tools
 ===========
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -21,6 +21,7 @@ from guiguts.mainwindow import (
     statusbar,
     ErrorHandler,
 )
+from guiguts.misc_dialogs import PreferencesDialog
 from guiguts.misc_tools import basic_fixup_check
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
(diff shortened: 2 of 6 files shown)


windymilla:
Testing notes - since #215 is not yet merged, the "old" status bar clicks are necessary to show/hide the line numbers, and turn auto img on/off. These features should now also be available via the new Prefs dialog.

srjfoo:
> 5\. Should be able to change tabs in the dialog by using Ctrl-tab (maybe Cmd-tab on Macs?) and by using Alt+first letter of tab name (probably not on a Mac).

`Cmd`+`<tab>` is a system level shortcut that cycles among open applications; confirmed that gg2 does not (also should not) override this.

A common (though not universal) Mac keyboard shortcut for cycling among open windows in an app is `Cmd` + \` (correctly cycles among the main text window and any open widgets)

For switching tabs in the prefs pane in this branch on a Mac, `Ctrl`+`tab` and `Cmd`+`Option`+ left or right arrow both work (the latter work to rotate among open tabs in a browser window on Macs -- confirmed on all of FF, Chrome, Opera and Safari). I think something is slightly off, though.
* The first `Ctrl`+`tab` doesn't do anything. Only the second and beyond switches tabs.
* `Cmd`+`Option`+ left or right arrow don't work until after the first successful `Ctrl`+`tab`.
* Unpredictably, the content seems to be disappearing from the tabs, but will appear in the empty boxes if the mouse is moved and is over the widget:

image  image

And no, Alt/Option plus first letter on a Mac (at least given my settings and keyboards) does not switch tabs. The escape key will close the settings dialog, `Cmd`+`,` opens it.

The settings themselves work well. I don't know if the disappearing text is a Mac only issue, but I've seen it in several runs of the program.

srjfoo (approved):

PR #221, opened 2024-04-25
Downgrade invalid PrefKey message

windymilla:
An assertion is too annoying for developers and testers, because if you run a "later" release, it saves a new PrefKey in the Prefs file. Then when you run an "earlier" version, it reports that the new PrefKey it loads from the Prefs file is invalid.

Downgrade it to a warning, which will appear in the terminal, since it happens before the GUI has been initialized.

### DIFF
--- a/src/guiguts/preferences.py
+++ b/src/guiguts/preferences.py
@@ -165,7 +165,7 @@ class Preferences:
                 try:
                     self.dict[PrefKey(key)] = value
                 except ValueError:
-                    assert False, f"'{key}' is not a valid PrefKey"
+                    logger.warning(f"'{key}' is not a valid PrefKey - ignored")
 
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes - two options: 

1. Run GG in a branch that has a new PrefKey and make a change that will cause the PrefKey to be saved, (e.g. in PR #219, open the new Prefs dialog, and changed the Bell settings). Then run in`master` - you should see the assertion. Then run in this PR, and you should see the warning, not the assertion.
2. Manually edit `GGprefs.json` and change one of the keys, e.g. "rootgeometry" to "badgeometry", then compare `master` with this branch.

srjfoo (approved):

cpeel (approved):

PR #223, opened 2024-04-25
Change Page Label Config dialog to modeless

windymilla:
When it's modal, you can't scroll the text up and down to help decide what values you want to set in the dialog.

Change OkCancelDialog to OkApplyCancelDialog which has an Apply button that applies the changes made so far in the dialog. Use this new dialog for Page Label Config.

Note that (like other apps) once you click Apply, those changes are applied and can't be canceled by clicking Cancel. The OK button is equivalent to Apply followed by closing dialog.

(Note: nested modal dialogs are not implemented well in tkinter - when the dialog popped another to ask the page number, the dialog became modeless once the page number was entered.)

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -218,7 +218,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
 
     def show_page_details_dialog(self) -> None:
         """Show the page details display/edit dialog."""
-        PageDetailsDialog(root(), self.file.page_details)
+        PageDetailsDialog.show_dialog(page_details=self.file.page_details)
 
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: Pop the dialog, then use the main text window to scroll up & down to decide what settings to change. If you like, make some changes and Apply them, then go back to the main window and see that they are what you want, then go back to the Page Label dialog and make more changes before Applying/OKing/Canceling them. (Note that Cancel only cancels the latest set of changes made within the dialog - once you apply them to the file, that's it - you can't cancel them)

srjfoo (approved):

PR #224, opened 2024-04-25
Fix "Previous Selection" bug

windymilla:
Logic in the code to store the previous selection did not take into account the case where there was a selection, and then it was cancelled, leaving no selection.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -502,14 +502,19 @@ class MainText(tk.Text):
         """
         ranges = maintext().selected_ranges()
         # Inequality tests below rely on IndexCol/IndexRange having `__eq__` method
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
Expected behavior is that if you have a selection, e.g. the word "dog", and then at a later point you don't have "dog" selected, then clicking "Previous Selection" in the status bar, should re-select "dog" for you. This situation could happen because you make a new selection, e.g. "cat", or you cancel the selection so you have nothing selected. Either way, "previous selection" means "dog". 
Also, if you had "cat" selected, then use "Previous Selection" to re-select "dog", the previous selection is now "cat", so using "Previous Selection" will select "cat" (and previous selection is now "dog", and so on, and so on alternating between the "current" and the "previous" selection).

srjfoo (approved):

PR #225, opened 2024-04-28
Ensure document has a filename before running tool

windymilla:
1. If tool is run on a document that has never been saved, then it does not have a "project folder" in which to put files like the project dictionary. So, check file has a filename before running the tool.
2. This required something I've been avoiding for a while - making the single File object accessible from all the higher level routines. Function `the_file` now returns the single File object (similar to how `maintext` works).
3. Unfortunately file.py imported spell.py in order to reset the current spelling dictionary, which was the wrong way round, in that the file shouldn't know about the spell check tool, but came about because of my previous reluctance to do point 2 above.
4. Now point 2 has been done, the spell checker keeps a record of which languages it has loaded, and resets if it is asked to spell check a file with different languages.
5. The structure makes more sense now (at least to me) - there are just a few objects that are essentially singletons that it is a pain to pass around everywhere, so having a function that returns them seems like the least bad option.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -11,7 +11,7 @@ import unicodedata
 import webbrowser
 
 from guiguts.checkers import CheckerSortType
-from guiguts.file import File, NUM_RECENT_FILES
+from guiguts.file import File, the_file, NUM_RECENT_FILES
 from guiguts.maintext import maintext
 from guiguts.mainwindow import (
     MainWindow,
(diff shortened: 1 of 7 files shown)


windymilla:
@cpeel - requesting review particularly because of points 2 & 5 in the above description. Thanks

windymilla:
@srjfoo testing notes:
Start GG and try to run one of the tools without loading a file (you can type or paste text into the window if you want to) - it should make you save the file before it will let you. You should be able to cancel if you don't want to do that (but then it won't run the tool). Note that once a file is saved (or has been loaded from a file), you don't have to resave it every time you make a change - the tools will run on a "modified" file that hasn't been saved, but they won't run on text that does not have a filename.

cpeel (approved):
> so having a function that returns them seems like the least bad option.

Agree that this is the least-bad approach given the circumstances 👍

srjfoo (approved):

PR #226, opened 2024-04-28
Make PrefsKeys more readable

windymilla:
Add underscores between words to break them up - the long unbroken upper case constants were hard to read.


windymilla:
Trivial variable renaming - should be no noticeable change, @srjfoo, except that your prefs file will have badly named constants when you switch between this release and other releases. Maybe check that one or two things, like Search "regex" flag still gets remembered between runs (of this release, not runs of this release interspersed with runs of other releases). Thanks

srjfoo (approved):

PR #227, opened 2024-04-28
Tidy up when dialog is destroyed

windymilla:
For now, just tooltips need to be deleted when a dialog is destroyed, because the tooltips are Toplevel widgets in their own right, so are not automatically destroyed when the widget they belong to is destroyed.

Register each tooltip of a child of a TopLevelDialog, and when that dialog is destroyed, destroy the tooltips.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -295,6 +295,7 @@ class CheckerDialog(ToplevelDialog):
 
     def reset(self) -> None:
         """Reset dialog and associated structures & marks."""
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: Should be no behavior change - previously tooltips would be created each time a dialog is popped, and then not destroyed (even though they are invisible). Please just check that tooltips seem to work as they did before. Thanks.

cpeel (approved):

srjfoo (approved):

PR #230, opened 2024-05-07
Add Ctrl/Click functionality for error correction in Jeebies.

G4OEU:
1. Ctrl/Left-click toggles highlighted he/be on file line.

2. Fixed an oversight in checkers.py so that start mark now has left gravity,
and the end mark has right gravity.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -380,11 +380,13 @@ class CheckerDialog(ToplevelDialog):
         self.entries.append(entry)
 
         if text_range is not None:
-            maintext().mark_set(
-                self.mark_from_rowcol(text_range.start), text_range.start.index()
+            maintext().set_mark_position(
+                self.mark_from_rowcol(text_range.start),
+                text_range.start,
+                gravity=tk.LEFT,
             )
-            maintext().mark_set(
-                self.mark_from_rowcol(text_range.end), text_range.end.index()
+            maintext().set_mark_position(
+                self.mark_from_rowcol(text_range.end), text_range.end, gravity=tk.RIGHT
             )
 
     def display_entries(self) -> None:
--- a/src/guiguts/tools/jeebies.py
+++ b/src/guiguts/tools/jeebies.py
@@ -8,10 +8,11 @@ import importlib.resources
 import logging
 import regex as re
 
-from guiguts.checkers import CheckerDialog
+from guiguts.checkers import CheckerDialog, CheckerEntry
 from guiguts.data import dictionaries
 from guiguts.maintext import maintext
 from guiguts.misc_tools import tool_save
+from guiguts.utilities import cmd_ctrl_string
 from guiguts.preferences import (
     PersistentString,
     PrefKey,
@@ -64,7 +65,9 @@ class JeebiesChecker:
 
         # Create the checker dialog to show results
         checker_dialog = CheckerDialog.show_dialog(
-            "Jeebies Results", rerun_command=jeebies_check
+            "Jeebies Results",
+            rerun_command=jeebies_check,
+            process_command=self.process_jeebies,
         )
         frame = ttk.Frame(checker_dialog.header_frame)
         frame.grid(column=0, row=1, columnspan=2, sticky="NSEW")
@@ -102,7 +105,8 @@ class JeebiesChecker:
                 [
                     "Left click: Select & find he/be error",
                     "Right click: Remove he/be error from list",
-                    "Shift Right click: Remove all matching he/be errors",
+                    f"With {cmd_ctrl_string()} key: Also toggle queried he/be",
+                    "Shift Right click: Also remove all matching he/be errors",
                 ]
             ),
             use_pointer_pos=True,
@@ -666,6 +670,17 @@ class JeebiesChecker:
                         self.dictionary[key] = int(value)
         except FileNotFoundError as exc:
             raise DictionaryNotFoundError(file) from exc
+
+    def process_jeebies(self, checker_entry: CheckerEntry) -> None:
+        """Process the Jeebies query."""
+        if checker_entry.text_range is None:
+            return
+        start_mark = CheckerDialog.mark_from_rowcol(checker_entry.text_range.start)
+        end_mark = CheckerDialog.mark_from_rowcol(checker_entry.text_range.end)
+        match_text = maintext().get(start_mark, end_mark)
+        # Toggle match text
+        replacement_text = "he" if match_text == "be" else "be"
+        maintext().replace(start_mark, end_mark, replacement_text)
 
 
 def jeebies_check() -> None:


windymilla (approved):
Looks good to me. Thanks for including the bug fix to `checkers.py` too - it would have turned up to bite us at a later date if you hadn't caught it.

srjfoo:
I need some clarification on the meaning of the tooltips. Here's a screenshot of what the jeebies one looks like for me (note to @G4OEU -- the tooltip for me shows `Command` where you wouild see `Control`):

image


I'd like to see one of two things; since `ctrl`/`cmd` applies only to the left click (as documented in the [gg2 manual notes](https://www.pgdp.net/wiki/PPTools/Guiguts2_Manual_Notes#Jeebies), change the text on that tooltip to be explicit, or move it to just under the left-click (and possibly indent it so that it's obvious that it applies only to that option -- though that would imply that shift-right-click should be indented, too).

\<tb>

Digression: I took the liberty of reformatting the tool options in the notes. The first three have `<code>` tags wrapped around the shortcuts, the last one, to show another option, has `<kbd>` which shows in red for the DP wiki, as we've added CSS to do that. I'm happy to go with either one, but wanted y'all to see what they both looked like ( I like the `<kbd>` better for this purpose, I think, but am okay with either).

srjfoo (approved):
Works as documented in the gg2 manual notes; I'd still like to see some adjustment of the tooltips, but am fine with that being done in a separate PR.

G4OEU:
> Works as documented in the gg2 manual notes; I'd still like to see some adjustment of the tooltips, but am fine with that being done in a separate PR.

Thank you @srjfoo for the improvements to the tooltips part of the Jeebies manual text draft. I've reordered things as you suggest so that the left-click pair are together, ditto for the right-click pair and I've reverted to highlighting the last tooltip as `<code>` for consistency. I hope these changes address the key adjustments you wanted to see. 

Re. the new `<kbd>` markup, I would prefer to use that only where we want to bring to the attention of readers, warnings or other special situations in the manual. I guess though, how it is best deployed will ultimately be a decision for the manual author to make.

windymilla:
> since `ctrl`/`cmd` applies only to the left click

I'm not sure that's true - it's certainly not intended to be true. You can use Cmd-right-click to fix the error and remove it from the list in one click. The tooltip tries to give that info in a brief form (remembering that the tool tip is mostly a reminder, or an alert to the user that there is a feature they could look up in the manual - the tooltip is not a full training manual in itself, and we don't want it to be intrusive). That's why the tooltip says "With Cmd/Ctrl..." - it means that with the addition of Cmd/Ctrl to the above click, you also get the he/be error fixed. GG1 behaves like this (I believe).

I'd be interested in both your views on whether the above paragraph accurately describes the behavior, and is that intended behavior desirable.

### Summary (should probably go in the manual in an adapted form):
Left=view
Right=remove
Cmd/Ctrl=Fix (in addition to left/right action)
Shift=All - do it (whatever "it" is) to all same/similar, not just the clicked one.

G4OEU:
My apologies to @srjfoo. I misunderstood her comments above and the screenshot she provided. I have only updated the draft manual text, not the text produced for the 'tooltip' in Jeebies. That is now out of line with the tootip text produced by Jeebies. I'll sort all this out when a resolution of this thread is reached.

windymilla:
I think the tooltip text does describe what happens - Sharon is going to have another look. If we decide to update any of the tooltips for the various checker tools, we can do that on a fresh branch & PR.

srjfoo:
> I think the tooltip text does describe what happens - Sharon is going to have another look. If we decide to update any of the tooltips for the various checker tools, we can do that on a fresh branch & PR.

Yes, `cmd`/`ctrl`+`right-click` does toggle he/be and remove from list. There **_is_** one small problem, though: It doesn't retain initial caps, but converts all to lower case. 😁 

> I'll sort all this out when a resolution of this thread is reached.

@G4OEU -- I've taken the liberty of re-rearranging the notes description. I took out the NB about Mac shortcuts. We should certainly put an explanation at the beginning of the manual, but what form that will take depends on whether we can stick to `Cmd` on macOS being used everywhere that `Ctrl` is on Windows. There's already one Mac command that doesn't have an equivalent on Windows.

G4OEU:
The , `cmd/ctrl`+`right-click` action that @srjfoo described above needs to be added to the Jeebies tooltips display code. I think the entries there should also be reordered. The Python code that currently says
```
                    "Left click: Select & find he/be error",
                    "Right click: Remove he/be error from list",
                    f"With {cmd_ctrl_string()} key: Also toggle queried he/be",
                    "Shift Right click: Also remove all matching he/be errors",
```
would then look like
```
                    "Left click: Select & find he/be error",
                    f"With {cmd_ctrl_string()} key: Also toggle queried he/be",
                    "Right click: Remove he/be error from list",
                    f"With {cmd_ctrl_string()} key: Also toggle queried he/be before removing from list",
                    "Shift Right click: Also remove all matching he/be errors",
```
I think that reordering is within the spirit of the earlier post by @windymilla

> .... You can use Cmd-right-click to fix the error and remove it from the list in one click. The tooltip tries to give that info in a brief form (remembering that the tool tip is mostly a reminder, or an alert to the user that there is a feature they could look up in the manual - the tooltip is not a full training manual in itself, and we don't want it to be intrusive). That's why the tooltip says "With Cmd/Ctrl..." - it means that with the addition of Cmd/Ctrl to the above click, you also get the he/be error fixed. GG1 behaves like this (I believe).
> 
> I'd be interested in both your views on whether the above paragraph accurately describes the behavior, and is that intended behavior desirable.
> Summary (should probably go in the manual in an adapted form):
> 
> Left=view
> Right=remove
> Cmd/Ctrl=Fix (in addition to left/right action)
> Shift=All - do it (whatever "it" is) to all same/similar, not just the clicked one.

That preceding 4-line summary is new to me and is key to understanding Cmd/Ctrl+clicking behaviours. I got caught out when I experimented with Ctrl+right-click which is not currently in the Jeebies tool tip dialog. If I had understood the behaviours better I would have known what would happen.

windymilla:
> The , `cmd/ctrl`+`right-click` action that @srjfoo described above needs to be added to the Jeebies tooltips display code.

I think what is there does convey the same, but without the almost duplicated line, and would be sufficient as a reminder for users (though no substitute for reading the manual of course. 

The current version says "Left does this, Right does that, With Ctrl also fixes hebe", the implication being that the Ctrl additional action applies to Left & Right. This is in contrast to Shift, which explicitly says it applies to right-clicking.

However, if prevailing opinion is that it's not clear enough, I've no real objection to expanding the tooltip, though it will then be 5 lines long!

srjfoo:
> > The , `cmd/ctrl`+`right-click` action that @srjfoo described above needs to be added to the Jeebies tooltips display code.
> 
> I think what is there does convey the same, but without the almost duplicated line, and would be sufficient as a reminder for users (though no substitute for reading the manual of course.
> 
> The current version says "Left does this, Right does that, With Ctrl also fixes hebe", the implication being that the Ctrl additional action applies to Left & Right. This is in contrast to Shift, which explicitly says it applies to right-clicking.
> 
> However, if prevailing opinion is that it's not clear enough, I've no real objection to expanding the tooltip, though it will then be 5 lines long!

I don't see anything wrong with the manual and the tooltip not saying exactly the same thing. It's often good to explain things different ways in different places. I vote for deferring the decision, but like @windymilla, am okay with expanded tooltips if necessary.

@windymilla -- a tooltip section of the manual notes might be a good idea?

windymilla:
> @windymilla -- a tooltip section of the manual notes might be a good idea?

Would that be to explain how tooltips work, or is it to note that we might need to add/adjust tooltips for specific dialogs? If the latter, I think that's better done through a Github issue, since it's code that needs fixing, not the manual. If the former, then quite happy for one to be created.

windymilla:
I just thought of a further "complication" - you can use Ctrl+Shift+left-click or Ctrl+Shift+right-click. They either select the error (left) and fix (Ctrl) all matching errors (Shift), or fix (Ctrl) and remove (Right) all matching errors (Shift).

Of course for jeebies, it will be rare for there to be any matching errors, but it could easily happen for other checks. You can artificially make an example of a matching Jeebies error, by copying a line with a hebe query to another location in the file. In this case "matching" error means the whole line is identical.

G4OEU:
Yep, I did that to test that Ctrl+right-click worked as expected and it did. And, yes, I discovered that the whole (duplicate) line had to be identical.

PR #231, opened 2024-05-07
Add rewrap functionailty

### LINKED ISSUE #54: Rewrap features
Algorithm should match GG1 pretty closely as it's been tailored for our needs.

windymilla:
Same behavior as Guiguts 1, as documented in the manual: https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_Manual/Tools_Menu#Rewrapping

More tolerant regarding blank lines than GG1, e.g. if no blank line before/after markup or between nested markups, it should still work correctly.

Only blockquote markup can have other markup nested inside.

Wrapping margins can be set in the Prefs dialog, or as described in the manual, can be overridden using markup for some types, e.g. `/#[8.4,68]`

At present, rewrapping uses the "greedy" algorithm (via the standard python `textwrap` module), that used to be in GG 1.0.25. The other (Knuth-Plass) algorithm that was in GG 1.0.25 may be added to GG2 at a later date if desired.

FIxes #54

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -285,6 +285,15 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(
             PrefKey.JEEBIES_PARANOIA_LEVEL, JeebiesParanoiaLevel.NORMAL
         )
+        preferences.set_default(PrefKey.WRAP_LEFT_MARGIN, 0)
+        preferences.set_default(PrefKey.WRAP_RIGHT_MARGIN, 72)
+        preferences.set_default(PrefKey.WRAP_BLOCKQUOTE_INDENT, 2)
+        preferences.set_default(PrefKey.WRAP_BLOCKQUOTE_RIGHT_MARGIN, 72)
+        preferences.set_default(PrefKey.WRAP_BLOCK_INDENT, 2)
+        preferences.set_default(PrefKey.WRAP_POETRY_INDENT, 4)
+        preferences.set_default(PrefKey.WRAP_INDEX_MAIN_MARGIN, 2)
+        preferences.set_default(PrefKey.WRAP_INDEX_WRAP_MARGIN, 8)
+        preferences.set_default(PrefKey.WRAP_INDEX_RIGHT_MARGIN, 72)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -418,17 +427,20 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
 
     def init_tools_menu(self) -> None:
         """Create the Tools menu."""
-        menu_edit = Menu(menubar(), "~Tools")
-        menu_edit.add_button("Basic Fi~xup", basic_fixup_check)
-        menu_edit.add_button("~Word Frequency", word_frequency)
-        menu_edit.add_button(
+        menu_tools = Menu(menubar(), "~Tools")
+        menu_tools.add_button("Basic Fi~xup", basic_fixup_check)
+        menu_tools.add_button("~Word Frequency", word_frequency)
+        menu_tools.add_button(
             "~Spelling Check",
             lambda: spell_check(
                 self.file.project_dict, self.file.add_good_word_to_project_dictionary
             ),
         )
-        menu_edit.add_button("PP~txt", lambda: pptxt(self.file.project_dict))
-        menu_edit.add_button("~Jeebies", jeebies_check)
+        menu_tools.add_button("PP~txt", lambda: pptxt(self.file.project_dict))
+        menu_tools.add_button("~Jeebies", jeebies_check)
+        menu_tools.add_separator()
+        menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
+        menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
 
     def init_view_menu(self) -> None:
         """Create the View menu."""
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -10,7 +10,7 @@ from typing import Any, Callable, Final, TypedDict, Literal, Optional
 
 import regex as re
 
-from guiguts.maintext import maintext, PAGE_FLAG_TAG
+from guiguts.maintext import maintext, PAGE_FLAG_TAG, PAGEMARK_PIN
 from guiguts.page_details import (
     PageDetail,
     PageDetails,
@@ -695,6 +695,71 @@ class File:
         """
         if self.project_dict.add_good_word(word):
             self.project_dict.save(self.filename)
+
+    def rewrap_selection(self) -> None:
+        """Wrap selected text."""
+        ranges = maintext().selected_ranges()
+        if not ranges:
+            sound_bell()
+            return
+        # Adjust selection to begin & end at start of lines
+        ranges[0].start.col = min(ranges[0].start.col, 0)
+        if ranges[0].end.col > 0:
+            ranges[0].end.col = 0
+            ranges[0].end.row += 1
+        self.rewrap_section(ranges[0])
+
+    def rewrap_all(self) -> None:
+        """Wrap whole text."""
+        self.rewrap_section(IndexRange("1.0", maintext().index(tk.END)))
+
+    def rewrap_section(self, section_range: IndexRange) -> None:
+        """Wrap a section of the text, preserving page mark locations.
+
+        Args:
+            section_range: Range of text to be wrapped.
+        """
+        maintext().undo_block_begin()
+        maintext().strip_end_of_line_spaces()
+        mark_list = self.pin_page_marks()
+        maintext().rewrap_section(section_range)
+        self.unpin_page_marks(mark_list)
+
+    def pin_page_marks(self) -> list[str]:
+        """Pin the page marks to locations in the text, by inserting a special
+        character at each page mark.
+
+        This allows operations such as rewrapping to happen without losing
+        page mark positions. Marks are unpinned by `unpin_page_marks`.
+
+        Returns:
+            List of page mark names in reverse order to pass to `unpin_page_marks`.
+        """
+        mark_list = []
+        mark: str = tk.END
+        while mark := page_mark_previous(mark):
+            mark_list.append(mark)
+            maintext().insert(mark, PAGEMARK_PIN)
+        return mark_list
+
+    def unpin_page_marks(self, mark_list: list[str]) -> None:
+        """Unpin the page marks by setting them to the locations of the special
+        characters inserted by `pin_page_marks`.
+
+        Args:
+            mark_list: List of page mark names in reverse order from `pin_page_marks`.
+        """
+        # First set each page mark to the pinned location - this avoids setting two
+        # page marks adjacent to one another (due to the pin character) because their
+        # order isn't guaranteed if you do that, so they can end up swapped.
+        found = "1.0"
+        while found := maintext().search(PAGEMARK_PIN, found, "end"):
+            maintext().mark_set(mark_list.pop(), found)
+            found = f"{found} +1c"
+        # Now it's safe to remove the pins, in reverse order to simplify loop
+        found = tk.END
+        while found := maintext().search(PAGEMARK_PIN, found, backwards=True):
+            maintext().delete(found, f"{found} +1c")
 
 
 def page_mark_previous(mark: str) -> str:
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -720,10 +720,18 @@ class File:
             section_range: Range of text to be wrapped.
         """
         maintext().undo_block_begin()
+
+        # Dummy insert & delete, so that if user undoes the wrap, the insert
+        # cursor returns to its previous point, not the first pin page mark position.
+        insert_index = maintext().get_insert_index()
+        maintext().insert(insert_index.index(), " ")
+        maintext().delete(insert_index.index())
+
         maintext().strip_end_of_line_spaces()
         mark_list = self.pin_page_marks()
-        maintext().rewrap_section(section_range)
-        self.unpin_page_marks(mark_list)
+        maintext().rewrap_section(
+            section_range, lambda: self.unpin_page_marks(mark_list)
+        )
 
     def pin_page_marks(self) -> list[str]:
         """Pin the page marks to locations in the text, by inserting a special
@@ -735,6 +743,12 @@ class File:
         Returns:
             List of page mark names in reverse order to pass to `unpin_page_marks`.
         """
+        # Ensure no pins left around from previous wraps - should never happen,
+        # but if it has, we want to clear them.
+        found: str = tk.END
+        while found := maintext().search(PAGEMARK_PIN, found, backwards=True):
+            maintext().delete(found, f"{found} +1c")
+
         mark_list = []
         mark: str = tk.END
         while mark := page_mark_previous(mark):
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1277,6 +1277,12 @@ class MainText(tk.Text):
 
         # Output any last paragraph
         if paragraph:
+            # If paragraph runs right to end of file, ensure it has a terminating newline
+            if (
+                IndexRowCol(maintext().index(f"{line_start} +1l")).row
+                == maintext().end().row
+            ):
+                maintext().insert(tk.END, "\n")
             self.wrap_paragraph(
                 paragraph_start,
                 f"{line_start} +1l",
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2,6 +2,7 @@
 
 
 import logging
+from textwrap import TextWrapper
 import tkinter as tk
 from tkinter import ttk
 from tkinter import font as tk_font
@@ -21,7 +22,11 @@ from guiguts.utilities import (
 logger = logging.getLogger(__package__)
 
 TK_ANCHOR_MARK = "tk::anchor1"
+WRAP_NEXT_LINE_MARK = "WrapParagraphStart"
+INDEX_NEXT_LINE_MARK = "IndexLineStart"
+WRAP_END_MARK = "WrapSectionEnd"
 PAGE_FLAG_TAG = "PageFlag"
+PAGEMARK_PIN = "\x7f"  # Temp char to pin page mark locations
 
 
 class FindMatch:
@@ -35,6 +40,22 @@ class FindMatch:
     def __init__(self, index: IndexRowCol, count: int):
         self.rowcol = index
         self.count = count
+
+
+class WrapParams:
+    """Stores parameters used when wrapping."""
+
+    def __init__(self, left: int, first: int, right: int):
+        """Initialize WrapParams object.
+
+        Args:
+            left: Left margin position.
+            first: Left margin for first line.
+            right: Right margin position.
+        """
+        self.left = left
+        self.first = first
+        self.right = right
 
 
 class TextLineNumbers(tk.Canvas):
@@ -349,6 +370,29 @@ class MainText(tk.Text):
         self.delete("1.0", tk.END)
         self.set_modified(False)
         self.edit_reset()
+
+    def undo_block_begin(self) -> None:
+        """Begin a block of changes that will be undone with one undo operation.
+
+        Block is automatically closed when system becomes idle.
+
+        Note: this version does not support nesting of blocks.
+        """
+        self.config(autoseparators=False)
+        self.edit_separator()
+        self.after_idle(self.undo_block_end)
+
+    def undo_block_end(self) -> None:
+        """End a block of changes that will be undone with one undo operation.
+
+        Normally called automatically when system becomes idle, but can safely be
+        called manually if required, e.g. to start & end two blocks within one
+        user operation.
+
+        Note: this version does not support nesting of blocks.
+        """
+        self.edit_separator()
+        self.config(autoseparators=True)
 
     def get_insert_index(self) -> IndexRowCol:
         """Return index of the insert cursor as IndexRowCol object.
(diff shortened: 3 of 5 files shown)


G4OEU:
On a 'Rewrap All' and comparing (using Notepad++) the same file after doing this with GG1 and GG2, there is not much in it. Sometimes the Knuth algorithm (GG1) has marginally tidier rewrapping than the 'greedy' algorithm of GG2 but sometimes the opposite is the case.

On a 'normal' book file there doesn't seem to be much in it either way and the GG2 output overall is quite acceptable. On the first book I tested with I suppose it's no surprise that the Knuth algorithm wraps so the resulting file is about 72 lines longer than the 'greedy' algorithm result from GG2.

 I'll try to find some pathological text examples if I can, perhaps by using a narrower page width?

I've yet to test rewrap on a selection (but should be similar result to the above) or when rewrap markers are included. The latter will be more interesting!

windymilla:
Thanks for the testing @G4OEU - one way to create a pathological case for Mr Greedy versus Mr Knuth involves having some short words at the end of one line, then a long word at the end of the next line, e.g. 
```
This has short words at the end of the line as you can see. But when you
have the last word too long to fit, for example this one: please-do-not-split-me,
line 2 wraps a lot shorter than lines 1 and 3.
```
Mr Greedy wraps this as (right margin 72):
```
This has short words at the end of the line as you can see. But when you
have the last word too long to fit, for example this one:
please-do-not-split-me, line 2 wraps a lot shorter than lines 1 and 3.
```
Mr Knuth wraps it as:
```
This has short words at the end of the line as you can see. But
when you have the last word too long to fit, for example this one:
please-do-not-split-me, line 2 wraps a lot shorter than lines 1 and 3.
```

As you say, there are also cases where you can get better results with greedy than Knuth, and to some extent, it is subjective. I'm also not sure how much/often it would detract from the reader's enjoyment with real books.

srjfoo:
> As you say, there are also cases where you can get better results with greedy than Knuth, and to some extent, it is subjective. I'm also not sure how much/often it would detract from the reader's enjoyment with real books.

It would be interesting to know how many people actually read the text version. I'm right that this really affects only those reading the text version?

windymilla:
Yes, only the text version is affected

srjfoo:
Since I'm not a practiced PPer, I tend to poke at a lot of stuff in no especial order (I really should be more orderly about it!) I found a couple of things. In this screenshot, notice that the selection "accidentally" includes the first line of the next no-wrap block (and includes a page marker between blocks). Both blocks were indented, no error was generated:
image

Another thing that can be pretty annoying -- if I don't like what the rewrap selection does, and "Undo" it, the cursor is moved to the beginning of the line on which the last page separator or page marker is.

---

I had a block-quote spanning two pages with a signature line at the end. I took out the bq markers spanning the page marker (which perhaps I shouldn't have done?), gave the bq a special indent, and left-aligned the signature. It all worked nicely, but I got that error. I'm also getting an error on just rewrapping a selection, so I probably need to start over with a clean file and a new run of gg2.
```
18:22:16: ERROR - root.py:45 - Tkinter Exception
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/tkinter/__init__.py", line 1967, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/Users/sharonjoiner/dpdev/guiguts-py-3.11/src/guiguts/file.py", line 710, in rewrap_selection
    self.rewrap_section(ranges[0])
  File "/Users/sharonjoiner/dpdev/guiguts-py-3.11/src/guiguts/file.py", line 726, in rewrap_section
    self.unpin_page_marks(mark_list)
  File "/Users/sharonjoiner/dpdev/guiguts-py-3.11/src/guiguts/file.py", line 757, in unpin_page_marks
    maintext().mark_set(mark_list.pop(), found)
                        ^^^^^^^^^^^^^^^
IndexError: pop from empty list
```

Based on `File "/Users/sharonjoiner/dpdev/guiguts-py-3.11/src/guiguts/file.py", line 757, in unpin_page_marks
    maintext().mark_set(mark_list.pop(), found)`, it has something to do with the page markers, so it's undoubtedly something I'm not understanding about what I'm supposed to be doing.

G4OEU:
Rewrap selection involving /CC/ markers is not working correctly. 

Select and repeatedly 'Rewrap Selection' the following 4 lines to see what I mean:

```
/C
————
XX
C/
```

GG1.6.1 correctly centers the block the first time and dosen't change that on subsequent repeats of the command.

GG2 correctly centers the block the first time but then keeps shifting it right, with other changes to relative positions, on subsequent repeats of the command.

G4OEU:
Poetry markup with 'Rewrap Selection' seems to be working correctly with an assumed default poetry indentation of 4 spaces.

It moves the left-most line of the block left or right as appropriate so that it is indented 4 spaces. The rest of the lines in the block are shifted appropriately so as to maintain their same relative positions to that line.

Subsequent repeats of the 'Rewrap Selection' command leave that initial rewrapping unchanged.

/PP/, /Pp/ and /pp/ are recognised.

windymilla:
> 'Rewrap Selection' involving /CC/ markers is not working correctly.
> 
> Select and repeatedly 'Rewrap Selection' the following 4 lines to see what I mean:
> 
> ```
> /C
> ————
> XX
> C/
> ```

Thanks for catching that - caused by a last-minute change to the method of centering that I didn't test fully. I'll push a commit to fix it later.

cpeel (approved):
Code structure LGTM and nothing jumped out at me as problematic. Nice job on handling some really complex edgecases.

windymilla:
> Since I'm not a practiced PPer, I tend to poke at a lot of stuff in no especial order (I really should be more orderly about it!) I found a couple of things. In this screenshot, notice that the selection "accidentally" includes the first line of the next no-wrap block (and includes a page marker between blocks). Both blocks were indented, no error was generated:

Thanks - it was just checking there was closing markup later in the file, instead of restricting that check to the selected region. Will fix.

windymilla:
> Another thing that can be pretty annoying -- if I don't like what the rewrap selection does, and "Undo" it, the cursor is moved to the beginning of the line on which the last page separator or page marker is.

Interesting - I'm sure similar effects must happen in GG1, where Undo leaves your cursor where you didn't expect it - it goes to the first insert/delete done (which is the last when you are undoing of course). I've added a dummy insert & delete of a space at the current cursor position, so an undo should put the cursor back at that spot. Will commit later.

windymilla:
> I had a block-quote spanning two pages with a signature line at the end. I took out the bq markers spanning the page marker (which perhaps I shouldn't have done?), gave the bq a special indent, and left-aligned the signature. It all worked nicely, but I got that error. I'm also getting an error on just rewrapping a selection, so I probably need to start over with a clean file and a new run of gg2.

@srjfoo - please can you give me a test file and instructions, assuming this is repeatable - thanks. No need for it to be a minimal file - anything that demonstrates it will be good. Thanks

G4OEU:
Have not been able to break /#...#/ rewrap markup when rewrapping plain text. It behaves as described in the GG manual.

However there is one noticeable difference between GG2 and GG1.6.1 block rewrap in operation - with GG2, if you dont include the closing #/ line in the selection, it still works quite happily BUT no longer inserts an extra blank line after the rewrapped text as GG1.6.1 does. I think this is a 'good thing'.

Also tried block rewrap with an included /R..R/ block. As per the manual, the right margin of the containing block is used unless
/R[n] overrides it. 

Another difference between GG2 and GG1.6.1 can be seen if a 'Rewrap All' is dones on a file containing the following text and markup. Notice that this example breaks the rules set out in the GG manual that requires that the "opening and closing markers (/*, etc.) should stand alone on a line, but an opening markup must be preceded by a blank line, another opening rewrap marker, or a page separator (if you haven't already removed them), and a closing markup must be followed by a blank line, another closing rewrap marker, or a page separator (etc.). If the blank lines, permitted rewrap markers or page separators are missing, Guiguts will fail to recognize either the beginning or the end of the markup and mis-wrapping will result."

GG2 correctly rewraps this example, including the embedded /R..R/ markup, but adds a blank line after 'Line 3'.

GG1.61. does not recognise the /R..R/ markup and treats it as text that is rewrapped by the enclosing /#..#/ markup. It also adds a blank line after 'Line 1', 'Line 2' and 'Line 3'.

```
Line 1
/#[4.2,60]
  The shorter service was a comfortable change for those
whose uneasy task it was to keep their children
    “behaving themselves” in the church. James Spence, whose
    regular remark to his neighbours on
/R[70]
                          RIGHT TEXT
R/
their road home had been, “He said a great deal, really he
    said a very great deal,” gave as his opinion that “he
   really thocht that there was less skailt noo that the
    kirk skailed sooner,[9] for even yet there was a great
 deal said that we couldna carry home, and mair that we
    didna carry oot; and wi’ the bairns no’ bein’ sae
    fashious through no’ bein’ sae lang in, we could attend
   better.”
#/
Line 2
/#[8.2,70]
  The shorter service was a comfortable change for those whose uneasy
        task it was to keep their children “behaving themselves” in
        the church. James Spence, whose regular remark to his
        neighbours on their road home had been, “He said a great deal,
     really he said a very great deal,” gave as his opinion that
        “he really thocht that there was less skailt noo that the kirk
        skailed sooner,[9] for even yet there was a great deal said
        that we couldna carry home, and mair that we didna carry oot;
        and wi’ the bairns no’ bein’ sae fashious through no’ bein’
        sae lang in, we could attend better.”
#/
Line 3
```

G4OEU:
Additional comments re. ```/#..#/``` markup with embedded markup of other types.

If, in the comment above, you replace the
```
/R[70]
                          RIGHT TEXT
R/
```
block with, say,
```
/$
RIGHT TEXT
MORE TEXT
$/
```
and then 'Rewrap All', the text remains unchanged and against the left margin as expected.

Changing the ```/$..$/``` markup to ```/*..*/``` markup then doing a 'Rewrap All' indents the text six characters. I'm not sure whether that is correct. The GG manual says ```/*..*/``` markup "defaults to no-wrap indentation specified on the Preferences Menu".

windymilla:
> Changing the `/$..$/` markup to `/*..*/` markup then doing a 'Rewrap All' indents the text six characters. I'm not sure whether that is correct. The GG manual says `/*..*/` markup "defaults to no-wrap indentation specified on the Preferences Menu".

I'm also not absolutely sure, but I made a decision to code it to be different to GG1. It makes much more sense logically, if you have some `/*` markup nested inside a blockquote for it to indent from the blockquote margin (unless you specify a left margin, in which case it respects that). Similarly centering and right aligning take their margins from the "current" margins, i.e. the ones in the block you are currently in, which might be a (potentially nested) blockquote. This should be legal:
```
outside
/#[5,50]
first level with specific margin
/#
second level
/*
indented block inside double nested blockquote
*/
/*[25]
specifically indented block
*/
also second
#/
also first
#/
also outside
```

srjfoo:
> > I had a block-quote spanning two pages with a signature line at the end. I took out the bq markers spanning the page marker (which perhaps I shouldn't have done?), gave the bq a special indent, and left-aligned the signature. It all worked nicely, but I got that error. I'm also getting an error on just rewrapping a selection, so I probably need to start over with a clean file and a new run of gg2.
> 
> @srjfoo - please can you give me a test file and instructions, assuming this is repeatable - thanks. No need for it to be a minimal file - anything that demonstrates it will be good. Thanks

In the clearer light of day, I went through the process again, writing down the steps as I did them. It's entirely possible that I was messing something up last night in my imperfect understanding of how to mark up special indents, because it worked fine this time. I'll keep an eye out for a repetition and let you know.

Footnote (I double-checked, and I confirm the behavior on Mac that Quentin saw for repeated rewrap selection on the centering markup.

windymilla:
> In the clearer light of day, I went through the process again, writing down the steps as I did them. It's entirely possible that I was messing something up last night in my imperfect understanding of how to mark up special indents, because it worked fine this time. I'll keep an eye out for a repetition and let you know.

Looking at the place that error would appear, I do see a reason why it could happen (though I think have fixed it while fixing something else). To make it happen you would have to have done something like:
1. rewrap select region with badly nested markup so you get the error.
2. undo (once I think) and you end up with `\x7f` characters in the file (a second undo would remove them, but without it they would be left behind
3. do another rewrap (which adds yet more `\x7f` characters to the ones left over from 2.
4. Once it finishes, it tries to set a page marker for each of the `\x7f` characters, but of course there are too many of them, so it tries to pop more data off the stack than is stored there.

windymilla:
@srjfoo & @G4OEU - I have pushed another commit (you will probably need to do "Fetch Origin" if using Github Desktop to get the update). Hopefully that fixes the bugs, and other points have been addressed above.

srjfoo:
> @srjfoo & @G4OEU - I have pushed another commit (you will probably need to do "Fetch Origin" if using Github Desktop to get the update). Hopefully that fixes the bugs, and other points have been addressed above.

I don't know how Github Desktop works, but from the command line, I have your fork added as a remote. So, when I need to get an updated version, I `git fetch windymilla` and `git rebase <branchname>` to apply the updates. Or simply delete my existing local branch and recreate it from the newly updated one.

G4OEU:
There is a small problem with the @windymilla example below and the very latest GG2.
```
outside
/#[5,50]
first level with specific margin
/#
second level
/*
indented block inside double nested blockquote
*/
/*[25]
specifically indented block
*/
also second
#/
also first
#/
also outside
```
A 'Rewrap All' does most of what you expect except for the very last line, 'also outside',  which gets appended to the last closing ```#/``` marker.

windymilla:
> A 'Rewrap All' does most of what you expect except for the very last line, 'also outside', which gets appended to the last closing `#/` marker.

Thanks - it just appears to go wrong if there's no newline on the end of the last line "also outside". I'll look at it.

windymilla:
@G4OEU - hopefully my latest commit will have fixed that. It's actually good for each line to have a newline character, so if it detects the last line doesn't have a newline on the end, it adds one.

srjfoo:
> @G4OEU - hopefully my latest commit will have fixed that. It's actually good for each line to have a newline character, so if it detects the last line doesn't have a newline on the end, it adds one.

It fixes it on macOS, and the added newline is included/visible in the selection range.

srjfoo:
> > Another thing that can be pretty annoying -- if I don't like what the rewrap selection does, and "Undo" it, the cursor is moved to the beginning of the line on which the last page separator or page marker is.
> 
> Interesting - I'm sure similar effects must happen in GG1, where Undo leaves your cursor where you didn't expect it - it goes to the first insert/delete done (which is the last when you are undoing of course). I've added a dummy insert & delete of a space at the current cursor position, so an undo should put the cursor back at that spot. Will commit later.

I don't ever remember seeing it in GG1, but then I didn't really trust `undo` in GG1 much, anyway because of some of the funky ways it works on a Mac.

The commit you added seems to fix it, though, thanks!

G4OEU:
Trying to understand the behaviour of the second ```/c..c/``` block (YY) in the following:
```
outside
/#[5,50]
     first level with specific margin
/c
                    ----
                     XX
c/
/#
       second level
/*
         indented block inside double nested blockquote
*/
/*[25]
                         specifically indented block
*/
/c
                   ----
                    YY
c/
       also second
#/
     also first
#/
also outside
```

G4OEU:
Looking again at the rewrap markers example in my previous post, I think the centered XX block should be 5 more places to the right. It is correctly 'centered' in a 45-character wide text block but that text block is indented 5 characters from the left margin so the position of the second X should be at column 27, not 22?

windymilla:
> Looking again at the rewrap markers example in my previous post, I think the centered XX block should be 5 more places to the right. It is correctly 'centered' in a 45-character wide text block but that text block is indented 5 characters from the left margin so the position of the second X should be at column 27, not 22?

Quite right, thanks - I've pushed another commit to fix that. It's actually simplified the indent calculation, so that's a good thing too.

G4OEU:
Your latest push has fixed the first, XX, centering issue. Thanks.

However I still don't understand why the second pair of centering markers, around the YY block, are acting as they do. What are the implied margins within which they are centering?

windymilla:
> However I still don't understand why the second pair of centering markers, around the YY block, are acting as they do. What are the implied margins within which they are centering?

The outer blockquote is 5 to 50.
The inner blockquote is therefore 7 to 50, because each blockquote depth adds the blockquote indent (2) to the left margin, and leaves the right margin unchanged.
The YY block is centered within those margins from 7 to 50, and is centered on column 28. 

It's clearer with more text than in my example:
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed varius
vehicula sodales. Etiam leo lorem, scelerisque ut porta et, dictum vel
ipsum.
/#[5,50]
     The shorter service was a comfortable change
     for those whose uneasy task it was to keep
     their children “behaving themselves” in the
     church.
/c
                         ----
                          XX
c/
/#
       James Spence, whose regular remark to his
       neighbours on their road home had been, “He
       said a great deal, really he said a very
       great deal,” gave as his opinion that
/c
                          ----
                           YY
c/
       “he really thocht that there was less
       skailt noo that the kirk skailed sooner,[9]
       for even yet there was a great
#/
     The shorter service was a comfortable change
     for those whose uneasy task it was to keep
     their children “behaving themselves” in the
     church.
#/
Praesent tristique velit a nisl tincidunt convallis. Sed id sem felis.
Donec non ultricies metus. Donec turpis diam, maximus in risus ut,
aliquet maximus purus.

```

G4OEU:
OK, I get that. Thanks.

windymilla:
> OK, I get that. Thanks.

Thanks @G4OEU - I realized while checking the answer to your question that everything was there (but just ignored) for allowing the user to customize the centering. So the latest commit allows
`/C[35]`
to center on column 35, for example.
Of course `/C` with no customization behaves as before, centering in the "current" margins.

G4OEU:
Neat. I assume that it will be easy to generate the required HTML for that customisation?

The GG manual will need updating for this change to the ``/C..C/`` rewrap markers. Is there a way to flag comments in a conversation like this that the GG manual author/updater can readily locate? Your explanation above re. the way indenting operates in nested ``/#..#/`` blocks would also be something they may want to use in the manual.

windymilla:
> Neat. I assume that it will be easy to generate the required HTML for that customisation?

That's a very good question. It should be possible, since centering on 25, is identical to centering in margins 0 to 50. Hopefully it will be OK. If not, we'll just have to let it do default centering in HTML and if the PPer is concerned, they will have to adjust it manually.

> The GG manual will need updating for this change to the `/C..C/` rewrap markers. Is there a way to flag comments in a conversation like this that the GG manual author/updater can readily locate? Your explanation above re. the way indenting operates in nested `/#..#/` blocks would also be something they may want to use in the manual.

Funnily enough, while you were typing that comment, I was creating this: https://www.pgdp.net/wiki/PPTools/Guiguts2_Manual_Notes

Sharon & I had discussed the need for it previously, for the same reasons as you mention. It is possible to add a link to this PR (or any other) if needed, but in this case, I thought a comment about the changes would suffice.

You and @srjfoo should definitely feel free to edit that wiki page with anything that seems useful - thanks

windymilla:
Further note about documentation after suggestion by @G4OEU (thanks):
I have added a new label "Manual" which can be applied to a PR (or an issue, I guess).
I have added it to half a dozen PRs that have info that could be relevant to the GG2 manual writer.
To see them, you can either click "Labels" at the top of the [PRs page](https://github.com/DistributedProofreaders/guiguts-py/pulls), which will show open PRs by default, but you can delete the `is:open` from the filter field to show all "Manual" PRs.
Alternatively, you can just type "label:manual" into the filter field.
Or, you can visit this URL: https://github.com/DistributedProofreaders/guiguts-py/issues?q=label%3Amanual

srjfoo:
> > OK, I get that. Thanks.
> 
> Thanks @G4OEU - I realized while checking the answer to your question that everything was there (but just ignored) for allowing the user to customize the centering. So the latest commit allows `/C[35]` to center on column 35, for example. Of course `/C` with no customization behaves as before, centering in the "current" margins.

I was playing around with customizing centering earlier (before you added this), but what I tried (channeling how it worked for the others) was `/c[8,50]` to force centering between columns 8 and 50. Which, of course, didn't work at that point in time. I forced it by wrapping a dummy `/#` block around it. I'm just wondering if it makes more sense to keep that notation (a range of columns) instead of centering on a specific column.

G4OEU:
It's good that we are all thinking ahead about updating the manual to reflect the changes users will see in GG2.

To that end I've updated Guiguts2_Manual_notes to include a draft text to replace the current Jeebies entry in the Tools Menu section of the GG1 manual.

I'll also draft a new text for a PPtxt entry in the Tools Menu section. There is currently a two-line stub entry for PPtxt in the Text Menu section of the GG1 manual so it needed updating in any case.

windymilla:
> I'm just wondering if it makes more sense to keep that notation (a range of columns) instead of centering on a specific column.

I did do that to start with - great minds think alike. However, there are several reasons I think the single number is better: None of the other markup has exactly 2 numbers. They either have 1 or 3 (although you can omit 1 or 2 of the 3 numbers). The extra number in the "3" case is the "first line" margin, for (hanging) indents. They only apply to things where the text is wrapped, like blockquote and index. The other cases, that have 1 number, are the block markups, like Right-align (where it's the right margin), and `/*` (where it's the left margin), which I think are more similar to the Center block case, where you don't actually wrap the lines, just shift them left or right. Another reason is that `/C[0,50]`, `/C[10,40]`, `/C[20,30]`, `/C[25,25]` would all mean exactly the same, whereas a user might easily think it was going to somehow keep the text within those margins, which the `/C` block does not do. So all of those are equivalent to `/C[25]`.

G4OEU:
I've edited the draft replacement text for the Jeebies manual entry I referred to above. One of its paragraphs says:

> Two other radio buttons on the report menu allow you to sort Jeebies report entries by "Line & Col" (the default) or by "Alpha/Type". These sorting options are standard across most Guiguts tools but in the case of Jeebies the default sorting by line number and column is the most useful.

Am I correct in saying that these sorting options will be standard across most Guiguts tools?

windymilla:
> Am I correct in saying that these sorting options will be standard across most Guiguts tools?

Yes - I've built into the code that specific tools can optionally pass into the general code two functions to sort by row/col or by alpha/type when they create the CheckerDialog. So wherever it could be helpful to sort in different ways, we should be able to make use of that. Default row/col sorting is to preserve header/footer ordering, and order the rest by row & col, within sections. Default alpha sorting is preserve header/footer ordering, and order the rest by the highlighted text in the message, or if there is no highlight it uses the whole message text.

G4OEU:
Question for @windymilla: In the Guiguts2 Manual Notes file you say at point 2 under "Rewrapping" that

> Block markup that is inside block quote markup, e.g. /* inside /# markup, will work relative to the current blockquote margins, e.g. if inside double nested block quotes (i.e. left margin = 2 * 2 = 4), the /* will indent the block a total of 6 (not 2 as in GG1)

Should that read ``2 + 2 = 4`` rather than ``2 * 2 = 4``?

windymilla:
> Should that read `2 + 2 = 4` rather than `2 * 2 = 4`?

Because it is double nested, then it gets two times the indent, so multiplication is correct. If it were triple nested, it would be 3 * 2 = 6. Or perhaps you were querying my unthinking use of "*" for multiply? For clarity for non-programmers, It would be better if I had used the multiplication sign "2 × 2 = 4".

G4OEU:
OK, thanks. I see the effect of nesting as being additive; that is each level of nesting adds another indent of 2. But I can see that if the question being asked was "What is the indent after three levels of `/#..#/` nesting" then the answer is more likely to be given as "3 x 2" rather than "2 + 2 + 2" although I would probably say the latter!

srjfoo:
> OK, thanks. I see the effect of nesting as being additive; that is each level of nesting adds another indent of 2. But I can see that if the question being asked was "What is the indent after three levels of `/#..#/` nesting" then the answer is more likely to be given as "3 x 2" rather than "2 + 2 + 2" although I would probably say the latter!

For what it's worth, I see it as being additive, also, and just hadn't gotten around to asking the same question, because if you had, say, three nested `/#` customized blocks, it _**would**_ be additive unless they were all the same. (I acknowledge that's stressing the limits of what would look acceptable, and highly unlikely to happen, but the principle is the same.)

srjfoo (approved):
@windymilla, I defer to you and @G4OEU, as y'all are the practiced PPers, and I'm sure there are some things remaining that need to be worked out, but I'm happy with it as-is unless y'all have come up with something else specific that needs to be tested for macOS.

I took a relatively short file (no, it didn't test the index rewrapping, but I'm happy with my earlier test of that fix), went through and added centering markup, some special indents, and even an unintended unmatched markup error that caught me by surprise that I know what to reuse it for. Rewrap all looked good except for the places I hadn't customized the centered places and the unmatched markup (thankfully a short section). Rewrap selection worked well when fixing the two unresolved cases.

🎉  for the unrelated things we found that got fixes done (or issues logged)!

PR #232, opened 2024-05-10
Implement Page Separator Fixup

windymilla:
1. Menu entry added to Tools menu.
2. Behavior generally similar to GG1
3. "Auto Fix" option is like "99% Auto" in GG1, but a more conservative when there are hyphens or hyphen-asterisks before/after the page separator. These need a human eye.
4. Reason for Undo/Redo buttons is that these need to normal undo/redo but then to re-view the first page separator that needs fixing, which a simple Undo/Redo can't do.
5. The "undo block" methods will be replaced by more general MainText methods at a later date. The idea is that a single "undo/redo" will undo/redo all the changes made by the last mouse click.
6. Various key bindings have been set up, to mostly match or augment those in GG1. In Windows, at least, the buttons have and underline under the character that will work as a shortcut when focus is on the dialog (where it remains during the process). No need for Cmd or Ctrl, just the letter, e.g. "j" for "Join". Some have an additional shortcut, e.g. the Blank, Section, Chapter ones also allow you to press the key for the number of blank lines i.e. 1, 2 or 4. Also, the normal undo & redo shortcuts (like Ctrl+Z) work as shortcuts to the Undo/Redo buttons in addition to "u" and "e".
7. The "Auto" setting is stored in the prefs file. Note that with "Auto Fix", it doesn't start fixing as soon as the dialog is popped, because users found that disconcerting when GG1 did it. So, just like GG1, you need to do something in the dialog, typically click Refresh or the "Auto Fix" radio button, but it could be one of the manual fix buttons, like Delete, which it will first do, then continue automatically.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -22,7 +22,7 @@ from guiguts.mainwindow import (
     ErrorHandler,
 )
 from guiguts.misc_dialogs import PreferencesDialog
-from guiguts.misc_tools import basic_fixup_check
+from guiguts.misc_tools import basic_fixup_check, page_separator_fixup, PageSepAutoType
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
 from guiguts.root import root
@@ -294,6 +294,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(PrefKey.WRAP_INDEX_MAIN_MARGIN, 2)
         preferences.set_default(PrefKey.WRAP_INDEX_WRAP_MARGIN, 8)
         preferences.set_default(PrefKey.WRAP_INDEX_RIGHT_MARGIN, 72)
+        preferences.set_default(PrefKey.PAGESEP_AUTO_TYPE, PageSepAutoType.AUTO_FIX)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -438,6 +439,8 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         menu_tools.add_button("PP~txt", lambda: pptxt(self.file.project_dict))
         menu_tools.add_button("~Jeebies", jeebies_check)
+        menu_tools.add_separator()
+        menu_tools.add_button("Fixup ~Page Separators", page_separator_fixup)
         menu_tools.add_separator()
         menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
         menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -367,16 +367,24 @@ class PageSeparatorDialog(ToplevelDialog):
         if maybe_hyphen == "*":
             prev_eol = f"{prev_eol} -1c"
             maybe_hyphen = maintext().get(f"{prev_eol}-1c", prev_eol)
-
-        # If word is hyphenated, join previous end of line to next beg of line
+        # Check for emdash (double-hyphen)
+        maybe_second_hyphen = ""
         if maybe_hyphen == "-":
-            # Adjust effective end of line position depending if hyphen is being kept
-            if not keep_hyphen:
-                prev_eol = f"{prev_eol} -1c"
-            # Omit any "*" at beginning of next line
-            next_bol = f"{sep_range.start.index()} linestart"
-            if maintext().get(next_bol, f"{next_bol}+1c") == "*":
-                next_bol = f"{next_bol}+1c"
+            maybe_second_hyphen = maintext().get(f"{prev_eol}-2c", f"{prev_eol} -1c")
+        # Adjust effective end of line position depending if hyphen is being kept
+        # (always keep emdash)
+        if not keep_hyphen and maybe_hyphen == "-" and maybe_second_hyphen != "-":
+            prev_eol = f"{prev_eol} -1c"
+        # If emdash at start of second page, also want to join,
+        # even if no hyphen at end of first page.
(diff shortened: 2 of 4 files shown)


cpeel (approved):

srjfoo:
As with the find feature elsewhere, I'm seeing that when this stops at a line, it's at the bottom of the page, so that any context following the page separator is lost. It seems more appropriate for the "cursor" to be in the middle of the page, but I'd be happy with just a line or three of context following the page separator.

When joining hyphenated words -- it looks like:
* only words where there is a hyphen on the first page will be joined (that is, if an em-dash separates phrases across a page boundary, if the em-dash is at the beginning of the second page, no joining takes place.
* If it's an em-dash (i.e. 2 hyphens) instead of a hyphen, you have to use "Join, Keep Hyphen", or it removes a single hyphen.
* Is it expected that no-wrap and bq are not joined across page dividers?

Most of the diffs I got in testing the three different approaches could be accounted for by user error. I double checked this one, by extracting the place where I saw the difference and running all three approaches on matching copies (zip below, including the original I started with). Only full auto removed the interior italics markup when joining a phrase marked up across the page boundary. (I'm assuming that that's to be expected. Either way, all three should give consistent results?
[godey.zip](https://github.com/DistributedProofreaders/guiguts-py/files/15328650/godey.zip)

windymilla:
@srjfoo, is [this new issue](https://github.com/DistributedProofreaders/guiguts-py/issues/244) sufficient to describe what is going wrong with the visibility of lines - it didn't appear to be specific to Page Sep Fixup, hence the new issue.

windymilla:
> When joining hyphenated words -- it looks like:
> 
> * only words where there is a hyphen on the first page will be joined (that is, if an em-dash separates phrases across a page boundary, if the em-dash is at the beginning of the second page, no joining takes place.
> * If it's an em-dash (i.e. 2 hyphens) instead of a hyphen, you have to use "Join, Keep Hyphen", or it removes a single hyphen.
> * Is it expected that no-wrap and bq are not joined across page dividers?
> 
That is how it's coded - I was quite conservative. However, I think the first one (em-dash at start of second page) would be safe to join. The second one is an oversight - I'll look at that. The third one is correct. IMO, that should really have a human eye to check that the reason for the block markup (of whatever type) is the same and that it would be appropriate to remove the surplus markup.

> Most of the diffs I got in testing the three different approaches could be accounted for by user error. I double checked this one, by extracting the place where I saw the difference and running all three approaches on matching copies (zip below, including the original I started with). Only full auto removed the interior italics markup when joining a phrase marked up across the page boundary. (I'm assuming that that's to be expected. Either way, all three should give consistent results? [godey.zip](https://github.com/DistributedProofreaders/guiguts-py/files/15328650/godey.zip)
Although I only coded italic removal in the auto case, I agree that it would be preferable if it was also removed in the other cases.

srjfoo:
> @srjfoo, is [this new issue](https://github.com/DistributedProofreaders/guiguts-py/issues/244) sufficient to describe what is going wrong with the visibility of lines - it didn't appear to be specific to Page Sep Fixup, hence the new issue.

Yes, I think that's good enoufh. If you'd like, I can add a screenshot. And that is correct. I have seen the same issue with other tools.

windymilla:
No need for a screenshot. The built-in `see` method only claims to make the line visible, so if it's already visible (even at bottom of page), it will do nothing. I'm not sure what it does (or even if it's the same across platforms) if the line is just off the page (above or below). I'll devise cases to test it out - if I suspect that I need the results of running them on macOS, I'll holler.

windymilla:
@srjfoo - I think all the points you mentioned have been addressed in the latest commit:
1. Cross-page markup is now removed properly
2. Emdashes don't get the second hyphen trimmed
3. Auto-join does join in the emdash case, just not in the hyphen case.

srjfoo:
> @srjfoo - I think all the points you mentioned have been addressed in the latest commit:
> 
> 1. Cross-page markup is now removed properly
> 2. Emdashes don't get the second hyphen trimmed
> 3. Auto-join does join in the emdash case, just not in the hyphen case.

One final question -- and apologies for not thinking to ask it earlier. The blank line before the first page separator. Is that supposed to be automatically removed when the first page separator is removed, or is the PPer expected to special-case that? I don't remember for sure, but I think auto-fix does remove it, but the other two don't. One of the others I removed manually, and the other, I didn't, so when I was diffing the files I ran, that showed up as a diff.

windymilla:
> One final question -- and apologies for not thinking to ask it earlier. The blank line before the first page separator. Is that supposed to be automatically removed when the first page separator is removed, or is the PPer expected to special-case that? I don't remember for sure, but I think auto-fix does remove it, but the other two don't. One of the others I removed manually, and the other, I didn't, so when I was diffing the files I ran, that showed up as a diff.

I don't think any of the options remove it automatically, not even auto-fix. At least it didn't for me just now. That's a special case, and the PPer will want blank lines at the top anyway most likely, so I don't think I'll change it.

srjfoo:
> > One final question -- and apologies for not thinking to ask it earlier. The blank line before the first page separator. Is that supposed to be automatically removed when the first page separator is removed, or is the PPer expected to special-case that? I don't remember for sure, but I think auto-fix does remove it, but the other two don't. One of the others I removed manually, and the other, I didn't, so when I was diffing the files I ran, that showed up as a diff.
> 
> I don't think any of the options remove it automatically, not even auto-fix. At least it didn't for me just now. That's a special case, and the PPer will want blank lines at the top anyway most likely, so I don't think I'll change it.

Okay. In the example I was using, that left 5 blank lines.

FWIW, I think GG1 doesn't do any better. At least it didn't for my quick test. 99% auto matches GG2, the others don't.

srjfoo (approved):

PR #233, opened 2024-05-10
Theming

windymilla:
Use Edit->Preferences to change the theme


windymilla:
@cpeel - no need for a detailed look. I've only spent a short time on this so far, and would like your opinion on whether it is good/ok to rely on an external module like `ttkbootstrap`?

@srjfoo - just a first attempt for you to have a little play with. Trying to decide if it's a good way forward and/or worth pursuing further. Thanks

cpeel:
> @cpeel - no need for a detailed look. I've only spent a short time on this so far, and would like your opinion on whether it is good/ok to rely on an external module like `ttkbootstrap`?

I think it's perfectly fine to rely on something like `ttkboostrap`. The things I generally look at when pulling in a new module are: is it maintained and what dependencies is it pulling in. The `ttkbootstrap` repo has had commits within the last year and the only dependency is `pillow` which is one of GG2's dependencies already 👍

windymilla:
> I think it's perfectly fine to rely on something like `ttkboostrap`. The things I generally look at when pulling in a new module are: is it maintained and what dependencies is it pulling in. The `ttkbootstrap` repo has had commits within the last year and the only dependency is `pillow` which is one of GG2's dependencies already 👍

Thanks - experimental atm, and I'd also like to code it so that it's possible to run GG2 without any of the ttkbootstrap theming. Unfortunately, that doesn't seem to be an option once you actually make use of it (like choose theme=None) - it just uses its own default theme. So, will need some more sneakiness in the code.

srjfoo:
It will come as no surprise to you that I'm not overly fond of any of the light themes. 🤣 

However, one comment about (at least most of) the light themes on at least macOS with the OS system dark theme enabled: the line numbers don't show up, because the text is white on white.

I haven't actually done an in-depth look at any of them to see how all the widgets look (I did pull up s/r in `solar`, and it looked pretty reasonable, but I was just doing some cursory checking.

All in all, looks pretty good.

windymilla:
Abandoned this method of theming, since it takes over and doesn't allow easy non-themed theme!

PR #235, opened 2024-05-11
Add search term to history during replace (all)

### LINKED ISSUE #234: Replace All doesn't add Search string to Search history
Type a new string into Search box, and a replacement into the Replace box, then click Replace All. Only the replacement gets added to the history pulldown, not the search string.

windymilla:
Replace All did not add the search term to the search history.

Fixes #234


srjfoo (approved):
Confirmed that search term is added to search history, and replacement term is added to replace history.

PR #236, opened 2024-05-13
Add text file "Unmatched..." checks

windymilla:
1. Unmatched block markup (copes with `/#` being nestable but other types not nestable). 
2. Unmatched brackets (not nestable)
3. Unmatched curly quotes (not nestable)
4. Unmatched DP markup (not nestable)
5. "Nestable" above means that it will permit `/#` to follow `/#` without an intervening `#/`, for example. But it complains about `[` followed by `[` without intervening `]`. It would be possible to change which are nestable and which are not, but this matches GG1 behavior.
6. First enhancement compared to GG1: you can use upper/lower case indiscriminately for block markup  - in GG1, if you use upper case to open, e.g. `/C`, you must use upper case to close.
7. Second enhancement is that it checks that the matching block markup is at the start of a line. In GG1, if you open with `/C`, then close with `     /C` it will not complain.
8. Also, you can sort errors either by row & column, or by type (e.g. all `/c` errors will precede all `/r` errors). I could change the type sorting if my arm were twisted enough. At the moment, it will list all unmatched open markup before all unmatched close, then within those two groups, it is sorted alphabetically by the type of markup. As I say, that could be changed quite easily, but I think it's OK as it is really - I'll leave it to the intrepid testers to decide if they think it's worth it.
9. Menu entries are all in the Tools menu.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -453,7 +453,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
         menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
         menu_tools.add_separator()
-        menu_tools.add_button("~Unmatched DP Markup", unmatched_markup_dp)
+        menu_tools.add_button("~Unmatched DP Markup", unmatched_dp_markup)
         menu_tools.add_button("Unmatched Bloc~k Markup", unmatched_block_markup)
         menu_tools.add_button("Unmatched ~DP Markup", unmatched_dp_markup)
         menu_tools.add_button("Unmatched ~Brackets", unmatched_brackets)
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -22,7 +22,12 @@ from guiguts.mainwindow import (
     ErrorHandler,
 )
 from guiguts.misc_dialogs import PreferencesDialog
-from guiguts.misc_tools import basic_fixup_check, page_separator_fixup, PageSepAutoType
+from guiguts.misc_tools import (
+    basic_fixup_check,
+    page_separator_fixup,
+    PageSepAutoType,
+    unmatched_markup_dp,
+)
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
 from guiguts.root import root
@@ -444,6 +449,8 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_tools.add_separator()
         menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
         menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
+        menu_tools.add_separator()
+        menu_tools.add_button("~Unmatched DP Markup", unmatched_markup_dp)
 
     def init_view_menu(self) -> None:
         """Create the View menu."""
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -22,7 +22,7 @@ from guiguts.mainwindow import (
     ErrorHandler,
 )
 from guiguts.misc_dialogs import PreferencesDialog
-from guiguts.misc_tools import basic_fixup_check
+from guiguts.misc_tools import basic_fixup_check, page_separator_fixup, PageSepAutoType
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
 from guiguts.root import root
@@ -294,6 +294,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(PrefKey.WRAP_INDEX_MAIN_MARGIN, 2)
         preferences.set_default(PrefKey.WRAP_INDEX_WRAP_MARGIN, 8)
         preferences.set_default(PrefKey.WRAP_INDEX_RIGHT_MARGIN, 72)
+        preferences.set_default(PrefKey.PAGESEP_AUTO_TYPE, PageSepAutoType.AUTO_FIX)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -438,6 +439,8 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         menu_tools.add_button("PP~txt", lambda: pptxt(self.file.project_dict))
         menu_tools.add_button("~Jeebies", jeebies_check)
+        menu_tools.add_separator()
+        menu_tools.add_button("Fixup ~Page Separators", page_separator_fixup)
         menu_tools.add_separator()
         menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
         menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -26,7 +26,10 @@ from guiguts.misc_tools import (
     basic_fixup_check,
     page_separator_fixup,
     PageSepAutoType,
-    unmatched_markup_dp,
+    unmatched_dp_markup,
+    unmatched_brackets,
+    unmatched_curly_quotes,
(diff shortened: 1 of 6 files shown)


windymilla:
@cpeel - if you have time for a quick look over the structure of the new code in `misc_tools.py`, particularly how `unmatched_markup_check` takes two function arguments and optional regexes. It makes sense to me as `unmatched_markup_check` can then be used for all the checks, but there may be a better way. If you don't get to it, don't worry.

G4OEU:
Refreshed the current branch, pr/231, to do this #236 review. Out of curiosity I wanted to see how GG2 dealt with an opening `/##` marker. Got some odd behaviour. See first image after which I did "Rewrap All" and got the GG2 state shown in the second image.
GG2-bug_cropped

GG2-bug-2of2_cropped

What is happening in the two images above maybe because I deleted some text from the loaded file. To see this happening in a repeatable way it is best to start with the following, then save it, then do "Rewrap All". The error message flags a line that no longer exists in the displayed file contents - see image below.
```
/##
  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed
        varius vehicula
  /c
  ----
  XX
  c/
#/
```
GG2-bug-3rd_cropped

G4OEU:
Oops! I should have made my current branch pr/236. Have done so now. But what's happended to the "Rewrap All" and "Rewrap Selection" options in the Tools menu?

I can see the 4 new 'Unmatched markup/brackets' options but no rewrap options.

windymilla:
> Oops! I should have made my current branch pr/236. Have done so now. But what's happened to the "Rewrap All" and "Rewrap Selection" options in the Tools menu?
> 
> I can see the 4 new 'Unmatched markup/brackets' options but no rewrap options.

Each PR is based on the current `master` branch. So, until the work in a PR is merged, you won't normally see it in another PR. Since #231 has not finished testing by Sharon, and so hasn't been merged, its rewrap features are not in #236. After some work has been merged, it is possible to "rebase" a branch on the updated master. For example, when #231 is merged which produces an updated master, #236 could be rebased on the new master, i.e. the commits in #236 could be applied to a copy of the updated master. After rebasing, #236 would then have any new work that had been merged into master plus its own work. 
If there is some significant work merged, I quite often rebase any outstanding PRs, partly because it helps keep things simpler when those PRs are eventually merged, but also to stop testers from saying "I saw that XXX bug again - I thought you'd fixed it".

windymilla:
Regarding your original query about `/##`, @G4OEU, I think it is explained by the code not recognising `/##` as valid blockquote markup (which it isn't), and so just treats it as a normal word and wraps the whole block as normal. Since the `/c` markup is not at the start of a line, it is also not valid markup, so it gets ignored too. I think that it is desirable for the markup checks to be quite strict, or we get situations where normal text gets interpreted as markup, for example I had some `</i>` markup accidentally interpreted as the start of index markup `/i` in an early version of some of my code.

G4OEU:
> Regarding your original query about /##, @G4OEU, I think it is explained by the code not recognising /## as valid blockquote markup (which it isn't), ...

Fair enough, but what about the error message quoting line 8 which is no longer seen in the refreshed GG window with the rewrapped text.

windymilla:
> Fair enough, but what about the error message quoting line 8 which is no longer seen in the refreshed GG window with the rewrapped text.

That's a tricky one - it has found the faulty `#/` which is an error because it's the first valid piece of markup in the file, and is a close blockquote when none was open. It was on line 8 before you rewrapped, but due to the wrapping has now moved. I think generally, it will only move by the amount that the paragraph grows/shrinks during rewrapping (in this extreme case 5 lines, but typically one or two lines). You'll noticed the message says "near line 8" not "at line 8" because I was aware of this limitation, which  isn't that easy to remove.

G4OEU:
Hmmh, I can see the difficulty. Given that the line number at which a fault is located will likely change/disappear as a result of the rewrapping, couldn't the error message go a little further in its wording to acknowledge that limitation? For example it could say something like:
`Block quote markup error near line 8 in unwrapped text [use Ctrl/Z to restore]` 
or
`Block quote markup error near line 8 before wrapping`
etc.

windymilla:
> couldn't the error message go a little further in its wording to acknowledge that limitation?

Possibly - I did also consider forcing an undo when there was an error, but thought that might hide the error when bad rewrapping might reveal it.

G4OEU:
If the code does not recognise `/##` as a valid blockquote markup then "Unmatched Block Markup" should flag the closing `#/` in the following but isn't.

```
/##
  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed
        varius vehicula
  /c
  ----
  XX
  c/
#/
```

windymilla:
> "Unmatched Block Markup" should flag the closing `#/` in the following but isn't.

Yes, it should. I was trying to avoid all the complicated alternatives, like `/#[27.5,80]` but looks as though I'll need to do something about it.

srjfoo:
There are two "unmatched DP markup" entries in the tools menu (at least on the copy I have :D

windymilla:
> Hmmh, I can see the difficulty. Given that the line number at which a fault is located will likely change/disappear as a result of the rewrapping, couldn't the error message go a little further in its wording to acknowledge that limitation? For example it could say something like: `Block quote markup error near line 8 in unwrapped text [use Ctrl/Z to restore]` or `Block quote markup error near line 8 before wrapping` etc.

@G4OEU - I began work on this to change the error messages, and then realized why I'd done what I did. The problem is that the line number isn't the unwrapped line number. It's the line number that the opening markup was on when the wrapping code got to that point in the file. I'll look to see if there's a simple way to get a more accurate line number or whether it will just have to continue to say "near..."

windymilla:
Good news - I think I've got the exact line number within the wrapped text, so no need for "near" or to mention "undo".

windymilla:
@G4OEU & @srjfoo - I think all your queries are resolved now. The latest commits:
1. Remove duplicate menu entry
2. Improve line number in error message when error in rewrap
3. Match the tests done in this PR with the regex used in the rewrap code so they are (fairly) consistent - it may still be possible to construct obscure edge cases, but hopefully this will be satisfactory.

srjfoo:
Attaching my test file: the original seeded version, plus my fixup run (only the fixup run has a `.json` file--I did the seeding in BBEdit), plus the notes file documenting roughly where the seeding took place and what I mucked with.

The two that didn't behave as I expected were a malformed closing bq, and a malformed opening bq (the column spec was malformed).

[text-file-fixup.zip](https://github.com/DistributedProofreaders/guiguts-py/files/15359850/text-file-fixup.zip)

windymilla:
> Attaching my test file: the original seeded version, plus my fixup run (only the fixup run has a `.json` file--I did the seeding in BBEdit), plus the notes file documenting roughly where the seeding took place and what I mucked with.
> 
> The two that didn't behave as I expected were a malformed closing bq, and a malformed opening bq (the column spec was malformed).
Since the two malformed ones are not recognised as bq markup, and one is a close and the other is an open, the remaining correctly formed markup is actually paired correctly. If you do a rewrap of the whole file, you can see that it just treats `#/` on the end of a line as normal characters. Similarly for the `/#[8.r]`- it's not bq markup, so it just gets wrapped. That is a deliberate decision - the regex used only matches valid markup, and is the same regex in the wrap code as in the checking code. 

It would be possible (under a separate PR, though) to add a further test to the "Unmatched block markup" that looks for things that might have been intended to be block markup, but which weren't. Where you draw the line once you allow errors in the markup is obviously a tricky decision. For example, should `#[2]` be reported as faulty markup, or `/[3.2]`? Or maybe only check things that actually have `/#` or `#/`? If you do that, would you do the same for the other checks, e.g. when checking for italic, would you issue warnings for `i>5` (which could be some math, or a malformed `<i>5...`? You see why I'd like to consider it under a different PR?

srjfoo (approved):

srjfoo:
> You see why I'd like to consider it under a different PR?

Absolutely. Those were unexpected results; I mentioned them because I didn't know if they should have been detected or not.

PR #241, opened 2024-05-14
Improve column selection

### LINKED ISSUE #239: Column-select enhancement request
In BBEdit, and macOS's Terminal program, I can click to set the placement of the starting cursor, move to the end of the column-select range, hold down shift-option, and click at the ending location, and the column-select is done.

Sublime Text does not seem to do this (though it will allow a simple column-select); VSCode doesn't have a keyboard shortcut for a column select, but a separate menu item which, when turned on, makes any selections into column selections).

Possibly the best alternative would be to implement what GG1 has: set the beginning and end points of where the column select should be by clicking and dragging (or the equivalent), then use a keyboard shortcut on the Selection button in the taskbar (for me on macOS, it's `shift`+`right-click`).

Note that this falls into the "nice-to-have-but-not-critical" refinement on what we have already category (once #237 is fixed :D).

windymilla:
1. Allow user to Shift-click with Option/Alt key pressed to make a column selection. So, user can click at start, then Shift-click to make a regular selection. Or, click at start, then Shift-Alt-Click (use Option instead of Alt on Macs) at end to make a column selection. Multiple Shift-Alt-Clicking may not do what you expect due to ambiguity over the start/end points of the selection - you have been warned!
2. User can Shift click the Selection label in the status bar to switch selection from a regular selection to a column selection.

Fixes #239

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -557,12 +557,15 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
 
         the_statusbar.add(
             "selection",
-            tooltip="Click: Restore previous selection",
+            tooltip="Click: Restore previous selection\nShift Click: Toggle column/regular selection",
             update=selection_str,
             width=16,
         )
         the_statusbar.add_binding(
             "selection", "ButtonRelease-1", maintext().restore_selection_ranges
+        )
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
Beautiful!

PR #243, opened 2024-05-16
Hebe fix bug

windymilla:
The he/be fixing code (via Ctrl-click) only worked for lowercase he/be.
Only "be" was converted to "he". Any other he/be word ("Be", "HE", "bE", etc.) was converted to lowercase "be".

Also (since most, but not all, in jeebies code) silence warnings from latest version of pylint, mostly by passing arguments into nested helper functions, rather than relying on access to variables defined in the outer function. It does make the functions clearer too, IMO.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -739,6 +739,7 @@ class MainText(tk.Text):
         minrow = max(minrow, toprow)
         maxrow = max(anchor_rowcol.row, cur_rowcol.row)
         maxlen = -1
+        y_maxlen = -1
         for row in range(minrow, maxrow + 1):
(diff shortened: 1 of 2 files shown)


G4OEU:
Toggling with `Ctrl`+`left-click` now works correctly to maintain the case of the first letter of 'he' or 'be' that is toggled.

Other tool tip commands continue to work as expected.

Compared pr/43 jeebies.py with current master jeebies.py and noted the updates to the former. Both versions of Jeebies report the same number of 'errors' in each of the three sensitivity modes. 

Approved this PR.

srjfoo (approved):
LGTM

PR #245, opened 2024-05-16
Position "seen" line better in window

### LINKED ISSUE #244: Visibility of line of interest
When the code shows a line (`see` method) it seems that, at least on macOS, it is happy to have that line at the bottom or top of the screen, rather than in/near the middle. This is not very helpful for the user who wants to see the context around the line.

Dev note: consider calling `see` a few lines above the line of interest, then a few lines below, then the line of interest. "A few" could possibly be half the visible window height, or maybe just a fixed number (10?) would do.

windymilla:
Operations like Search make the found line visible. However the built-in routine that does this only make it visible, not central, so it may be the first or last line in the window, meaning user has no context lines around it.

Fixed by repeatedly seeing lines ever further from the desired line up to half the window height, which leaves the desired line in the middle of the window.

Fixes #244

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -416,6 +416,17 @@ class MainText(tk.Text):
             focus: Optional, False means focus will not be forced to maintext
         """
         self.mark_set(tk.INSERT, insert_pos.index())
+        # The `see` method can leave the desired line at the top or bottom of window.
+        # So, we "see" lines above and below desired line incrementally up to
+        # half window height each way, ensuring desired line is left in the middle.
+        # If performance turns out to be an issue, consider giving `step` to `range`.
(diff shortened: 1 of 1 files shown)


windymilla:
@srjfoo, although this branch does not have page separator fixup work in yet, you reported the same issue with the Search dialog. The routine used is the same, so if this works when Searching, it should also work everywhere else.

srjfoo (approved):
Looks good to me.

Hopefully, performance won't be an issue. Just to give it a good workout, I searched for `\bthe\b`, and I didn't see a noticeable delay in the results list coming up (and no, I didn't click on every one of the 3000-odd results :D). Not every result was in the exact middle of the page, but all that I clicked on gave plenty of context.

PR #246, opened 2024-05-17
Themes2

windymilla:
This uses the standard themes, but then with a couple of added
ones. This method means you can essentially have the non-themed
default, or any of the other (dreadful) ttk themes, or the two
added ones, which are called awdark and awlight.
I think I could download and add other themes like these.


srjfoo:
> This uses the standard themes, but then with a couple of added ones. This method means you can essentially have the non-themed default, or any of the other (dreadful) ttk themes, or the two added ones, which are called awdark and awlight. I think I could download and add other themes like these.

Define "standard themes"? I don't see anything that gives me the non-themed default. At least not on macOS. Does "default" give you that on Windows? It doesn't me -- it forces a light theme. (And the checkboxes remind me of GG1 :D) ... come to think of it, so do the scrollbars, so maybe that's what "default" means: tk default? Interestingly, 6 of the 7 immediately available themes do bear a resemblance to GG1. The only one that doesn't is (for me) the aqua theme (which I believe is specific to Macs?). The main thing I object to here is that the search results pages all have a white background with black text:

image

tangledhelix:
> Define "standard themes"? I don't see anything that gives me the non-themed default. At least not on macOS. Does "default" give you that on Windows? It doesn't me -- it forces a light theme. (And the checkboxes remind me of GG1 :D) ... come to think of it, so do the scrollbars, so maybe that's what "default" means: tk default?

I had the same thought, and I'm pretty sure it's the default for Tk/tkinter, or the themes module Nigel used here.

windymilla:
Closed, since restarted with "3rd attempt" using lessons learnt from this.

PR #251, opened 2024-05-21
Allow selection in checker dialog with arrow keys

windymilla:
If the focus is on the checker dialog, and up/down arrow is pressed, it should select the previous/next line in the dialog.

On Macs, selecting a line didn't force focus to the main text window anyway. This commit also doesn't force focus for other platforms if the selection was changed by arrow key (otherwise, you have to refocus on the checker dialog to hit the arrow key again).

Note that Ctrl+arrow key, etc., do not do the processing, fixing, deleting all similar, etc., that can be done with the mouse. This is only for browsing through the errors.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -181,6 +181,8 @@ class CheckerDialog(ToplevelDialog):
             "Shift+Cmd/Ctrl+3",
             lambda event: self.process_remove_entry_by_click(event, all_matching=True),
         )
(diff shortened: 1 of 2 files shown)


srjfoo:
I couldn't use the arrows in Word Frequency, but everything else worked great!

windymilla:
> I couldn't use the arrows in Word Frequency, but everything else worked great!

WF isn't a checker dialog, which is why it doesn't work there. I will see if I can easily make the same edit there, since it would be better if it works the same.

srjfoo (approved):

PR #252, opened 2024-05-21
Add Pref for Tear-Off menus

windymilla:
Note that GG2 will need to be restarted after changing this Pref. All the menus would have to be deleted and recreated, which is just not worth the time or code complexity for something that most users will use once (or never).

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -319,6 +319,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
                 theme_name_internal_from_user(value)
             ),
         )
+        preferences.set_default(PrefKey.TEAROFF_MENUS, False)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -9,7 +9,7 @@ from guiguts.preferences import (
     PersistentInt,
     PersistentString,
 )
-from guiguts.utilities import is_windows
+from guiguts.utilities import is_mac
 from guiguts.widgets import ToplevelDialog, ToolTip
 
 
@@ -43,7 +43,7 @@ class PreferencesDialog(ToplevelDialog):
             variable=PersistentBoolean(PrefKey.TEAROFF_MENUS),
         )
         tearoff_check.grid(column=0, row=2, sticky="NEW", pady=5)
-        if is_windows():
+        if is_mac():
             tearoff_check["state"] = tk.DISABLED
             ToolTip(tearoff_check, "Not available on macOS")
         ttk.Checkbutton(
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1,5 +1,6 @@
 """Miscellaneous dialogs."""
 
+import tkinter as tk
 from tkinter import ttk
 
 from guiguts.preferences import (
@@ -8,6 +9,7 @@ from guiguts.preferences import (
     PersistentInt,
     PersistentString,
 )
+from guiguts.utilities import is_windows
 from guiguts.widgets import ToplevelDialog, ToolTip
 
 
@@ -35,11 +37,15 @@ class PreferencesDialog(ToplevelDialog):
         ttk.Label(
             appearance_frame, text="(May need to restart program for full effect)"
         ).grid(column=0, row=1, sticky="NSEW")
-        ttk.Checkbutton(
+        tearoff_check = ttk.Checkbutton(
             appearance_frame,
             text="Use Tear-Off Menus (requires restart)",
             variable=PersistentBoolean(PrefKey.TEAROFF_MENUS),
-        ).grid(column=0, row=2, sticky="NEW", pady=5)
+        )
+        tearoff_check.grid(column=0, row=2, sticky="NEW", pady=5)
+        if is_windows():
+            tearoff_check["state"] = tk.DISABLED
+            ToolTip(tearoff_check, "Not available on macOS")
         ttk.Checkbutton(
             appearance_frame,
             text="Display Line Numbers",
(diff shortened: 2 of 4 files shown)


srjfoo:
I guess I need some clarification: this is just adding the preference? There are currently no tear-off menus?

tangledhelix:
> I guess I need some clarification: this is just adding the preference? There are currently no tear-off menus?

There aren't on Mac, because the Mac menu system has no way to do that. The tear-offs are working correctly on Ubuntu, so far as I can tell, and their availability is properly controlled by the preference.

I suggest hiding this preference from the prefs dialog on Mac, as it will be confusing and have no effect. On Mac, so far as I can see, the preference has no effect on operation or appearance at all.

tangledhelix:
Ubuntu screenshot:

Screenshot 2024-05-22 at 12 40 39 AM

srjfoo:
> > I guess I need some clarification: this is just adding the preference? There are currently no tear-off menus?
> 
> There aren't on Mac, because the Mac menu system has no way to do that. The tear-offs are working correctly on Ubuntu, so far as I can tell, and their availability is properly controlled by the preference.
> 
> I suggest hiding this preference from the prefs dialog on Mac, as it will be confusing and have no effect. On Mac, so far as I can see, the preference has no effect on operation or appearance at all.

Thanks, I was wondering. This is ironic, since it was my suggestion, and my goal was to get the unmatched markup options out of a linear menu so that I don't (or rather Mac users don't) have to go to the menu for each check.

tangledhelix:
> Thanks, I was wondering. This is ironic, since it was my suggestion, and my goal was to get the unmatched markup options out of a linear menu so that I don't (or rather Mac users don't) have to go to the menu for each check.

We could still do that (platform-agnostic) by using a dialog instead of a menu. There are numerous examples of that in GG1, I think:

- Word Frequency dialog
- Txt Conversion Palette
- HTML Markup
- Unicode picker

Another option, although I don't have any idea how complex it would be in Tk (or how platform-agnostic) might be dockable tools, i.e. the sort of thing Photoshop uses for tool panels. (I'll find screenshots if it's not apparent what I mean.)

Side note — I don't want to be prescriptive, but I will share something I found on [tkdocs.com][1], whose front page header is "Modern Tk Best Practices":

> **Before you Start**
>
> It's essential to put the following line in your application somewhere before you start creating menus.
>
> `root.option_add('*tearOff', FALSE)`
>
> Without it, each of your menus (on Windows and X11) will start with what looks like a dashed line and allows you to "tear-off" the menu, so it appears in its own window. You should eliminate tear-off menus from your application as they're not a part of any modern user interface style.
>
> >  _This is a throw-back to the Motif-style X11 that Tk's original look and feel were based on. Get rid of them unless your application is designed to run only on that old box collecting dust in the basement. We'll all look forward to a future version of Tk where this misguided paean to backward compatibility is removed._
> >
> > _While on the topic of ancient history, the option_add bit uses the option database. This provided a standardized way to customize some aspects of X11 user interfaces through text-based configuration files, but it's no longer used today. Older Tk programs may use the option command internally to separate style configuration options from widget creation code. That approach pre-dated themed Tk styles, which should be used for that purpose today. However, it's somehow fitting to use the obsolete option database to automatically remove the obsolete tear-off menus._

[1]: https://tkdocs.com/tutorial/menus.html

Take that as you will - I'm not here to convince anyone to keep them or drop them. Just something I found in my travels while looking into this. I was hoping to find documentation plainly stating tearOff didn't work on Mac (which I did not find, although this text seems to allude to the lack of tearOff support on Mac).

srjfoo:
> We could still do that (platform-agnostic) by using a dialog instead of a menu. There are numerous examples of that in GG1, I think:
> 
> * Word Frequency dialog
> * Txt Conversion Palette
> * HTML Markup
> * Unicode picker

You beat me to it. I was literally just about to bring up WF as an example of using that layout option, and ask if we could have something like that. You said it in a great more detail than I would have been able to, though. Thanks!

---

Interesting advice on tear-offs in Tk. It would certainly level the playing field, wouldn't it?

windymilla:
> Side note — I don't want to be prescriptive, but I will share something I found on [tkdocs.com](https://tkdocs.com/tutorial/menus.html), whose front page header is "Modern Tk Best Practices":

It was because of reading that document (before I started coding GG2) that I had that line of code in GG2 pretty much from the beginning:
`root.option_add('*tearOff', False)`

At the time I was slightly conflicted. On the one hand, it's good to follow "modern best practices", and it's good to have be platform-agnostic. On the other hand, we're not trying to sell a commercial product that needs to compete against other tools with a modern style, and why should Windows & Linux GG users (who outnumber Mac users by a considerable margin) not have a feature which I know some/several/many(?) PPers like in GG1, just because Mac users can't have it?

This is why I made it a Pref (which I've tried to avoid adding as much as possible) rather than just setting Tear-Off to be True. You could argue that people who have it on their platform but don't want to use it, aren't suffering much - they just have the little tear-off line at the top of each menu. However, to allow those who want a "modern best practice" interface, but cater for those who might use it, having a Pref seems like a reasonable compromise.

I agree that the Pref should not appear for Mac users (or maybe have it disabled, so that we don't get complaints that the Prefs dialog doesn't match a screenshot in the manual?). It could have a tooltip (need to check if they work when a widget is disabled) that explains what it does and says that it's not available on macOS.

windymilla:
> We could still do that (platform-agnostic) by using a dialog instead of a menu. There are numerous examples of that in GG1, I think:
> 
> * Word Frequency dialog
> * Txt Conversion Palette
> * HTML Markup
> * Unicode picker

Although there may be some dialogs which are essentially menus in disguise, I'm not sure I agree that those examples really are in that category. WF, for example, has a text field, and a list of words, so couldn't just be a menu. Similarly for Txt Conversion & HTML Markup.

I couldn't find an obvious dialog in GG1 that essentially consisted of buttons, but I agree that there could be cases, like the multiple "unmatched" checks, that could be accessible via a dialog/palette of buttons, rather than a menu or sub-menu.

srjfoo (approved):
😁 -- I'm tempted to get into an involved discussion about menus, related items, etc. But I won't. Not right now, anyway.

Showing the option greyed out for Macs seems the best approach at the moment.

PR #256, opened 2024-05-25
Add bookmarks

windymilla:
1. Sub menu of `Search` menu to set/goto bookmarks
2. Keyboard shortcuts like GG1: `Cmd/Ctrl+1`, `2`, ...`5` to go to, and `Shift+Cmd/Ctrl+1` to set the bookmark.
3. Brief highlight (like GG1) to show it's been set/found (bell if not)
4. Bookmarks stored in project .json file.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -465,16 +465,18 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             ("percent",),
         ]
         # Shift+Cmd+number no good for Mac due to clash with screen capture shortcuts
-        modifier = "Option" if is_mac() else "Shift"
+        # Ctrl+number could also clash on Mac with virtual desktop switching, but
+        # the best option we have that works at the moment.
+        modifier = "Shift"
         for bm, keys in enumerate(shortcuts, start=1):
             bookmark_menu.add_button(
                 f"Set Bookmark {bm}",
                 lambda num=bm: self.file.set_bookmark(num),  # type:ignore[misc]
-                f"{modifier}+Cmd/Ctrl+Key-{bm}",
+                f"{modifier}+Ctrl+Key-{bm}",
             )
             # Add extra shortcuts to cope with keyboard layout differences
             for key in keys:
-                (_, key_event) = process_accel(f"{modifier}+Cmd/Ctrl+{key}")
+                (_, key_event) = process_accel(f"{modifier}+Ctrl+{key}")
                 maintext().key_bind(
                     key_event,
                     lambda _event, num=bm: self.file.set_bookmark(  # type:ignore[misc]
@@ -487,7 +489,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             bookmark_menu.add_button(
                 f"Go To Bookmark {bm}",
                 lambda num=bm: self.file.goto_bookmark(num),  # type:ignore[misc]
-                f"Cmd/Ctrl+Key-{bm}",
+                f"Ctrl+Key-{bm}",
             )
 
     def init_tools_menu(self) -> None:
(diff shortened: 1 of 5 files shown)


tangledhelix:
For posterity, adding my notes about Mac keyboard shortcuts. I looked at any keybind where the base key is a numeral, with any modifier.

Out of the box, with no modifications, these keys are mapped:
- `Shift-Cmd-3`: Take screenshot of entire screen
- `Shift-Ctrl-Cmd-3`: Copy screenshot of entire screen to clipboard
- `Shift-Cmd-4`: Take screenshot of selected area
- `Shift-Ctrl-Cmd-4`: Copy screenshot of selected area to clipboard
- `Shift-Cmd-5`: Open screenshot and screen recording options

The above would be captured by the OS and unavailable to Guiguts.

In addition, in the Settings app, there are various keys which are pre-defined, but not activated. They can be activated there, or else in some cases they can be activated using a hotkey to toggle an Accessibility feature. I would suggest we avoid all of these as well, because we never know whether they will be active. Technically, any user can define any keybind on their own, but since these ship with the system as checkbox-toggle items, they are more likely to be in use.

- `Ctrl-1` and 2, 3, ...: Switch to virtual desktop (n)
- `Opt-Cmd-8`: Toggle screen zoom
- `Ctrl-Opt-Cmd-8`: Invert colors

Guiguts 1 only binds bookmarks 1 to 5, so Opt-Cmd is available (so long as we stick to 5, or even to at most 7). And note here that the GG1 bookmark switch keys of Ctrl-1, Ctrl-2, etc may be bound on many systems, so I'd avoid that and stick with `Cmd-1`, `Cmd-2`, ...

Attaching a few screenshots illustrating the default settings for an unmodified user.

Screenshot 2024-05-26 at 11 51 03 AM
Screenshot 2024-05-26 at 11 50 56 AM
Screenshot 2024-05-26 at 11 50 42 AM

srjfoo:
The current commit tries `shift`+`ctrl` plus the number to set and `ctrl` plus the number to go to the bookmark, I believe (I still need to test that). The next previous one tried `cmd`+`option` plus the number to set, and back to `cmd` plus the number to navigate. That was a failure, too. One of the five worked for me: `cmd` + `option` + `3` was successful, but nothing else (one of the others may possibly succeeded in setting, but not in navigating, but I'm not even sure of that). (Note that setting and navigating to bookmarks via the menu does work.)

srjfoo (approved):
> * `Ctrl-1` and 2, 3, ...: Switch to virtual desktop (n)

This may be something to be aware of (perhaps it should be documented both in the code and in the gg2 manual?). I don't use virtual desktops, so it's not an issue for me.

windymilla:
Thanks @tangledhelix for the notes about Mac system-defined shortcuts - useful to have it documented. It would certainly be better if your Cmd+Option could be made to work. However, I'm going to merge this as-is for now, since it's the only version we've been able to get working on Windows & Mac.

Please feel free to experiment further with trying different shortcuts - it's not really practical for me to do that without a Mac. Let me know if it's not clear how the shortcut keys are set up:
A string is passed to `Menu.add_button`, something like `Shift+Ctrl+Key-3`. This gets converted into a string to display in the menu (`Shift+Ctrl+3`) and a string to pass to the binding function (`Shift-Control-Key-3`). Generally, `+` is changed to `-` and some words are adapted, e.g. `Ctrl` to `Control`.

windymilla:
Rebased, rather than squashed, to preserve previous attempts.

windymilla:
> > * `Ctrl-1` and 2, 3, ...: Switch to virtual desktop (n)
> 
> This may be something to be aware of (perhaps it should be documented both in the code and in the gg2 manual?). I don't use virtual desktops, so it's not an issue for me.

I've added a note to the [GG2 Manual Notes wiki page](https://www.pgdp.net/wiki/PPTools/Guiguts2_Manual_Notes#MacOS)

PR #257, opened 2024-05-26
Fix labels / variable names in search menu

tangledhelix:
The Search menu's docstring and variable names seem to have been copy/pasted from init_view_menu; just some cleanup.


windymilla (approved):
Oops - thanks. I found a similar example of that a couple weeks back, but still didn't go through and check all of them.

Issue #43, opened 2023-12-15
Highlight Alignment Column


windymilla:
In GG1 Search menu

tangledhelix:
I'm looking into this.

tangledhelix:
Key bind note: the GG1 key binding for this is `Ctrl+Shift+a`, suggesting a Mac key bind of `Shift-Cmd-a`. Just a note to check into this later: on my Mac, this keystroke, when invoked in a text control (such as the GG2 editor window), calls the `Search man Page Index in Terminal` system service. I don't believe I activated that, so I believe it may be a default binding. Also need to check whether a keybind we set would override it, or if the system service would take precedence.

Issue #42, opened 2023-12-15
Highlight surrounding quotes/brackets


windymilla:
In GG1 Search menu

tangledhelix:
I'm looking into this.

tangledhelix:
The Perl implementation of this comes from the Perl TK module, it's not our own code. So far I haven't found a similar function in tkinter, so most likely I'll port the Perl code and make it part of the GG2 codebase itself.

See `HighlightSinglePairBracketingCursor` function defined here (starting on line 432 as of this writing):

https://metacpan.org/release/SREZIC/Tk-804.036/source/Tk/TextEdit.pm

PR #259, opened 2024-05-27
Work in progress on highlighter tickets

tangledhelix:
This is not ready to merge, by a long shot, but creating a PR for review purposes.

I'm working on #43 and #42, both of which share a couple of characteristics:
- When choosing the menu item you toggle an on/off state for the feature
- Code needs to run on some interval so that the highlight changes as the text or cursor position changes

I have the fundamentals working (tested at least on Mac, so far - testing on Win/Linux is pending):
- The menu item's on/off state works
- Saving that state to the preferences works
- That setting persists across restarts
- The callback runs when the pref is True, and does not run when it's False

My observations so far:

On startup, the preference is read; if it's False, then the callback is run only one time (due to the fact all callbacks are run on startup) but it doesn't do anything and it doesn't run again. But if True on startup, it will do exactly what it should: it runs, and it runs again every 2000ms (as designed).

If I change the preference, it turns the callback on and off like it should.

So far so good - all of this seems proper. The one thing is that, if the app starts with the pref as False, and I then turn it on, it starts to call the callback **twice** per interval, not once. This is in contrast to when the pref is True on startup. In that case, it calls the callback only **once** per interval (which is the desired behavior).

Any advice is appreciated. And for @windymilla specifically I'd like to hear if any of this works at all on Windows, as it's not tested there. (I will test on Ubuntu at some point myself, but I haven't yet.)

For now I am just trying to get the infrastructure in place for a repeating callback to run, controlled by a boolean preference. Obviously later this will be calling real code to do real work.

### DIFF
--- a/TODO.py
+++ b/TODO.py
@@ -1,0 +1,138 @@
+# https://tkdocs.com/tutorial/text.html
+#
+# https://stackoverflow.com/a/3781773
+# The basic concept is, you assign properties to tags, and you apply tags to
+# ranges of text in the widget. You can use the text widget's search command to
+# find strings that match your pattern, which will return you enough information
+# apply a tag to the range that matched.
+#
+# https://stackoverflow.com/a/3781773
+# This answer seems to have a way to do a search/match with a regex? Also Nigel
+# probably has done a lot of work here that can be used as an example of even
+# re-used.
+#
+# Relevant tickets (all from GG1 search menu):
+#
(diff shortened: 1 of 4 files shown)


tangledhelix:
I figured this out in the interim. I needed to separate the concerns, having one function with the sole job of setting the preference to a new value, separate from the callback function, whose job is to decide whether or not to call the function that starts the loop going, when the preference is changed. Previously, the combined function was being called twice (once by the callback handler, and once by the event handler when the menu item is selected). This resulted in two nearly-concurrent loops beginning.

PR #262, opened 2024-05-29
Remove PageSepDialog's `undo_block_begin`

windymilla:
I put a method in there temporarily until I had the full version completed and merged. Replaced with the
full version from `maintext`


windymilla:
Testing notes - should be no change in behavior. If anything is broken, it will be "undo"ing Page Separator Fixup operations.

srjfoo (approved):

PR #263, opened 2024-05-31
Tidy/harmonize menu entries

windymilla:
Some menu entries did not have `...` following them. It appears the most common use of this is when a new dialog will appear that the user can interact with to complete the task, and not for just "View xyz" or
"Help About" entries.

Also, put the 4 unmatched markup/bracket/quote
entries into a submenu, to stop the Tools menu becoming too long.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -373,7 +373,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             self.file.add_good_and_bad_words,
         )
         proj_menu.add_button(
-            f"Set ~Scan Image {'Folder' if is_windows() else 'Directory'}",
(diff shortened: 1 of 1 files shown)


srjfoo:
> Also, put the 4 unmatched markup/bracket/quote
> entries into a submenu, to stop the Tools menu becoming too long.

Just a suggestion -- you could do this to the case change ops in the Edit menu, perhaps call it "Change case"?

srjfoo (approved):

windymilla:
> Just a suggestion -- you could do this to the case change ops in the Edit menu, perhaps call it "Change case"?

Good idea - I'll do that. If we change our mind later about menu layout, it's easy to adjust.

PR #265, opened 2024-06-01
Add first part of Footnote Fixup

windymilla:
Since footnote fixup is a large task, split it into pieces.

This piece just lists all the footnotes and anchors in a Checker Dialog.

Algorithm is to find occurrences of "[footnote", allowing capitalization & space variation. For each one, get its label, e.g. "4" in "[Footnote 4:", then search backwards for an anchor that matches it. Don't use "/#[4]" though, because that defines width parameter for block markup.

Each anchor or footnote is added to the checker dialog, and the whole footnote should be highlighted when the line in the dialog is selected. Note that one level of nested square brackets is permitted inside a footnote, i.e. a footnote starts & ends with square brackets, and may contain open & matching close square brackets, but not brackets within those brackets.

Sort by alpha/type puts all anchors first, then all  footnotes.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -71,6 +71,11 @@ Word Frequency
 ==============
     
 .. automodule:: word_frequency
+ 
+Footnotes
+==============
+    
+.. automodule:: footnotes
 
 PPtxt
 =====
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: 
I used projectID6434788b92eee since it has footnotes, but any file should be OK - you can edit individual footnote to test our scenarios. Tools->Fixup Footnotes...
At the moment, there are no actual checks, like reporting if "Footnote" is badly capitalized, or that there is no anchor for a footnote, or an anchor that appears to link to two footnotes, etc.
Just a general sanity check would be great, thanks.

windymilla:
Rebased on latest master

srjfoo (approved):

PR #273, opened 2024-06-04
Fix cut/copy to BBedit bug

windymilla:
Where clipboard is added to by Tcl/Tk, schedule a call to a routine that sets the clipboard to its own contents, but using the `pyperclip` module instead of Tcl/Tk.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2,6 +2,7 @@
 
 
 import logging
+import subprocess
 from textwrap import TextWrapper
 import tkinter as tk
 from tkinter import ttk
@@ -706,8 +707,25 @@ class MainText(tk.Text):
         rowcol = self.rowcol(f"{start_rowcol.index()} + {len(clipline)}c")
         self.set_insert_index(rowcol)
 
+    def affirm_clipboard_contents(self) -> None:
+        """Ensure clipboard is set to its "current" contents.
+
+        The purpose of this is to set the clipboard by non-Tcl/Tk means.
+        This should bypass the bug where Tcl/Tk doesn't update the
+        system clipboard counter. Some apps, e.g. BBEdit, need this to detect the
+        clipboard has changed: https://github.com/python/cpython/issues/104613
+        """
+        if not is_mac():
+            raise NotImplementedError("This function only works on macOS")
+
+        # Use pbcopy macOS command to "touch" the clipboard contents
+        with subprocess.Popen(["/usr/bin/pbcopy"], stdin=subprocess.PIPE) as proc:
+            proc.communicate(input=self.clipboard_get().encode())
+
     def smart_copy(self) -> str:
         """Do column copy if multiple ranges selected, else default copy."""
+        if is_mac():
+            self.after_idle(self.affirm_clipboard_contents)
         if len(self.selected_ranges()) <= 1:
             return ""  # Permit default behavior to happen
         self.column_copy_cut()
@@ -715,6 +733,8 @@ class MainText(tk.Text):
 
     def smart_cut(self) -> str:
         """Do column cut if multiple ranges selected, else default cut."""
+        if is_mac():
+            self.after_idle(self.affirm_clipboard_contents)
         if len(self.selected_ranges()) <= 1:
             return ""  # Permit default behavior to happen
         self.column_copy_cut(cut=True)


cpeel (commented):
I'd feel better about using pyperclip if it was more actively maintained. And while the use of it is pretty minimal it might be more than we need since it's only for Mac.

Something like this (which is what pyperclip is doing) might be simpler:
```python
#!/usr/bin/env python

import subprocess

def affirm_clipboard_contents() -> None:
    if not is_mac():
        raise NotImplementedError("This function only works on macOS")

    # use pbpaste / pbcopy macOS commands to "touch" the clipboard contents
    paste = subprocess.Popen(['pbpaste'], stdout=subprocess.PIPE)
    subprocess.check_output(['pbcopy'], stdin=paste.stdout)
    paste.wait()
```

tangledhelix:
I'd prefer to see fully-qualified paths to the binaries, or else we should override the path from the calling environment. By default, tkinter will use whatever the calling environment used for `PATH`, falling back to `os.defpath`. An example from my system after adding a bit of debug code to GG:

```
path[]=/Users/dan/Library/Caches/pypoetry/virtualenvs/guiguts-6B18dp7I-py3.11/bin
path[]=/Users/dan/.pyenv/shims
path[]=/opt/homebrew/lib/ruby/gems/3.2.0/bin
path[]=/opt/homebrew/Cellar/ruby/3.2.2_1/bin
path[]=/Users/dan/perl5/perlbrew/bin
path[]=/Users/dan/perl5/perlbrew/perls/perl-5.35.6/bin
path[]=/Users/dan/bin
path[]=/opt/homebrew/bin
path[]=/opt/homebrew/sbin
path[]=/usr/local/bin
path[]=/usr/bin            <--- only here do we get to where these binaries live
path[]=/usr/sbin
path[]=/bin
path[]=/sbin
path[]=/opt/X11/bin
path[]=/usr/X11R6/bin
path[]=/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin
path[]=/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin
path[]=/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
path[]=/Library/Apple/usr/bin
path[]=/opt/X11/bin
path[]=/Applications/Wireshark.app/Contents/MacOS

which pbcopy: /usr/bin/pbcopy
which pbpaste: /usr/bin/pbpaste
```

It just so happens that there are no `pbcopy` or `pbpaste` earlier in my `PATH`, but one must program defensively when forking child programs in 2024... I might be an outlier, but we can't assume what we'll find on a running system.

srjfoo:
This is what I got testing the latest iteration. What the screenshot doesn't show is that the cursor is the dread spinning beachball. Eventually (after a couple of minutes that seemed longer), Python responded and the beachball went away, but there was nothing on the clipboard to paste into BBEdit. (😁 -- yes, that testing text is from an earlier branch)

image

windymilla:
Tried an alternative to see if it avoids an interprocess communication delay (if that's the problem).

windymilla:
Or the timeout may have been caused by my lint-silencing change which was to use 
`with subprocess.Popen(["/usr/bin/pbpaste"], stdout=subprocess.PIPE) as paste:`
@srjfoo & I can investigate some more

windymilla:
@cpeel & @tangledhelix - if you don't mind, could you cast your eye over the latest version to see if you're happy with it. It uses the idea Casey extracted from the`pyperclip` code, but only the `pbcopy` call, since we can get the clipboard contents directly from Tk without any problems, so no need to use `pbpaste`. It also uses the full path to `pbcopy` as Dan suggested.

I switched to using `with` to stop lint complaining. Also, apparently `subprocess.communicate()` closes stdin once it has sent the data, which should cause `pbcopy` to exit promptly. 

Sharon has confirmed:

1. It works when copying text and pasting in BBEdit
2. It encodes Unicode strings OK
3. There's no "beachball"
4. There's no noticeable delay.

Thanks all!

cpeel (approved):
This change LGTM!

I suggest dropping the `poetry.lock` change from the PR and squashing all of the commits together so there isn't a add-then-remove a package as part of the git history for future code splunkers.

tangledhelix:
I don't have BBedit, so I'll leave that for Sharon, but I can paste/copy back and forth between TextEdit and GG2, including Unicode (I used some Greek text from a project) and even Emoji, it all works perfectly fine. No delays observed.

srjfoo (approved):

srjfoo:
In a pinch, it would have been okay to just document the problem, but I'm glad it was fixable. We don't know where else the issue might have come back to bite us in the backside.

windymilla:
Thanks all!

PR #277, opened 2024-06-08
Add error checking to Footnote Fixup

### LINKED ISSUE #275: Put some error checking into Footnote Fixup

windymilla:
1. If a footnote has no anchor
2. If two footnotes have the same anchor
2. If the anchors/footnotes are out of sequence with each other. Note that the actual numbers don't matter, e.g. if anchor 2 appears before anchor 1, it's OK as long as footnote 2 appears before footnote 1, but not if footnote 2 appears after footnote 1.

Fixes #275


srjfoo (approved):
Works as advertised, as far as I can tell!

I don't know if this would be useful in general, or just something that would be useful in testing: since I was seeding errors, and each error I seeded actually caused two or more lines to be marked to be looked at, it occurred to me that it would be nice to have a count of the number of marked lines in the header (say, in parentheses following the count of the number of entries).

PR #280, opened 2024-06-12
Add compose sequence dialog

windymilla:
Shortcut is `Ctrl/Cmd+semicolon` or use the Tools--> Unicode menu.
The only sequences defined so far are:
1. Up to 4 hex digits - dialog closes when 4 are typed.
2. Hex digits can optionally be preceded by `x`, `U+`, etc.
3. Or use `#nnn` where `nnn` is any number of decimal digits. Use Apply/OK to terminate these.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -25,7 +25,7 @@ from guiguts.mainwindow import (
     ErrorHandler,
     process_accel,
 )
-from guiguts.misc_dialogs import PreferencesDialog
+from guiguts.misc_dialogs import PreferencesDialog, ComposeSequenceDialog
 from guiguts.misc_tools import (
     basic_fixup_check,
     page_separator_fixup,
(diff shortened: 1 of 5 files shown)


windymilla:
Testing notes for @srjfoo - description of latest commit:

> Retain `xxxx` (where `x` is a hex digit) to be accepted immediately without clicking OK/Return. This covers characters in the Unicode BMP.
> 
> User can enter an ordinal with fewer digits by clicking OK/Return, e.g.  e.g. `b1<Return>` for `±`.
> 
> By using a prefix (U, U+, 0x, \x, x) user can type more 4 digits, then press OK/Return, e.g. `U+23456<Return>` for `𣑖` .
> 
> Prefix `#` continues to be interpreted as decimal.
> 
> This seems to be the best compromise: the most common characters are in the range 0000-ffff, and can be typed without a prefix or using OK/Return. Less common characters (above ffff) are available by using a prefix.

srjfoo (approved):

PR #285, opened 2024-06-14
Add "Suspects Only" toggle to checker dialogs

windymilla:
Only currently needed for Footnote Fixup, but
implementation is general, so can be used for future checkers.


windymilla:
Testing notes: 

- Run footnote fixup. 
- In checker dialog, toggle "Suspects Only" on/off.
- If you have a suspect selected, it should remain selected when you toggle "Suspects Only".
- If you have a non-suspect selected, then turn on "Suspects Only", the first suspect should be selected.

srjfoo (approved):

PR #291, opened 2024-06-17
Disable WF case buttons if `Case Ignore`

### LINKED ISSUE #289: Word Frequency: Initial Caps & Mixed Case don't find anything if Ignore Case is set
It kind of makes sense, because Ignore Case treats all the words as lower case, but could cause confusion to the user.

windymilla:
When Case Ignore is on, words that are ALL CAPS, Initial Caps, or MiXeD cASe aren't detected, which could
confuse the user. Therefore disable those buttons when Case Ignore is on, and enable them when it's off. Display a tooltip to explain to the user if they are disabled.

Fixes #289

### DIFF
--- a/src/guiguts/widgets.py
+++ b/src/guiguts/widgets.py
@@ -428,8 +428,8 @@ class ToolTip(tk.Toplevel):
         super().destroy()
 
 
-def unbind_from(widget: tk.Widget, seq: str, funcid: str) -> None:
-    """Unbind function associated with `funcid`.
+def unbind_from(widget: tk.Widget, sequence: str, func_id: str) -> None:
+    """Unbind function associated with `func_id`.
 
     Based on: https://github.com/python/cpython/issues/75666#issuecomment-1877547466
     Necessary due to long-standing bug in tkinter: when `unbind` is called
@@ -438,15 +438,17 @@ def unbind_from(widget: tk.Widget, seq: str, funcid: str) -> None:
 
     Args:
         widget: Widget to unbind from.
-        seq: Sequence function was bound to.
-        funcid: Function ID returned by original `bind` call.
+        sequence: Sequence function was bound to.
+        func_id: Function ID returned by original `bind` call.
     """
     # Make dictionary of all bindings to seq
-    bindings = {x.split()[1][3:]: x for x in widget.bind(seq).splitlines() if x.strip()}
+    bindings = {
+        x.split()[1][3:]: x for x in widget.bind(sequence).splitlines() if x.strip()
+    }
     # Delete the binding associated with `funcid`
-    del bindings[funcid]
+    del bindings[func_id]
     # Re-bind all the other bindings to the widget
-    widget.bind(seq, "\n".join(list(bindings.values())))
+    widget.bind(sequence, "\n".join(list(bindings.values())))
 
 
 def grab_focus(
--- a/src/guiguts/widgets.py
+++ b/src/guiguts/widgets.py
@@ -422,10 +422,31 @@ class ToolTip(tk.Toplevel):
     def destroy(self) -> None:
         """Destroy the ToolTip and unbind all the bindings."""
         if self.widget.winfo_exists():
-            self.widget.unbind("<Enter>", self.enter_bind)
-            self.widget.unbind("<Leave>", self.leave_bind)
-            self.widget.unbind("<ButtonPress>", self.press_bind)
+            unbind_from(self.widget, "<Enter>", self.enter_bind)
+            unbind_from(self.widget, "<Leave>", self.leave_bind)
+            unbind_from(self.widget, "<ButtonPress>", self.press_bind)
         super().destroy()
+
+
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

tangledhelix:
✅ When setting "ignore case" the buttons are properly disabled, their labels are grayed out, and the tooltips are working.

✅ If you then uncheck "ignore case" the text labels on the buttons go back to normal, and the tooltips disappear.

❌ But, the buttons themselves are still disabled - for all 3 buttons I cannot use that function anymore, unless I close WF and open a new WF dialog.

srjfoo:
> ❌ But, the buttons themselves are still disabled - for all 3 buttons I cannot use that function anymore, unless I close WF and open a new WF dialog.

Confirmed. I didn't think to check this in my initial testing. In this round of testing, I tried manually re-running it, but as @tangledhelix says, only closing WF and re-starting it worked.

windymilla:
> ❌ But, the buttons themselves are still disabled - for all 3 buttons I cannot use that function anymore, unless I close WF and open a new WF dialog.

Also confirmed (and embarrassed not to have noticed while testing)

windymilla:
I'm very hesitant to say it's a "python/tkinter/tk/tcl/etc" bug, but in this case it is: https://github.com/python/cpython/issues/75666

So, slightly less embarrassment, and hopefully now fixed. Thanks @tangledhelix for spotting this.

srjfoo:
Is it supposed to default to All Words if you select "Ignore Case" while one of the case-sensitive options is selected? Right now, everything is deselected for me (macOS). Which I don't have a problem with, but just wanted to make sure that's the expected behavior.

windymilla:
For me, it still shows ALL CAPS selected, even though it is disabled, and the list of words is empty. If you hover over it, it displays the tooltip, and if you turn Ignore Case off again, you get back the ALL CAPS words. I'm happy with that behaviour - is that different to what you see? I

tangledhelix:
> For me, it still shows ALL CAPS selected, even though it is disabled, and the list of words is empty. If you hover over it, it displays the tooltip, and if you turn Ignore Case off again, you get back the ALL CAPS words. I'm happy with that behaviour - is that different to what you see?

What you describe is what I see also. You have to look a bit carefully maybe to notice - the disabled widget makes it hard-ish to notice (at least on Mac, at least in Light Mode; I didn't try in Dark Mode).

Screenshot 2024-06-21 at 7 39 49 PM

I think the current behavior is acceptable.

My one gripe is that when checking / un-checking `Ignore Case`, it's not always obvious that the re-run is currently running and the window freezes up for a couple of seconds (at least on my machine, with my test file, it was a couple of seconds). Auto-triggering a re-run is the correct behavior, but I wonder if there is some non-complex way to indicate the re-run is occurring. I'm not sure what tkinter might have in that regard. It could be as simple as a status label somewhere that switches between two states, e.g. `Running ...` and `Ready.` or similar?

An animated would be even better but I don't know if there's a widget like that, or if something easy like an animated GIF would render in this context.

animated-gif

tangledhelix:
... I did not expect that spinner graphic to be so large! 🤨

tangledhelix (approved):

srjfoo:
> > For me, it still shows ALL CAPS selected, even though it is disabled, and the list of words is empty. If you hover over it, it displays the tooltip, and if you turn Ignore Case off again, you get back the ALL CAPS words. I'm happy with that behaviour - is that different to what you see?
> 
> What you describe is what I see also. You have to look a bit carefully maybe to notice - the disabled widget makes it hard-ish to notice (at least on Mac, at least in Light Mode; I didn't try in Dark Mode).
> 
> I think the current behavior is acceptable.

That's what I'm seeing now, too. I'm not sure what caused the button last selected not to be selected any longer the time I tried it earlier, but that's not what I'm seeing now.
> 
> My one gripe is that when checking / un-checking `Ignore Case`, it's not always obvious that the re-run is currently running and the window freezes up for a couple of seconds (at least on my machine, with my test file, it was a couple of seconds). Auto-triggering a re-run is the correct behavior, but I wonder if there is some non-complex way to indicate the re-run is occurring. I'm not sure what tkinter might have in that regard. It could be as simple as a status label somewhere that switches between two states, e.g. `Running ...` and `Ready.` or similar?
> 
> An animated graphic (spinner, bar) would be even better but I don't know if there's a widget like that, or if something easy like an animated GIF would render in this context.
> 
Maybe not as obvious, but would it be possible to force the re-run button to change color as it does when you click on it? It's not quite as obvious on macOS with the default theme when the system is on dark mode as it is on light mode, but there is a difference. Or put an additional drop-shadow on the button while the function is working? (No, please not a spinner of that size! 🤣)

srjfoo (approved):

windymilla:
I wonder if it would be better to implement a general "busy" state, like GG1 has, which changes the cursor to the busy cursor (whatever that is for your system) and then changes it back when it's done. That seems to be a fairly standard way to indicate that something is happening.

PR #296, opened 2024-06-21
Highlight single/double quotes in a selection

### LINKED ISSUE #260: Highlight (single, double) quotes in selection
In GG1 Search Menu

@windymilla if you can assign to me, I'm looking into this alongside #41, #42, and #43.

tangledhelix:
Add ability to highlight single or double quotes in the current selection, and to remove current highlights. Based on GG1's behavior. All 3 are available as menu items in the Search menu.

Testing notes:
- Select some text, then choose each item from the menu
- Modes are exclusive; highlight single, then double, will un-highlight single before highlighting double
- Remove highlights should remove any current highlights
- Both straight and curly quotes are supported
- Column selection is supported
- If you select text, highlight, then select different text (even if they overlap) and highlight again, old highlights are removed first, then new highlights in the selected area applied.
- Choosing the menu item with no selection in effect, does nothing

All of these behaviors are based on GG1's behavior with these exceptions:
- GG1 doesn't support these functions in column select; so that is a new behavior
- The highlight color is configurable in GG1; I have them static here for now. This probably needs to be thought through, i.e. can this be integrated into the theme system?

Fixes #260

### DIFF
--- a/src/guiguts/highlight.py
+++ b/src/guiguts/highlight.py
@@ -1,0 +1,102 @@
+"""Highlight functionality"""
+
+from enum import auto
+
+from guiguts.maintext import maintext
+from guiguts.preferences import preferences, PrefKey
+
+
+class Highlight:
+    """Global highlight settings"""
+
+    TAG_QUOTEMARK = auto()
+
+    # Possible future enhancement:
+    #
+    # GG1 allowed you to set three colors:
+    # - Background color (text areas, text inputs, main editor textarea)
+    # - Button highlight color (hover on buttons, checkboxes, radio selects)
+    # - Scanno/quote highlight color (which would apply here)
+    #
+    # Unclear what we should/will do in GG2 with themes & dark mode support.
+
+    # Must be a definition for each available theme
+    COLORS_QUOTEMARK = {
+        "Light": {"bg": "#a08dfc", "fg": "black"},
+        "Dark": {"bg": "#a08dfc", "fg": "white"},
+        "Default": {"bg": "#a08dfc", "fg": "black"},
+    }
+
+
+def highlight_selection(
+    pat: str,
+    tag_name: str,
+    nocase: bool = False,
+    regexp: bool = False,
+    wholeword: bool = False,
+) -> None:
+    """Highlight matches in the current selection.
+    Args:
+        pat: string or regexp to find in the current selection
+        tag_name: tkinter tag to apply to matched text region(s)
+    Optional keyword args:
+        nocase (default False): set True for case-insensitivity
+        regexp (default False): whether to assume 's' is a regexp
+        wholeword (defalut False): whether to match only whole words
+    """
+
+    if not (ranges := maintext().selected_ranges()):
+        return
+
+    for _range in ranges:
+        matches = maintext().find_matches(
+            pat, _range, nocase=nocase, regexp=regexp, wholeword=wholeword
+        )
+        for match in matches:
+            maintext().tag_add(
+                tag_name, match.rowcol.index(), match.rowcol.index() + "+1c"
+            )
+
+
+def get_active_theme() -> str:
+    """Return the current theme name"""
+    return preferences.get(PrefKey.THEME_NAME)
+
+
+def remove_highlights() -> None:
+    """Remove acvite highlights"""
+    maintext().tag_delete(str(Highlight.TAG_QUOTEMARK))
+
+
+def highlight_quotemarks(pat: str) -> None:
+    """Highlight quote marks in current selection which match a pattern"""
+    theme = get_active_theme()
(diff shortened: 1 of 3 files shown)


src/guiguts/highlight.py
```diff
@@ -0,0 +1,90 @@
...
+    # - Button highlight color (hover on buttons, checkboxes, radio selects)
+    # - Scanno/quote highlight color (which would apply here)
+    #
+    # Unclear what we should/will do in GG2 with themes & dark mode support.
+
+    # Must be a definition for each available theme
+    COLORS_QUOTEMARK = {
+        "Light": {"bg": "#a08dfc", "fg": "black"},
+        "Dark": {"bg": "#a08dfc", "fg": "white"},
+        "Default": {"bg": "#a08dfc", "fg": "black"},
```
srjfoo, line 27: If "fg" is supposed to be white on the `Dark`, it's not working for me (macOS Sonoma). All three of them, I see the same fg and bg colors. I.e. black text on `#a08dfc`.

src/guiguts/highlight.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def highlight_single_quotes() -> None:
+    """Highlight single quotes in current selection"""
+    highlight_quotemarks("['‘’]")
+
+
+def highlight_double_quotes() -> None:
+    """Highlight double quotes in current selection"""
+    highlight_quotemarks('["“”]')
```
srjfoo, line 90: Should this include guillemets? I'd suggest the high and low quote marks, also, except that those are currently automatically converted to normal straight single and double on project load, and won't be accepted in the proofing rounds, so would only be useful for verifying after conversion.

tangledhelix: I did wonder about guillemets, and I am happy to add them, but they were not included in GG1, so I didn't include them here either.

srjfoo: I thought it might be something like that, but figured that the question was still worth asking :)

src/guiguts/highlight.py
srjfoo: Nevermind (🤦‍♀️). I forgot until I was looking back at it that you have to restart to guarantee that the new theme takes full effect. It's working as advertised.

tangledhelix: I think the built-in themes in tkinter natively support dark/light mode but I don't see any code in GG that knows the difference when the theme is `Default`.

I found a library that claims to be able to do this detection on Mac, Windows, and Linux, but I haven't tried it.

https://pypi.org/project/darkdetect/

srjfoo: > Here `Dark` means the Guiguts theme "Dark" in the preferences - which is not the same thing as using Dark mode in macOS, with `Default` theme in Guiguts. I'll try to figure out how to detect the system dark/light mode when in Default theme.

Yes, I realized that -- and that was what I was seeing, but as my later post mentions, I forgot that for certain things, to guarantee that you're seeing it correctly, you have to restart GG2.

tangledhelix: Detecting Dark mode from the shell is "easy" although... a bit stupid.

```
❯ defaults read -g AppleInterfaceStyle;echo $?
Dark
0

❯ defaults read -g AppleInterfaceStyle;echo $?
2024-06-21 00:41:52.661 defaults[57537:10085832]
The domain/default pair of (kCFPreferencesAnyApplication, AppleInterfaceStyle) does not exist
1
```

In other words, it doesn't tell you the mode if it's light, you get an error instead. You wouldn't even need to read the output, I guess; the return value of the program tells the story.

I don't propose using this, by the way - although it seems to work, if nothing better emerges...

src/guiguts/highlight.py
tangledhelix: Added more matched characters.

## single quotes
- `‹›` SINGLE {LEFT, RIGHT}-POINTING ANGLE QUOTATION MARK
- `‛‚` SINGLE {HIGH-REVERSED-9, LOW-9} QUOTATION MARK

## double quotes
- `«»` {LEFT, RIGHT}-POINTING DOUBLE ANGLE QUOTATION MARK
- `‟„` DOUBLE {HIGH-REVERSED-9, LOW-9} QUOTATION MARK

windymilla (changes requested):
Please could you add a section to `index.rst` so that sphinx generates the docs for `highlight.py`.
Apart from that, the code and behavior look great - thanks!

srjfoo (approved):

tangledhelix:
> Please could you add a section to `index.rst` so that sphinx generates the docs for `highlight.py`. Apart from that, the code and behavior look great - thanks!

Sure thing - that's done now.

windymilla (approved):

PR #298, opened 2024-06-21
Add remaining compose sequences

windymilla:
This adds fractions, various other characters, quotes, super/subscript numbers, and Greek characters.

Also adds a Help-->Compose Sequences dialog, which lists all valid compose sequences, and allows the user to click on a letter in that help to insert it directly.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -570,7 +570,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         unicode_menu.add_button(
             "~Compose Sequence...",
             ComposeSequenceDialog.show_dialog,
-            "Cmd/Ctrl+M",
+            "Cmd/Ctrl+I",
         )
 
     def init_view_menu(self) -> None:
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -31,7 +31,11 @@ from guiguts.mainwindow import (
     ErrorHandler,
     process_accel,
 )
-from guiguts.misc_dialogs import PreferencesDialog, ComposeSequenceDialog
+from guiguts.misc_dialogs import (
+    PreferencesDialog,
+    ComposeSequenceDialog,
+    ComposeHelpDialog,
+)
 from guiguts.misc_tools import (
     basic_fixup_check,
     page_separator_fixup,
@@ -586,6 +590,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_help = Menu(menubar(), "~Help")
         menu_help.add_button("Guiguts ~Manual", self.show_help_manual)
         menu_help.add_button("About ~Guiguts", self.help_about)
+        menu_help.add_button("~Compose Sequences", ComposeHelpDialog.show_dialog)
 
     def init_os_menu(self) -> None:
         """Create the OS-specific menu.
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -569,7 +569,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         unicode_menu.add_button(
             "~Compose Sequence...",
             ComposeSequenceDialog.show_dialog,
-            "Cmd/Ctrl+;",
+            "Cmd/Ctrl+semicolon",
         )
 
     def init_view_menu(self) -> None:
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -336,6 +336,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             ),
         )
         preferences.set_default(PrefKey.TEAROFF_MENUS, False)
+        preferences.set_default(PrefKey.COMPOSE_HISTORY, [])
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -542,7 +543,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
         menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
         menu_tools.add_separator()
-        unmatched_menu = Menu(menu_tools, "~Unmatched")
+        unmatched_menu = Menu(menu_tools, "Un~matched")
         unmatched_menu.add_button("Bloc~k Markup...", unmatched_block_markup)
         unmatched_menu.add_button("~DP Markup...", unmatched_dp_markup)
         unmatched_menu.add_button("~Brackets...", unmatched_brackets)
@@ -569,7 +570,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         unicode_menu.add_button(
             "~Compose Sequence...",
             ComposeSequenceDialog.show_dialog,
-            "Cmd/Ctrl+semicolon",
+            "Cmd/Ctrl+M",
         )
 
     def init_view_menu(self) -> None:
@@ -590,7 +591,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_help = Menu(menubar(), "~Help")
         menu_help.add_button("Guiguts ~Manual", self.show_help_manual)
         menu_help.add_button("About ~Guiguts", self.help_about)
-        menu_help.add_button("~Compose Sequences", ComposeHelpDialog.show_dialog)
+        menu_help.add_button(
+            "List of ~Compose Sequences", ComposeHelpDialog.show_dialog
+        )
 
     def init_os_menu(self) -> None:
         """Create the OS-specific menu.
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -294,7 +294,7 @@ def init_compose_dict() -> None:
     init_chars("⅞", "7/8")
     init_chars("⅑", "1/9")
     init_chars("⅒", "1/10")
-    for num, char in enumerate("⁰ⁱ²³⁴⁵⁶⁷⁸⁹"):
+    for num, char in enumerate("⁰¹²³⁴⁵⁶⁷⁸⁹"):
         init_chars(char, f"^{num}")
     for num, char in enumerate("₀₁₂₃₄₅₆₇₈₉"):
         init_chars(char, f",{num}")
@@ -515,7 +515,7 @@ class ComposeHelpDialog(ToplevelDialog):
 
     def __init__(self) -> None:
         """Initialize class members from page details."""
-        super().__init__("Compose Sequences")
+        super().__init__("List of Compose Sequences")
 
         self.column_headings = ("Character", "Sequence", "Name")
         widths = (70, 70, 600)
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -236,6 +236,46 @@ def init_compose_dict() -> None:
     """Initialize dictionary of compose sequences."""
     if _compose_dict:
         return  # Already initialized
+    init_char_reversible("‘", "'<")
+    init_char_reversible("’", "'>")
+    init_char_reversible("“", '"<')
+    init_char_reversible("”", '">')
+    init_chars("±", "+-")
+    init_chars("·", "^.", "*.", ".*")
+    init_chars("×", "*x", "x*")
+    init_chars("÷", ":-")
+    init_char_reversible("°", "oo", "*o")
+    init_char_reversible("′", "1'")
+    init_char_reversible("″", "2'")
+    init_char_reversible("‴", "3'")
+    init_char_reversible(" ", "  ", "* ")
+    init_chars("—", "--")
+    init_char_reversible("–", "- ")
+    init_chars("⁂", "**")
+    init_char_reversible("º", "o_")
+    init_char_reversible("ª", "a_")
+    init_chars("‖", "||")
+    init_chars("¡", "!!")
+    init_chars("¿", "??")
+    init_chars("«", "<<")
+    init_chars("»", ">>")
+    init_char_case("Æ", "AE")
+    init_char_case("Œ", "OE")
+    init_char_case("ẞ", "SS")
+    init_char_case("Ð", "DH", "ETH")
+    init_char_case("Þ", "TH")
+    init_chars("©", "(c)", "(C)")
+    init_chars("†", "dag", "DAG")
+    init_chars("‡", "ddag", "DDAG")
+    init_accent("£", "L", "-")
+    init_accent("¢", "C", "/", "|")
+    init_chars("§", "sec", "s*", "*s", "SEC", "S*", "*S")
+    init_chars("¶", "pil", "p*", "*p", "PIL", "P*", "*P")
+    init_chars("ſ", "sf", "SF")
+    init_chars("‚", ",'")
+    init_chars("‛", "^'")
+    init_chars("„", ',"')
+    init_chars("‟", '^"')
     init_chars("½", "1/2")
     init_chars("⅓", "1/3")
     init_chars("⅔", "2/3")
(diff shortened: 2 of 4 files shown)


windymilla:
Testing notes:
If you pop up Help-->Compose Sequences, you can see the sequences that are now available. New feature - you can also click on an entry in the help dialog and insert it directly, rather than find it in the help, then go to the compose sequence dialog to insert it.

srjfoo:
Maybe should be fixed?

image

This struck me as really odd, so I went looking in the unicode chart. Superscripts 1, 2 and 3 are at U+00b2, 00b3 and 00b9, unlike the superscripts you have listed here.

Scrolling down the help list, the longer unicode names are truncated. For instance, `GREEK CAPITAL LETTER EPSILON WITH DASIA AND` (with the trailing D truncated) to take extreme examples. Widening the window does not show the rest of the names, so I'm guessing they're truncated when displayed?

`<tb>`

Random thoughts (from someone who hasn't used GG1 enough to become used to some of the weird quirks 😁):

* I find the help list very visually cluttered when there are four entries in a row for typing `LATIN SMALL LETTER E WITH ACUTE` (and similar). And eight different combinations for a pound or a cent sign. (This is not something I have a solution for; just thought I'd mention it.)

* If this is a help function, shouldn't it have instructions on how to use the compose feature in other ways? Or a tooltip on the insertion dialog, itself? I'm sure it's documented in the manual for GG1, but in the spirit of having useful information available in the application ... (I don't know if this is something you have in mind to tweak in a later PR, but wanted to go ahead and mention it.)

windymilla:
> This struck me as really odd, so I went looking in the unicode chart. Superscripts 1, 2 and 3 are at U+00b2, 00b3 and 00b9, unlike the superscripts you have listed here.

The help dialog is in order of Unicode ordinals - if you look a bit higher up you'll see the 1, 2 & 3. It seemed sensible to follow the Unicode order (like GG1). If you were looking for how to enter a particular character, how would you know where to look in the help list. 

> Scrolling down the help list, the longer unicode names are truncated.

I'll make the list wider, or see if it's possible to not give a width and let it grow to the size it needs to. The width limit is in the widget - I'm not truncating them deliberately.

> * I find the help list very visually cluttered when there are four entries in a row...

Yes - I don't know what the right balance is between telling the user about the options for typing a character and over-cluttering. One thing I could do is to remove some of the options (e.g. you can use `'` or `/` for acute accent, and two options for grave accent. I could simply remove one of those to just leave `/` and `\`, and possibly similar elsewhere. There are a few other things I could do to reduce the number of options, e.g. "dag" and "DAG" both insert a dagger, but it would probably be sufficient to just mention "dag". It would complicate the code quite a bit compared to currently where it just displays all the possibilities.

> * If this is a help function, shouldn't it have instructions on how to use the compose feature in other ways?

No, I don't think it should. It's a list of the compose sequences, not a dialog giving help on how to use the feature. I don't plan to add dialogs like that. I'm not sure what a tooltip would say in the compose sequence dialog apart from "Type a compose sequence here" but that seems pretty obvious if you've popped the Compose Sequence dialog. I'm not sure the fields in the dialogs all need tooltips unless their usage is not the obvious "type the thing here." I could change the label to say "Compose Sequence" instead of "Compose". I kept it short so the dialog was small, but it could go a bit wider if you think that would make it clearer.

windymilla:
I've attempted to address the above queries in the latest commit:
1. Display in order of creation (i.e. can be any order) rather than Unicode order.
2. Reduce clutter by hiding duplicates that only vary by case for a non-cased character, e.g. dag/DAG
3. Reduce clutter by putting reversibles on the same line, e.g. `.* or *.` for middle dot instead of separate entries.
4. Final column made wider to fit all character names.
5. Fixed exception when user clicked in header.

windymilla:
Rebased - should have everything in now.

srjfoo:
> > This struck me as really odd, so I went looking in the unicode chart. Superscripts 1, 2 and 3 are at U+00b2, 00b3 and 00b9, unlike the superscripts you have listed here.
> 
> The help dialog is in order of Unicode ordinals - if you look a bit higher up you'll see the 1, 2 & 3. It seemed sensible to follow the Unicode order (like GG1). If you were looking for how to enter a particular character, how would you know where to look in the help list.

That's fine. My fault for noting the later group, but not the earlir one. 🙄 

> > Scrolling down the help list, the longer unicode names are truncated.
> 
> I'll make the list wider, or see if it's possible to not give a width and let it grow to the size it needs to. The width limit is in the widget - I'm not truncating them deliberately.

Thanks, looks good! It wouldn't have been that big an issue if the text width had either scrolled the overflow, or grown with the width of the widget.

> > * I find the help list very visually cluttered when there are four entries in a row...
> 
> Yes - I don't know what the right balance is between telling the user about the options for typing a character and over-cluttering. One thing I could do is to remove some of the options (e.g. you can use `'` or `/` for acute accent, and two options for grave accent. I could simply remove one of those to just leave `/` and `\`, and possibly similar elsewhere. There are a few other things I could do to reduce the number of options, e.g. "dag" and "DAG" both insert a dagger, but it would probably be sufficient to just mention "dag". It would complicate the code quite a bit compared to currently where it just displays all the possibilities.

Thanks -- the newer version is still somewhat cluttered, but much better now.

> > * If this is a help function, shouldn't it have instructions on how to use the compose feature in other ways?
> 
> No, I don't think it should. It's a list of the compose sequences, not a dialog giving help on how to use the feature. I don't plan to add dialogs like that. I'm not sure what a tooltip would say in the compose sequence dialog apart from "Type a compose sequence here" but that seems pretty obvious if you've popped the Compose Sequence dialog. I'm not sure the fields in the dialogs all need tooltips unless their usage is not the obvious "type the thing here." I could change the label to say "Compose Sequence" instead of "Compose". I kept it short so the dialog was small, but it could go a bit wider if you think that would make it clearer.

That's fair. I did notice that with the name change to "Compose Sequence" (which I do like, and I think that the widget can be narrower than you have it if you want it to be), there's a potential confusion with the other widget being named "Compose Sequences".

And speaking of "Compose Sequences", I assume it's in the "Help" menu in GG2 because that's where it is in GG1. But the nature of it has changed. In GG1, it's just a reference document. In GG2, it's actually more useful for just clicking to add the character, rather than as a reference document. Should it be moved to be a submenu of `Tools->Unicode`?


> I've attempted to address the above queries in the latest commit:
> 
> 1. Display in order of creation (i.e. can be any order) rather than Unicode order.

I see you did reorder the superscripts -- am I blind, though, or is the superscripted `1` missing altogether now?

> 2. Reduce clutter by hiding duplicates that only vary by case for a non-cased character, e.g. dag/DAG
> 3. Reduce clutter by putting reversibles on the same line, e.g. .* or *. for middle dot instead of separate entries.
> 4. Final column made wider to fit all character names.
> 5. Fixed exception when user clicked in header.

👍

windymilla:
> That's fair. I did notice that with the name change to "Compose Sequence" (which I do like, and I think that the widget can be narrower than you have it if you want it to be), there's a potential confusion with the other widget being named "Compose Sequences".

I can't make it much narrower on my system, because OK, Apply, Cancel buttons need a certain width to fit in.

I've named the Help dialog "List of Compose Sequence" to make it clearer.

windymilla:
> And speaking of "Compose Sequences", I assume it's in the "Help" menu in GG2 because that's where it is in GG1. But the nature of it has changed. In GG1, it's just a reference document. In GG2, it's actually more useful for just clicking to add the character, rather than as a reference document. Should it be moved to be a submenu of `Tools->Unicode`?

I don't really want to do that. I think its primary purpose is help/reference. I just thought I may as well add the "click to insert" feature, since it's easy to add, and once the user has gone to find the sequence for a character, it seems silly to make them dismiss that dialog, then pop the Compose Sequence dialog and type the sequence. I don't think it will be (or should be) a main way to regularly enter common characters. There are other dialogs that do that in GG1, such as the [commonly used characters dialog](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_Manual/Unicode_Menu#The_Commonly-Used_Characters_Dialog) which is much more convenient than searching for the character in the List of Compose Sequences help dialog.

windymilla:
> I see you did reorder the superscripts -- am I blind, though, or is the superscripted `1` missing altogether now?

Not blind, though not "missing" so much as "wrong" - I've accidentally used the "Latin Small Letter I" when coding it, rather than the superscript 1. I had to redo everything manually from GG1 because of the different way characters are represented (Perl couldn't have Unicode characters in the source code of GG1). I'll fix that - thanks

windymilla:
Review fixes committed.
Should fix #301

src/guiguts/application.py
```diff
@@ -586,6 +590,7 @@ def init_help_menu(self) -> None:
         menu_help = Menu(menubar(), "~Help")
         menu_help.add_button("Guiguts ~Manual", self.show_help_manual)
         menu_help.add_button("About ~Guiguts", self.help_about)
+        menu_help.add_button("~Compose Sequences", ComposeHelpDialog.show_dialog)
```
srjfoo, line 593: Two things --
* Should the menu item be changed to match the title bar on the widget?
* This has the same `~` shortcut as line 570, above. Is that okay, since they're in different menus, or does it need to be changed?

srjfoo (approved):

tangledhelix:
## Mac re-test

Seems to work as advertised. I was confused why you'd need the 3 buttons, but I see now the difference between OK and Apply (one closes the window, one leaves it open for re-use). I'm not sure Cancel is needed since you can simply close the window. The `Cmd-;` keybind works.

## Linux re-test

The previous exception is now fixed, so GG runs normally. The compose tool also appears to work as it should. However, the key binding does not work. At first I thought that may be the syntax you used, but when I press `Ctrl-;` a strange little bubble with the letter `e` pops up on screen. It does not look at all Tk-ish, so I didn't think GG was producing it. Some research led me to find that (at least on Ubuntu) this keybind is intercepted by the OS for some kind of emoji character input. It can be turned off, but that would assume the user wants to turn it off or bind it to something else, and users would have to proactively make an OS-level change to make it work.

https://askubuntu.com/questions/1436093/in-kubuntu-what-ctrl-control-semicolon-shortkey-is-assigned-to-i-want

## General feedback

The cancel button seems unnecessary to me, but it's a matter of opinion I suppose.

I think it would be better if Apply didn't clear the input field. What if I wanted to compose the same character in numerous places? I could enter it, Apply, move the cursory, Apply... repeat however I like. But if Apply also clears the input box that isn't possible. I propose not clearing the field on Apply.

Going one step further I wonder if the input should be a pulldown of the last _n_ entries, like the search history in the S&R dialog... but that may be overkill.

src/guiguts/application.py
windymilla: > * Should the menu item be changed to match the title bar on the widget?

I've changed the name, though I don't think they always  have to be the same, since the menu also gives context. This menu item was Help->Compose Sequences, and is now Help->List of Compose Sequences

> * This has the same `~` shortcut as line 570, above. Is that okay, since they're in different menus, or does it need to be changed?

That's okay - as long as they are in different menus, they can have the same shortcut. You would do `Alt+H C` for this help dialog and `Alt+T U C` for the Compose Dialog itself.

🤣 - as I was checking the accuracy of the above sentence, I did come across a menu keystroke clash: `Tools->~Unmatched` and `Tools->~Unicode`. I've change the first of those to `Un~matched`

windymilla:
> However, the key binding does not work. (Ubuntu)

I've changed the key binding to `Cmd/Ctrl+M` on all devices, which is what it is on macOS in GG1. On Windows in GG1 it uses the AltGr key (which is apparently commonly used for compose sequences on Linux/Windows tools that provide the feature), but it's a bit of an odd key IIRC as some people didn't have it on their keyboards or some other confusion, and we had to allow the user to define their own compose key to get round the problem. I think Ctrl+M is safe to use for now (and Compose Sequence definitely needs a shortcut), and if/when we implement user-defined shortcuts, all such problems will magically go away (🤣).

windymilla:
> The cancel button seems unnecessary to me, but it's a matter of opinion I suppose.

I'd rather keep it, since having a cancel button is very common in dialogs, especially those which have other buttons.

> I think it would be better if Apply didn't clear the input field.

Yes, I agree - I've done that

> Going one step further I wonder if the input should be a pulldown of the last _n_ entries, like the search history in the S&R dialog... but that may be overkill.

At the risk of feature creep, I've added that too. When you recall a previously input sequence via the pulldown, it puts it in the text field and waits for you to decide if you want to OK, Apply or Cancel it. I did that for two reasons, one of which is that if you are using this feature you are obviously interacting in an extended manner with the dialog, almost certainly by mouse, rather than a keyboard-only `Ctrl+M ae` (for the ae dipthong), so you can carry on using the mouse. I'll leave you to guess what the second and more compelling reason is 😉

windymilla:
Latest commit is intended to address all the points - let me know if not:

1. Shortcut is Ctrl+M
2. Title of Compose Help dialog changed
3. Compose dialog Apply button does not clear field, but selects the text in it, so the user can just type to enter a new sequence without needing to delete the old one.
4. Field is now a ComboBox, with Pulldown
5. OK, Apply, Cancel buttons no longer accept keyboard focus, otherwise clicking Apply loses focus from the text field, which is annoying.

srjfoo:
> Latest commit is intended to address all the points - let me know if not:
> 
> 1. Shortcut is Ctrl+M

Note that `Cmd`+`m` is a macOS system-level shortcut for minimizing windows. And works quite nicely as such, thankyou. And `Ctrl`+`m` does not work on macOS. However, in the time-honored tradition of pushing buttons to see what they do, I tried `fn`+`Cmd`+`m`, and it does work. I'd like opinions from both @windymilla and @tangledhelix concerning whether this is a reasonable option. I'm assuming that there is nothing in the code that redirects the shortcut to  `fn`+`Cmd`+`m`, so it seems to be an OS-level redirect. Is it a good idea to depend on that? Is there a better choice?

> 2. Title of Compose Help dialog changed ✅ 

Is there something hidden that causes the menu width to be double the existing longest line?
image

> 3. Compose dialog Apply button does not clear field, but selects the text in it, so the user can just type to enter a new sequence without needing to delete the old one.

It should probably be noted in the manual that this only works for the case where it's possible to click on the apply button. The sequences that are predefined enter the character and close the dialog.

> 4. Field is now a ComboBox, with Pulldown 👍 
> 
> 5. OK, Apply, Cancel buttons no longer accept keyboard focus, otherwise clicking Apply loses focus from the text field, which is annoying. 👍

windymilla:
> > 1. Shortcut is Ctrl+M
> 
> Note that `Cmd`+`m` is a macOS system-level shortcut for minimizing windows. And works quite nicely as such, thankyou. And `Ctrl`+`m` does not work on macOS. However, in the time-honored tradition of pushing buttons to see what they do, I tried `fn`+`Cmd`+`m`, and it does work. I'd like opinions from both @windymilla and @tangledhelix concerning whether this is a reasonable option. I'm assuming that there is nothing in the code that redirects the shortcut to `fn`+`Cmd`+`m`, so it seems to be an OS-level redirect. Is it a good idea to depend on that? Is there a better choice?

I feel like I'm slightly going round in circles trying to avoid various 
> 
> > 2. Title of Compose Help dialog changed ✅
> 
> Is there something hidden that causes the menu width to be double the existing longest line? 
> 
> > 3. Compose dialog Apply button does not clear field, but selects the text in it, so the user can just type to enter a new sequence without needing to delete the old one.
> 
> It should probably be noted in the manual that this only works for the case where it's possible to click on the apply button. The sequences that are predefined enter the character and close the dialog.
> 
> > 4. Field is now a ComboBox, with Pulldown 👍
> > 5. OK, Apply, Cancel buttons no longer accept keyboard focus, otherwise clicking Apply loses focus from the text field, which is annoying. 👍

srjfoo:
> > > 1. Shortcut is Ctrl+M
> > 
> > 
> > Note that `Cmd`+`m` is a macOS system-level shortcut for minimizing windows. And works quite nicely as such, thankyou. And `Ctrl`+`m` does not work on macOS. However, in the time-honored tradition of pushing buttons to see what they do, I tried `fn`+`Cmd`+`m`, and it does work. I'd like opinions from both @windymilla and @tangledhelix concerning whether this is a reasonable option. I'm assuming that there is nothing in the code that redirects the shortcut to `fn`+`Cmd`+`m`, so it seems to be an OS-level redirect. Is it a good idea to depend on that? Is there a better choice?
> 
> I feel like I'm slightly going round in circles trying to avoid various OS-specific issues. Would it be possible to get a list of at least a few `Cmd+key` combinations that are macOS shortcuts, please?

I think we dug this out once before, but it doesn't hurt to have it in multiple places: [Mac keyboard shortcuts](https://support.apple.com/en-us/102650).


> > Is there something hidden that causes the menu width to be double the existing longest line?
> 
> No - there's no way to affect that width as far as I know. Since it's a Help menu, it's possible that macOS has some ideas about what could be in it, so is setting the width.

Well, bah! Looking at a branch without the Compose help in the Help menu, I see that it also has about the same amount of extra space to the right of the longest line. That seems silly, but does kind of match with it being automatic. \<sigh>

windymilla:
> 
> I think we dug this out once before, but it doesn't hurt to have it in multiple places: [Mac keyboard shortcuts](https://support.apple.com/en-us/102650).
> 
Thanks - you did dig that out before. This time, I've added a link to it in the [shortcuts section of our GG2 Manual notes page](https://www.pgdp.net/wiki/PPTools/Guiguts2_Manual_Notes#MacOS). Hopefully that will mean I will remember it, and remember where to find it, when we next need a shortcut.

@tangledhelix & @srjfoo - at a quick glance, it looks as though all/most Command sequences are already allocated. There are a few that are not, since they have meaning in some documents, such as Cmd+B, Cmd+I, Cmd+U (bold, italic, underline). Since those have no purpose in GG2, they should be available. I will switch to `Cmd/Ctrl+I` (for "Insert") for now for this PR. I'd prefer not to hold this PR up much longer.

As we've previously agreed, as we add features we will only allocate shortcuts to things that **really** need them. Changing those shortcuts will be very easy in a future PR if/when we settle on a firm policy.

windymilla:
@srjfoo - please could you check that `Cmd+I` pops the Compose Sequence dialog on macOS.
I have checked it on Windows and Linux (WSL)
Thanks

srjfoo:
> @srjfoo - please could you check that `Cmd+I` pops the Compose Sequence dialog on macOS. I have checked it on Windows and Linux (WSL) Thanks

`Cmd`+`i` works just fine to bring up the dialog.

srjfoo (approved):

tangledhelix:
> Note that `Cmd`+`m` is a macOS system-level shortcut for minimizing windows. And works quite nicely as such, thankyou. And `Ctrl`+`m` does not work on macOS. However, in the time-honored tradition of pushing buttons to see what they do, I tried `fn`+`Cmd`+`m`, and it does work. I'd like opinions from both @windymilla and @tangledhelix concerning whether this is a reasonable option. I'm assuming that there is nothing in the code that redirects the shortcut to `fn`+`Cmd`+`m`, so it seems to be an OS-level redirect. Is it a good idea to depend on that? Is there a better choice?

If it works, it's reasonable so long as the user has a `fn` key, I suppose. I don't know our user population well enough to say, but many Mac users out there are using non-Apple keyboards, including PC-style keyboards, which wouldn't have that key. Of course, it's always in the menus as well...

srjfoo:
> > Note that `Cmd`+`m` is a macOS system-level shortcut for minimizing windows. And works quite nicely as such, thankyou. And `Ctrl`+`m` does not work on macOS. However, in the time-honored tradition of pushing buttons to see what they do, I tried `fn`+`Cmd`+`m`, and it does work. I'd like opinions from both @windymilla and @tangledhelix concerning whether this is a reasonable option. I'm assuming that there is nothing in the code that redirects the shortcut to `fn`+`Cmd`+`m`, so it seems to be an OS-level redirect. Is it a good idea to depend on that? Is there a better choice?
> 
> If it works, it's reasonable so long as the user has a `fn` key, I suppose. I don't know our user population well enough to say, but many Mac users out there are using non-Apple keyboards, including PC-style keyboards, which wouldn't have that key. Of course, it's always in the menus as well...

Good point about non-Apple keyboards. That's not an issue for me, primarily because I use laptops.

PR #300, opened 2024-06-23
Add font family & size preferences

windymilla:
Add controls to Prefs dialog to allow user to select the font family and size used in the main text window.
Default is "Courier 10", which is guaranteed by Tk to be available on all platforms.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -345,7 +345,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             PrefKey.TEXT_FONT_FAMILY,
             lambda *value: maintext().set_font(),
         )
-        preferences.set_default(PrefKey.TEXT_FONT_SIZE, 10)
+        preferences.set_default(PrefKey.TEXT_FONT_SIZE, 12)
         preferences.set_callback(
             PrefKey.TEXT_FONT_SIZE,
             lambda *value: maintext().set_font(),
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -341,12 +341,12 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(PrefKey.TEXT_FONT_FAMILY, "Courier")
         preferences.set_callback(
             PrefKey.TEXT_FONT_FAMILY,
-            lambda value: maintext().font.config(family=value),  # type:ignore[arg-type]
+            lambda *value: maintext().set_font(),
         )
         preferences.set_default(PrefKey.TEXT_FONT_SIZE, 10)
         preferences.set_callback(
             PrefKey.TEXT_FONT_SIZE,
-            lambda value: maintext().font.config(size=value),  # type:ignore[arg-type]
+            lambda *value: maintext().set_font(),
         )
 
         # Check all preferences have a default
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -338,6 +338,16 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         preferences.set_default(PrefKey.TEAROFF_MENUS, False)
         preferences.set_default(PrefKey.COMPOSE_HISTORY, [])
+        preferences.set_default(PrefKey.TEXT_FONT_FAMILY, "Courier")
+        preferences.set_callback(
+            PrefKey.TEXT_FONT_FAMILY,
+            lambda value: maintext().font.config(family=value),  # type:ignore[arg-type]
+        )
+        preferences.set_default(PrefKey.TEXT_FONT_SIZE, 10)
+        preferences.set_callback(
+            PrefKey.TEXT_FONT_SIZE,
+            lambda value: maintext().font.config(size=value),  # type:ignore[arg-type]
+        )
 
         # Check all preferences have a default
         for pref_key in PrefKey:
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -338,7 +338,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         preferences.set_default(PrefKey.TEAROFF_MENUS, False)
         preferences.set_default(PrefKey.COMPOSE_HISTORY, [])
-        preferences.set_default(PrefKey.TEXT_FONT_FAMILY, "Courier")
+        # Since fonts aren't available until Tk has initialized, set default font family
+        # to be empty string here, and set the true default later in MainText.
+        preferences.set_default(PrefKey.TEXT_FONT_FAMILY, "")
         preferences.set_callback(
             PrefKey.TEXT_FONT_FAMILY,
             lambda *value: maintext().set_font(),
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -342,14 +342,12 @@ class MainText(tk.Text):
         )
 
         # On some systems, window isn't updated properly, so temporarily select all
-        # then restore selection to force it to update
-        def refresh_selection() -> None:
-            """Stash the selection, select all, then restore selection."""
-            ranges = self.selected_ranges()
-            self.do_select(IndexRange(self.start(), self.end()))
-            self.restore_selection_ranges(ranges)
-
-        self.after_idle(refresh_selection)
+        # then restore selection to force it to update.
+        # Also restore after idle, or Linux version doesn't update.
+        ranges = self.selected_ranges()
+        self.do_select(IndexRange(self.start(), self.end()))
+        self.restore_selection_ranges(ranges)
+        self.after_idle(lambda: self.restore_selection_ranges(ranges))
 
     def toggle_line_numbers(self) -> None:
         """Toggle whether line numbers are shown."""
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -90,17 +90,16 @@ class TextLineNumbers(tk.Canvas):
         self.x_offset = 10
         # Allow for non-zero additional line spacing in text widget
         self.y_offset = self.textwidget["spacing1"]
-        # Allow for 5 digit line numbers
-        width = self.font.measure("88888") + self.x_offset
-        tk.Canvas.__init__(
-            self, parent, *args, width=width, highlightthickness=0, **kwargs
-        )
+        tk.Canvas.__init__(self, parent, *args, highlightthickness=0, **kwargs)
         # Canvas needs to listen for theme change
         self.bind("<<ThemeChanged>>", lambda event: self.theme_change())
         self.text_color = "black"
 
     def redraw(self) -> None:
         """Redraw line numbers."""
+        # Allow for 5 digit line numbers
+        width = self.font.measure("88888") + self.x_offset + 5
+        self["width"] = width
         self.delete("all")
         text_pos = self.winfo_width() - self.x_offset
         index = self.textwidget.index("@0,0")
@@ -149,8 +148,16 @@ class MainText(tk.Text):
         self.frame.rowconfigure(0, weight=1)
 
         # Set up font
+        family = preferences.get(PrefKey.TEXT_FONT_FAMILY)
+        # If preference has never been set, then choose one of the preferred fonts
+        if not family:
+            families = tk_font.families()
+            for pref_font in "DP Sans Mono", "DejaVu Sans Mono", "Courier":
+                if pref_font in families:
+                    family = pref_font
+                    break
         self.font = tk_font.Font(
-            family=preferences.get(PrefKey.TEXT_FONT_FAMILY),
+            family=family,
             size=preferences.get(PrefKey.TEXT_FONT_SIZE),
         )
         # For some reason line spacing on Mac is very tight, so pad a bit here
@@ -251,6 +258,9 @@ class MainText(tk.Text):
 
         maintext(self)  # Register this single instance of MainText
 
+        # Need to wait until maintext has been registered to set the font preference
+        preferences.set(PrefKey.TEXT_FONT_FAMILY, family)
+
     def bind_event(
         self,
         event_string: str,
@@ -328,9 +338,9 @@ class MainText(tk.Text):
         # Mac doesn't update window properly, so temporarily select all
         # then restore selection to force it to update
         if is_mac():
-            self.save_selection_ranges()
+            ranges = self.selected_ranges()
             self.do_select(IndexRange(self.start(), self.end()))
-            self.restore_selection_ranges(self.current_sel_ranges)
+            self.restore_selection_ranges(ranges)
 
     def toggle_line_numbers(self) -> None:
         """Toggle whether line numbers are shown."""
@@ -653,7 +663,9 @@ class MainText(tk.Text):
         """
         if ranges is None:
             ranges = self.prev_sel_ranges
-        if len(ranges) == 1:
+        if len(ranges) == 0:
+            self.clear_selection()
+        elif len(ranges) == 1:
             self.do_select(ranges[0])
         elif len(ranges) > 1:
             col_range = IndexRange(ranges[0].start, ranges[-1].end)
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -87,9 +87,11 @@ class TextLineNumbers(tk.Canvas):
     ) -> None:
         self.textwidget = text_widget
         self.font = tk_font.nametofont(self.textwidget.cget("font"))
-        self.offset = 10
+        self.x_offset = 10
+        # Allow for non-zero additional line spacing in text widget
+        self.y_offset = self.textwidget["spacing1"]
         # Allow for 5 digit line numbers
-        width = self.font.measure("88888") + self.offset
+        width = self.font.measure("88888") + self.x_offset
         tk.Canvas.__init__(
             self, parent, *args, width=width, highlightthickness=0, **kwargs
         )
@@ -100,7 +102,7 @@ class TextLineNumbers(tk.Canvas):
     def redraw(self) -> None:
         """Redraw line numbers."""
         self.delete("all")
-        text_pos = self.winfo_width() - self.offset
+        text_pos = self.winfo_width() - self.x_offset
         index = self.textwidget.index("@0,0")
         while True:
             dline = self.textwidget.dlineinfo(index)
@@ -109,7 +111,7 @@ class TextLineNumbers(tk.Canvas):
             linenum = IndexRowCol(index).row
             self.create_text(
                 text_pos,
-                dline[1],
+                dline[1] + self.y_offset,
                 anchor="ne",
                 font=self.font,
                 text=linenum,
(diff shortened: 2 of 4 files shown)


tangledhelix:
First, thanks for this one, Nigel. The default font before was **so small**... 😃 

Testing on Mac: if I set the font to a new size (e.g. from 10 to 20) but without changing the font itself, the vertical line spacing is not corrected to match. See attached screenshot.

Doing a select-all of the text is enough to fix it, but it doesn't work itself out by default.

Screenshot 2024-06-23 at 4 32 45 PM

tangledhelix:
Forgot to mention - the same thing happens when changing the font itself (without changing size), at least assuming that the two fonts do not have exactly the same line-height.

tangledhelix:
The behavior I describe above does not occur on Ubuntu Linux; there, it behaves as one would expect.

srjfoo:
Thank you! Even with the current Mac gotchas, it's really nice to be able to adjust font and font size.

---

image

In the column selection range ...
* First selected column, second and third selected rows,`q` and `d`. 
* Second selected column, first and second selected rows, `p` and `b`.
* Penultimate selected column, `g` and `i` in second and third selected rows.

In general, the lines are just too close together, even once the line height gets to where it's programmed to be. I may be wrong, but I imagine that as tight as the line height is, at least one person will ask for that to be adjustable, too.

There's actually a good chance that even if you increase the line height to an easier-to-read height, someone will still want it to be customizable :grin: -- after all, I've seen it asked for in the proofreading interfaces. [This](https://stackoverflow.com/questions/65135455/tkinter-text-line-spacing) may or may not be useful

---

image

My test file just happened to have page marker flags turned on. This shows the problem with the highlighting. It doesn't match the selection range when the until after you perform some kind of operation that fully applies the change.

I haven't completely sussed out what will reliably do it. Just changing the font size doesn't. If I change the font size and then change the font, that seems to work. Scrolling seems to do it, but not always in a predictable fashion. As @tangledhelix said, select all seems to work.


---

I don't know if this is possible/easily done, but one thing that I find annoying about font selection in GG1 is that I can't start typing and have the selection range be narrowed down as I type. So I have to scroll down to find the font that I want.

Realistically, I suppose most PPers set the font once and never, or only rarely, change it (which is what I do for BBEdit), so perhaps it's really irrelevant and only something a tester would run into. Perhaps file it as an enhancement issue and forget about it for now?

windymilla:
I'll have a look at those issues - apparently on the Mac, some kind of "refresh" that's needed isn't happening, which would explain the yellow highlight not being updated and the line spacing not updating. 
Thanks for reporting them.

windymilla:
Commit should address the review comments. I don't intend to implement "begin typing to select" feature at this point. There's nothing built-in that does this, so it would need to be written from scratch.

NTS: If we do that in the future, we should derive our own ComboBox from the base one and add this behavior to that, then use that for all ComboBoxes, not just this one. [This code snippet](https://github.com/PySimpleGUI/PySimpleGUI/issues/5120#issuecomment-1019440038) could be useful.

srjfoo:
> Commit should address the review comments. I don't intend to implement "begin typing to select" feature at this point. There's nothing built-in that does this, so it would need to be written from scratch.
> 
> NTS: If we do that in the future, we should derive our own ComboBox from the base one and add this behavior to that, then use that for all ComboBoxes, not just this one. [This code snippet](https://github.com/PySimpleGUI/PySimpleGUI/issues/5120#issuecomment-1019440038) could be useful.

I don't know if this is a useful idea, but some applications also have a secondary part of the font selection menu that contains the 4 or 5 most recently used fonts at the top of the menu, for easy access.  And as @tangledhelix and I both acknowledge, easy font selection is likely not as much of an issue for normal users as it is for testers.

tangledhelix:
> > Commit should address the review comments. I don't intend to implement "begin typing to select" feature at this point. There's nothing built-in that does this, so it would need to be written from scratch.

I agree with your instinct; it's not worth spending time on that. Unless we find it is worth spending time on elsewhere, such as a unicode character picker or something. In that case I guess it could be re-used for this purpose too.

> I don't know if this is a useful idea, but some applications also have a secondary part of the font selection menu that contains the 4 or 5 most recently used fonts at the top of the menu, for easy access. And as @tangledhelix and I both acknowledge, easy font selection is likely not as much of an issue for normal users as it is for testers.

I think we should look at this. It could contain the few most recently used fonts once a user picks some, but we can also be opinionated and choose a few to put in the list. For example, the Tk default (Courier), and if they exist on the system, some others we "suggest" like DP Sans Mono, DejaVu Sans Mono ...  IMO DP Sans Mono is an excellent choice for PP'ing and if it's presented "front and center" it would be very easy for someone to choose it.

I daresay if we can detect that it's installed on the first run of GG2 (when no prefs file exists yet), it could even be the default 😉

windymilla:
> I think we should look at this. It could contain the few most recently used fonts once a user picks some, but we can also be opinionated and choose a few to put in the list. For example, the Tk default (Courier), and if they exist on the system, some others we "suggest" like DP Sans Mono, DejaVu Sans Mono ... IMO DP Sans Mono is an excellent choice for PP'ing and if it's presented "front and center" it would be very easy for someone to choose it.
> 
> I daresay if we can detect that it's installed on the first run of GG2 (when no prefs file exists yet), it could even be the default 😉

The first part of that would be quite easy, although I'm not sure there's any divider available as you often see between the recently used and the remainder. That's unless we dump the combobox idea and write our own which uses a regular menu - I'll have a look to see if that would be necessary - if so, it's not a good use of time ATM.

As for detecting DPSM on the first run, when a Pref hasn't been set, I did take a slightly lazy way out - the problem is that when you attempt to load the Prefs file, Tk hasn't been initialized, because Prefs need to be set very early. So then you can't call Tk routines to query the available fonts. Perhaps I'll be less lazy and implement the idea I previously had, which is to make the default family the empty string, and don't decide until you first need it (i.e. when maintext is created) which font to make the actual default (obviously DPSM if it exists, then perhaps DVSM, then Courier). I'll give it a try, and see what happens.

src/guiguts/application.py
```diff
@@ -332,6 +332,16 @@ def initialize_preferences(self) -> None:
             ),
         )
         preferences.set_default(PrefKey.TEAROFF_MENUS, False)
+        preferences.set_default(PrefKey.TEXT_FONT_FAMILY, "Courier")
+        preferences.set_callback(
+            PrefKey.TEXT_FONT_FAMILY,
+            lambda *value: maintext().set_font(),
+        )
+        preferences.set_default(PrefKey.TEXT_FONT_SIZE, 10)
```
srjfoo, line 340: Just curious -- why set the default to font size to 10?

srjfoo:
Can the menu be made to resize to the longest font name? Or if not, at least lengthened to accommodate `Noto Sans Mono ExtraCondensed`? 😁 A smidge extra padding on the left wouldn't come amiss, either, if possible.

image
image

Was leaving the whole file selected after changing font or font size intentional?

The width column that contains the line numbers should grow with the font size change. Currently, the fifth digit disappears completely with font sizes larger than 16 points/pixels. This screenshot is 24[1]: 

image

 
[1: I was curious about units, and went looking. According to what I found with a quick search (so maybe not so reliable), positive integers are points, negative integers are pixels.]

---

Thank you for increasing the line height. That makes it much more readable!

src/guiguts/application.py
windymilla: Originally because it's Courier 10 in GG1. 
Now I've changed the default to be DPSM if you have it installed, then 10 is actually quite reasonable for that font, whereas it's a little small for Courier. The user only has to change it once if they don't like it. If you make it 12, for example, then I for one, would not want that with DPSM. 🤷

windymilla:
> Can the menu be made to resize to the longest font name? Or if not, at least lengthened to accommodate `Noto Sans Mono ExtraCondensed`? 😁 A smidge extra padding on the left wouldn't come amiss, either, if possible.

Short answer "no". Resizing to the longest font name is too much work for the benefit - sizing works in pixels in some places, and characters in others, but for a non-mono font, the correct number of characters wouldn't guarantee it. Also, I bet there are some looooong font names that some people might have downloaded and installed. Also, there's not a padding easily available within the pulldown menu of the combo box - that's just what Macs do, apparently 😛 

However, I have set the width to 30, which seems pretty long to me. I don't have Noto Sans Mono ExtraCondensed (pity the name isn't extracondensed too!) so am not certain if it will fit completely (also depends on your system font)

> Was leaving the whole file selected after changing font or font size intentional?

No - I tested restoring a previous normal or column selection (after applying Dan's "select all" fix) but hadn't tested restoring a zero selection. If you had no selection previously, the restoration failed and left it with "select all" - now fixed.
 
> The width column that contains the line numbers should grow with the font size change. Currently, the fifth digit disappears completely with font sizes larger than 16 points/pixels. 

Yes - I've made it resize every time it redraws, which doesn't seem to take an appreciable time to do, and will certainly capture when the font changes.

windymilla:
I think all points have been addressed.

src/guiguts/maintext.py
```diff
@@ -310,6 +329,19 @@ def grid(self, *args: Any, **kwargs: Any) -> None:
...
 
+    def set_font(self) -> None:
+        """Set the font for the main text widget, based on the current Prefs values."""
+        self.font.config(
+            family=preferences.get(PrefKey.TEXT_FONT_FAMILY),
+            size=preferences.get(PrefKey.TEXT_FONT_SIZE),
+        )
+        # Mac doesn't update window properly, so temporarily select all
+        # then restore selection to force it to update
+        if is_mac():
```
tangledhelix, line 340: This works on Mac:
- Changing font with no selection changes/redraws as you expect. Nothing is selected after.
- With an existing selection or column select, the selection is preserved after the font change.

But on Linux:
- If nothing is selected, the font change works as desired.
- If a selection is active (either columnar or otherwise) then it is lost with the font change.

This code block is guarded by `is_mac()` which I had guessed was the reason, but adding an `is_x11()` test with the below patch did not fix it. Linux still loses the selection on font change.

```diff
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -14,6 +14,7 @@ import regex as re
 from guiguts.preferences import preferences, PrefKey
 from guiguts.utilities import (
     is_mac,
+    is_x11,
     IndexRowCol,
     IndexRange,
     force_tcl_wholeword,
@@ -337,7 +338,7 @@ class MainText(tk.Text):
         )
         # Mac doesn't update window properly, so temporarily select all
         # then restore selection to force it to update
-        if is_mac():
+        if is_mac() or is_x11():
             ranges = self.selected_ranges()
             self.do_select(IndexRange(self.start(), self.end()))
             self.restore_selection_ranges(ranges)
```

tangledhelix (changes requested):
I don't have any particular suggestion how to fix this on Linux. I can take a stab at it when I have more time if you want, but so far I only tried the obvious (noted in the patch).

I also tried removing the test entirely (in case `is_x11()` wasn't reliably detecting Linux), but that didn't change anything.

src/guiguts/maintext.py
```diff
@@ -146,8 +147,23 @@ def __init__(self, parent: tk.Widget, root: tk.Tk, **kwargs: Any) -> None:
...
+            for pref_font in "DP Sans Mono", "DejaVu Sans Mono", "Courier":
+                if pref_font in families:
+                    family = pref_font
+                    break
+        self.font = tk_font.Font(
+            family=family,
+            size=preferences.get(PrefKey.TEXT_FONT_SIZE),
+        )
+        # For some reason line spacing on Mac is very tight, so pad a bit here
+        line_spacing = 3 if is_mac() else 0
```
srjfoo, line 164: If it's okay with @tangledhelix, I'd like to suggest taking this to 4. Since it's specific to Macs, it shouldn't affect anyone else, and I think it looks slightly better. That being said, Dan, if you prefer 3, we can leave it that way.

srjfoo:
> > Can the menu be made to resize to the longest font name? Or if not, at least lengthened to accommodate `Noto Sans Mono ExtraCondensed`? 😁 A smidge extra padding on the left wouldn't come amiss, either, if possible.
> 
> Short answer "no". Resizing to the longest font name is too much work for the benefit - sizing works in pixels in some places, and characters in others, but for a non-mono font, the correct number of characters wouldn't guarantee it. Also, I bet there are some looooong font names that some people might have downloaded and installed. Also, there's not a padding easily available within the pulldown menu of the combo box - that's just what Macs do, apparently 😛

Ugh. Oh well. You're right that it's probably not worth the time to figure out how to do it if it can be done (both for the menu that grows or shrinks depending on the length of the longest string and for the extra padding to the left). Seems that all of the platforms have their issues.

> However, I have set the width to 30, which seems pretty long to me. I don't have Noto Sans Mono ExtraCondensed (pity the name isn't extracondensed too!) so am not certain if it will fit completely (also depends on your system font)

See review comment.
 
> > Was leaving the whole file selected after changing font or font size intentional?
> 
> No - I tested restoring a previous normal or column selection (after applying Dan's "select all" fix) but hadn't tested restoring a zero selection. If you had no selection previously, the restoration failed and left it with "select all" - now fixed.

Works for me, thanks!

> > The width column that contains the line numbers should grow with the font size change. Currently, the fifth digit disappears completely with font sizes larger than 16 points/pixels.
> 
> Yes - I've made it resize every time it redraws, which doesn't seem to take an appreciable time to do, and will certainly capture when the font changes.

Looks good, thanks!

src/guiguts/misc_dialogs.py
```diff
@@ -19,6 +20,8 @@
     OkApplyCancelDialog,
 )
 
+COMBO_SEPARATOR = "――――――――――――――――――――――――――――――"
```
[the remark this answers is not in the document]
windymilla: Needless to say, on Windows (or at least my default system font), the horizontal bar doesn't have gaps, and just draws a line across the menu. It does disappear off to the right, but that doesn't affect the menu. I made it 30 characters long, which is the width I set for the menu, but of course in a proportional font, that width isn't guaranteed.
image

I'll reduce to 20 and see if that's OK

src/guiguts/maintext.py
```diff
@@ -146,8 +147,23 @@ def __init__(self, parent: tk.Widget, root: tk.Tk, **kwargs: Any) -> None:
         self.frame.columnconfigure(1, weight=1)
         self.frame.rowconfigure(0, weight=1)
 
+        # Set up font
+        family = preferences.get(PrefKey.TEXT_FONT_FAMILY)
+        # If preference has never been set, then choose one of the preferred fonts
+        if not family:
+            families = tk_font.families()
+            for pref_font in "DP Sans Mono", "DejaVu Sans Mono", "Courier":
```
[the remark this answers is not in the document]
windymilla: I may as well add "Courier New" after "Courier" since it won't affect anyone who does have Courier.

src/guiguts/maintext.py
windymilla: No problem to change it to 4

src/guiguts/misc_dialogs.py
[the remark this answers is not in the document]
windymilla: I'll also change the font size to 12 on the basis that it's better to be a bit too large for some people, than a bit too small for other people.

src/guiguts/maintext.py
windymilla: Thanks for looking into it @tangledhelix. Earlier on you wrote

> The behavior I describe above does not occur on Ubuntu Linux; there, it behaves as one would expect.

So, I thought you didn't need that code on Linux (or is it a different flavor?)

One option I could try would be to tell it to do the "stash selection, select all, restore selection" after the system becomes idle - no real reason to think that would fix it, but there's no real reason to explain why it's not working.

src/guiguts/maintext.py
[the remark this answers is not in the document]
windymilla: Actually, I decided to make it choose Courier New if it exists, else Courier.

windymilla:
New commit should address everything mentioned so far.

If the `after_idle` fix has no effect on Ubuntu, then I can revert it to be Mac-only and not `after_idle` since it seems useful there. IMHO, if that happens, the bug it is fixing is so minor and rare (only on Ubuntu, only if the user changes font, only if they had a selection that they would be upset to lose), that it's probably not worth any further effort at this point in time.

src/guiguts/misc_dialogs.py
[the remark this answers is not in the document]
srjfoo: image

Yet another case of different platforms working differently. And yeah, you'd think that with a proportional font, the horizontal bars would fit together for the display, but as you can see from the screenshot, they don't. And yes, 20 is about right. A shame you can't just use a `<hr>`?

srjfoo (approved):
I think all of my concerns have been addressed.

And thanks for adding Courier New. It just occurred to me that it's needed on Macs if Courier doesn't come installed, for those who don't have either DVSM or DPSM installed.

windymilla:
@tangledhelix - I now have a version which works on Windows, and also Windows Subsystem for Linux, so I'm hoping that it will work for your Linux system. Basically, I use your "select all then immediately restore selection" idea, but I also restore the selection again after the system becomes idle. 

Without doing it after idle, the Linux version doesn't seem to update, so any current selection isn't restored.
Without doing it immediately (and leaving the restoration until after idle) the whole file is selected for an extended period of time, which means the flash of full selection is very visible, and also that if something else does a "select all & restore", it could end up restoring the "select all" version from the previous "select all & restore". This actually happens on system startup when the font family, then the font size, are set from preferences, leading to the program starting with the first line of text selected (because when the widget is empty, "select all" selects from 1.0 to end, where end is 2.0 for an empty file.)
Sorry, but you'll have to hack the "semicolon" shortcut to run this branch on Linux - just change the semicolon to "M" in line 580 of application.py: `"Cmd/Ctrl+M",` or comment that line out.

@srjfoo - please could you also check that this latest version still works on macOS - thanks.

srjfoo:
> @srjfoo - please could you also check that this latest version still works on macOS - thanks.

Still seems to work okay as far as I can tell at the moment.

> Sorry, but you'll have to hack the "semicolon" shortcut to run this branch on Linux - just change the semicolon to "M" in line 580 of application.py: "Cmd/Ctrl+M", or comment that line out.

Just curious ... does it just flat not run on Linux without the change? I was confused at first, because the Cmd/Ctrl+M is for the compose sequence dialog, and that's not what we're testing here.

---

BTW, for what it's worth, `Cmd`+ `,` (comma) appears to be the Mac system shortcut to open the preferences panes. It works for gg2, also.

windymilla:
> > @srjfoo - please could you also check that this latest version still works on macOS - thanks.
> 
> Still seems to work okay as far as I can tell at the moment.
> 
> > Sorry, but you'll have to hack the "semicolon" shortcut to run this branch on Linux - just change the semicolon to "M" in line 580 of application.py: "Cmd/Ctrl+M", or comment that line out.
> 
> Just curious ... does it just flat not run on Linux without the change? I was confused at first, because the Cmd/Ctrl+M is for the compose sequence dialog, and that's not what we're testing here.

No, because this branch doesn't have the edit to fix `Ctrl+;` that I made in the Compose branch. So, this branch fails on startup on Linux when it tries to bind to the `Ctrl+;` keystroke, even if you never try to use the shortcut.

src/guiguts/maintext.py
tangledhelix: 4 looks perfectly fine to me.

tangledhelix:
@windymilla In general the font resizing / changing the font-face both look good to me on Mac & Linux. If there is a selection it is preserved (regular or columnar selection). I don't see any noticeable flashing or anything. Looks good to merge.

PR #305, opened 2024-06-25
Preserve geometry of maximized main window

### LINKED ISSUE #302: Maximizing doesn't store main window position properly
This bug is in GG1 too:

On Windows at least, if you do the following, you won't get what you expect:
1. Ensure main window is not maximized.
2. Resize window so it's about half the screen size in both directions.
3. Position window away from the top-left corner of the screen, e.g. put the top-left of the window near the center of the screen.
4. Maximize window
5. Quit GG
6. Restart GG

GG will have its top-left in the position you located it in step 3, but will have its size at the maximized size (so some of the window goes off the right & bottom of the screen. 

It's pretty obvious in this case, but less obvious if the offsets are smaller, e.g. a user complained their status bar vanished off the bottom, cause by this bug.

NTS: See if it's possible to intercept/control maximizing - could store maximized flag in GGprefs, and make sure the position & size are the non-maximized values, and when GG starts, maximize if it was maximized on exit. If not, then maybe if maximized, force position to (0,0). Or (worst option), don't store size when maximizing & document that maximizing doesn't work well (on Windows) and explain how to resize window so it's more or less maximized without using the maximize button.

windymilla:
There is a bug/limitation in how the geometry of the root window is reported when the window is maximized (at least on Windows). It reports that the window is the size of the full screen, but that the top left corner of the window is wherever it was before maximizing.

This means (like GG1) that when you quit and restart, GG2 starts with its top left somewhere on the screen, but the size is the full screen size, so some of the window vanishes off the bottom & right.

Fixed by also saving the "state" flag (which is normal or zoomed) and restoring that when the program is
restarted. Also, not saving the bad geometry (i.e. big window but not at top-left) reported after maximizing.

Fixes #302

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -302,6 +302,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(PrefKey.SEARCHDIALOG_REGEX, False)
         preferences.set_default(PrefKey.DIALOG_GEOMETRY, {})
         preferences.set_default(PrefKey.ROOT_GEOMETRY, "800x400")
+        preferences.set_default(PrefKey.ROOT_GEOMETRY_STATE, "normal")
         preferences.set_default(PrefKey.DEFAULT_LANGUAGES, "en")
         preferences.set_default(PrefKey.WFDIALOG_SUSPECTS_ONLY, False)
         preferences.set_default(PrefKey.WFDIALOG_IGNORE_CASE, False)
--- a/src/guiguts/preferences.py
+++ b/src/guiguts/preferences.py
@@ -41,6 +41,7 @@ class PrefKey(StrEnum):
     CHECKERDIALOG_SUSPECTS_ONLY = auto()
     DIALOG_GEOMETRY = auto()
     ROOT_GEOMETRY = auto()
+    ROOT_GEOMETRY_STATE = auto()
     DEFAULT_LANGUAGES = auto()
     JEEBIES_PARANOIA_LEVEL = auto()
     WRAP_LEFT_MARGIN = auto()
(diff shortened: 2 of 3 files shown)


windymilla:
Testing notes: don't know if the issue this addresses happens on non-Windows platforms, and if it does whether this PR will fix it on non-Windows. On Windows, it does seems to though:
1. Have non-maximized window mid-screen
2. Maximize it
3. Quit program
4. Restart
5. Program should be maximized.
6. De-maximize
7. Program should be back to mid-screen as in point 1.

srjfoo (approved):

windymilla:
Testing with WSL: Linux systems don't support the `state()` method to get/set whether the window is maximized. There is an alternative, but it only sets, and the query always returns false, so it's no use.
Behavior is OK on Linux though - if maximized, then quit, then restarted, the resulting window is almost maximized.

srjfoo:
> Testing with WSL: Linux systems don't support the `state()` method to get/set whether the window is maximized. There is an alternative, but it only sets, and the query always returns false, so it's no use. Behavior is OK on Linux though - if maximized, then quit, then restarted, the resulting window is almost maximized.

Wow -- for once, macOS is the only one that does what's expected?

tangledhelix:
I'm late to my review (sorry), but on Linux I think I get the same result as you did, possibly. I'm using a VM on my Mac, not WSL. I'm not sure how they differ but I figured it was worth testing in both.

When I start from scratch (no prefs file), I get a window about 1/4 of the screen. Moving to the middle, then maximizing, then quitting, then restarting - it starts maximized, as desired. De-maximizing, it does not go back to the prior geometry or position. It just shrinks to something like 3/4 the size of the screen (full height, but not full width).

So it's not the most desirable outcome, I guess, but it doesn't push offscreen or do anything weird. (Ubuntu 22.04)

PR #313, opened 2024-06-28
Implement Knuth-Plass wrapping

windymilla:
If parameters are the same, wrapping behavior should now be identical to GG1.

This PR does not affect processing of block markup, just the actual wrapping of the words.

### DIFF
--- a/src/guiguts/utilities.py
+++ b/src/guiguts/utilities.py
@@ -360,3 +360,318 @@ class DiacriticRemover:
             for char in normalize("NFD", string.translate(DiacriticRemover.outliers))
             if not combining(char)
         )
+
+
+class TextWrapper:
+    """Provides Knuth-Plass rewrapping functionality.
+
+    Can be replaced with Python's textwrap/TextWrapper if "greedy" wrapping is preferred:
+        ``from textwrap import TextWrapper``
+
+        ``wrapper = TextWrapper(break_long_words=False, break_on_hyphens=False)``
+    """
+
+    def __init__(self) -> None:
+        """Initialize rewrap parameters."""
+
+        # The four attributes below match the Perl version's four
+        # options. However the Python version is normally passed
+        # just three of them, namely:
+        #
+        #    width
+        #    first_indent
+        #    subsequent_indent
+        #
+        # The Python version will calculate a default value for
+        # 'optimum_width' (see 'def fill' below) unless the caller
+        # sets a value for this attribute.
+        #
+        # Only consider setting a value for 'optimum_width' if you
+        # want to explore different optimal rewrappings. Using the
+        # default value means that Knuth-Plass wrapping behaviour
+        # is indentical in GG1 & GG2.
+
+        self.width = 72
+        self.optimum_width = -1
+        self.initial_indent = ""
+        self.subsequent_indent = ""
+
+    def fill(self, paragraph: str) -> str:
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes:

Select some text and wrap it. Compare results with GG1.

windymilla:
Many thanks to @G4OEU for doing all the work (which I didn't really fancy doing myself)

src/guiguts/utilities.py
```diff
@@ -360,3 +360,318 @@ def remove_diacritics(cls, string: str) -> str:
...
+        penaltylimit = 0x2000000
+        # Will hold the newline-terminated lines that are split from
+        # the passed in paragraph text. This is the source of the
+        # lines to be rewrapped by the Knuth algorithm.
+        from_lines = []
+        # Will hold the trimmed and newline-terminated lines that are
+        # the result of the Knuth rewrapping.
+        to_lines = []
+        # Used by reflow_trial() to pick the break for the last line
+        # which gives the least penalties penalties for previous lines.
```
srjfoo, line 438: penalties penalties?

src/guiguts/utilities.py
```diff
@@ -360,3 +360,318 @@ def remove_diacritics(cls, string: str) -> str:
...
+                # Save these breaks if they are an improvement.
+                if bestsofar < best:
+                    best_lastbreak = lastbreak
+                    linkbreak = my_linkbreak.copy()
+                    best = bestsofar
+                # Next opt
+            return best_lastbreak
+
+        def paragraph_to_lines(paragraph: str) -> None:
+            """Split paraqraph string into lines, keeping trailing empty lines.
```
srjfoo, line 524: paraqraph -> paragraph?

src/guiguts/utilities.py
```diff
@@ -360,3 +360,318 @@ def remove_diacritics(cls, string: str) -> str:
...
+            """
+            nonlocal words
+
+            # Protect " . . ." ellipses by replacing space with unused byte \x9f.
+            line = re.sub(r" \. \. \.", r"\x9f.\x9f.\x9f.", line)
+            line = re.sub(r"\. \. \.", r".\x9f.\x9f.", line)
+            # Splitting assumes a single space between words in 'line'
+            line = re.sub(r"\s+", " ", line)
+            # Remove any extra space at end of line.
+            line = re.sub(r"\s$", "", line)
```
srjfoo, line 639: Should that be 

line = re.sub(r"\s`+`$", "", line)

In case there are multiple spaces at the end of a line?

windymilla: Thanks for checking so carefully. In this case, it's not needed, because the line before replaces any multiple spaces with a single space.

windymilla:
Typos fixed

src/guiguts/utilities.py
srjfoo: > Thanks for checking so carefully. In this case, it's not needed, because the line before replaces any multiple spaces with a single space.

Oops, I saw that, but didn't bring it forward 🤦‍♀️

srjfoo (approved):

src/guiguts/utilities.py
windymilla: Well, I actually added the `+` in the code 🤦‍♀️, but once the `\s+` was in the code, my brain spotted the exact same string on the line above. :)

PR #315, opened 2024-06-30
Allow leading-zero omission in Goto Image

### LINKED ISSUE #307: Go to Img number doesn't ignore leading zeros like GG1 does.
From Slack:

> When I click the Status Bar icon (Img:028), the "Go To Page" window appears. However, entering "15" in the "Image Number" field and clicking "OK" doesn't work. The window closes, but I remain on page 28.

> To see if the format was the issue, I even tried entering the actual image filename, "015.png", in the "Img" box. Unfortunately, that didn't work either.

NTS: In GG1 (unsurprisingly) if the image number format is something like `a0015`, then typing `15` won't work. It's only leading zeros that can be omitted.

windymilla:
When clicking on `Img:` in status bar, allow user to type `3` when image name is actually `003`, up to 4 leading zeroes.
This mimics GG1 behavior.

Fixes #307

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -580,15 +580,20 @@ class File:
         """Go to page corresponding to the given image number.
 
         Args:
-            image_num: Number of image to go to, e.g. "001"
+            image_num: Number of image to go to, e.g. "001" (can omit leading zeroes)
         """
-        if image_num is not None:
+        if image_num is None:
+            return
+        for n_zero in range(5):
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes - make sure your test file has image names like `001, 002, etc`, not `a001, a002, etc` since only leading zeros are allowed for, not other characters.

srjfoo (approved):

PR #317, opened 2024-06-30
Make cursor visible after paste

### LINKED ISSUE #308: Put cursor at end of pasted text
This will match GG1's (and other editors') behavior.

windymilla:
Fixes #308

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -798,6 +798,8 @@ class MainText(tk.Text):
 
     def smart_paste(self) -> str:
         """Do column paste if multiple ranges selected, else default paste."""
+        # Afterwards, make sure insert cursor is visible
+        self.after_idle(lambda: self.see(tk.INSERT))
         if len(self.selected_ranges()) <= 1:
             return ""  # Permit default behavior to happen
         self.column_paste()


windymilla:
Notes for testing (reported via Slack)

> Pasting on the last line can still cause an issue. You won't see the pasted text at first, but the scrollbar will jump. Hitting the left arrow key reveals the pasted content, suggesting the screen update failed to keep the cursor in view.

Copy a paragraph, go to the end of a longish file (over 1 page of text), and paste. On Windows (at least) all the pasted text went below the bottom of the window, and the window didn't scroll up so you could see it. Fixed in this PR.

srjfoo (approved):
For what it's worth, in (the current) `master`, macOS, the text is pasted in and visible, and the cursor exists, but the display doesn't advance to show the cursor. In this branch, it does. [** edit: Looks like it's behaving the same, actually. I mis-read the original issue, I think, to mean that the text didn't appear.]

windymilla:
Could be windows only bug. As long as it's not broken on macOS

srjfoo:
> Could be windows only bug. As long as it's not broken on macOS

The only thing that could be termed broken was the visible text not advancing to show the cursor. Note that it's not just at the end of the file. I just went back in `master` and selected some text, and pasted it in at the end of the last visible line in the viewing area. It did not advance. In the branch, it did, so I think that's fixed.

PR #318, opened 2024-07-03
Wf bugs

### LINKED ISSUE #303: WF ae & oe ligature check not working correctly
Reported via Slack:

> The Word Frequency Ligature check does not work properly.  I selected Suspects Only, Sort: Aleph, and Ligatures.  The result was a complete list of all ae and oe ligatures in the project. There were no actual æ or œ characters in this file so I expected an empty list since I selected the Suspects only option.

### LINKED ISSUE #306: Improve WF Ital/Bold/Etc
Reported via Slack:

> The Word Frequency Ital/Bold/SC/Etc. - Word Threshhold Sort: Alpha ignores the initial tag in the sort.  That could be useful for folks looking for consistency issues especially if they have the Suspects Only option ticked.  I'd like to see an option which includes the initial tag in the sort.  This is helpful to me in looking for books or other works that need <cite> markup and to find foreign phrases which need language markup.

> Yes, that is the way GG1.6 works.  I prefer to work on one category of emphasis markup at a time.  Italic markup may mean a citation `<cite>`, high emphasis `<em>`, low emphasis `<i>`, or the formatters may have used it for underline `<u>`.  Bold also could be either `<b>` or `<strong>` emphasis

> Also, I would like to see `<cite>, <em>, <u>, and <strong>` tags included in the Word Frequency Ital/Bold/SC/Etc. - Word Threshhold Sort: Alpha.

windymilla:
Fix WF ligature bug
Although it was reporting the number of suspects
correctly, it was listing words that did not have a
ligature in the Suspects Only list, even when the
ligatured version didn't exist.

Fixes #303 

---

Make WF Ital/Bold include markup in alpha sort
GG1 includes the markup when doing alpha sort, but
the first GG2 attempt kept the marked and unmarked
words together by stripping markup.
This wasn't popular, so reverted to GG1 behavior.

Also add cite, em & strong to list of markups detected:
`i|b|sc|f|g|u|cite|em|strong`

Fixes #306

### DIFF
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -448,35 +448,19 @@ class WordFrequencyDialog(ToplevelDialog):
         """Display all the stored entries in the dialog according to
         the sort setting."""
 
-        def no_markup_key(word: str) -> tuple[str, ...]:
-            """Return additional sort keys to keep identical marked-up and non_marked-up phrases together."""
-            if preferences.get(PrefKey.WFDIALOG_DISPLAY_TYPE) == WFDisplayType.MARKEDUP:
-                no_markup = re.sub("<.*?>", "", word)
-                return (no_markup.lower(), no_markup)
-            return ()
-
         def sort_key_alpha(
             entry: WordFrequencyEntry,
         ) -> tuple[str, ...]:
             no_dia = DiacriticRemover.remove_diacritics(entry.word)
-            return no_markup_key(no_dia) + (no_dia.lower(), no_dia, entry.word)
+            return (no_dia.lower(), no_dia, entry.word)
 
         def sort_key_freq(entry: WordFrequencyEntry) -> tuple[int | str, ...]:
             no_dia = DiacriticRemover.remove_diacritics(entry.word)
-            return (
-                (-entry.frequency,)
-                + no_markup_key(no_dia)
-                + (no_dia.lower(), no_dia, entry.word)
-            )
+            return (-entry.frequency,) + (no_dia.lower(), no_dia, entry.word)
 
         def sort_key_len(entry: WordFrequencyEntry) -> tuple[int | str, ...]:
             no_dia = DiacriticRemover.remove_diacritics(entry.word)
-            no_markup = no_markup_key(no_dia)
-            if len(no_markup) == 0:
-                length = len(entry.word)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
1. Add `Mediaeval` and `Mediæval` to a file. Use WF Ligatures, and turn on/off Suspects only. It should correctly show them both. Now, delete one of those two words and re-check. Then restore that word and delete the other and re-check. You will see that it's partly treating `Mediaeval` as a suspect, even when the file doesn't contain `Mediæval`.
2. Have pairs, like `across` and `<i>across</i>`, `great` and `<b>great</b>`. Use WF Ital/Bold/etc check with Alpha sorting. Note that in GG1 (and this PR) the order will be:  `<b>great</b>`, `<i>across</i>`, `across`, `great`, i.e. all bold words, all italic words, etc., then non-marked words in alphabetic order. However, in master, the order is `<i>across</i>`, `across`, `<b>great</b>`, `great`, i.e. sorted by the base word, regardless of markup. Also note the additional supported markup types.

srjfoo (approved):
This branch behaved as I expected. That is, it listed words with actual ligatures if "suspects only" was not checked, and hid them if it was. 

I tested against gg2, current `master` and against gg1.6.2. Using the same file, this was what was displayed:
image

Note that "suspects only" is not checked. Shouldn't the words with ligatures have been listed? If not, I don't understand how it's supposed to work.

In any case, approved, because this branch works as I'd expect it to work, but if it's supposed to work the way gg1 does, either it doesn't, or I have my gg1 settings wrong somewhere (when testing gg1, I generally run using `--nohome`).

Let me know if you want a copy of the file I used for testing.

windymilla:
> Let me know if you want a copy of the file I used for testing.

Yes please.

What should happen without Suspects Only, is that any word with ae, oe or the ligatures will be listed, e.g.
```
Mediaeval
Mediæval
anæmia
fae
aeroplane
```

If you select Suspects Only, then it should only list those that exist with & without the ligature, in the above case, `Mediaeval `and `Mediæval`. They are suspect because they exist in both forms in the file.

windymilla:
Thanks for the test file @srjfoo
I have confirmed that GG2 behaves as GG1, which is what we want.

PR #320, opened 2024-07-06
Improve single quote "unmatched" check

### LINKED ISSUE #312: False positives of right single quotes on the Unmatched Curly Quotes Report
The report identified 355 instances of curly quotes that couldn't be matched with their corresponding opening or closing quotes. These are possessives ('s), but the system treats them as unmatched quotes.

Suggestion:

To reduce the number of false positives, consider excluding possessives from the Unmatched Curly Quotes Report. This would focus the report on true mismatched quotes.

Alternative:

If some users must manually review these possessives, then please provide an option to filter or hide them within the report to streamline the process and avoid unnecessary checks for others.

windymilla:
Single quote check needs to ignore things which are "clearly" apostrophes. For now, that means close-quote which has alphabetic character on both sides.

This logic didn't fit into the generic "unmatched" code, so extracted some customized code for single curly quotes.

Fixes #312

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -21,6 +21,7 @@ logger = logging.getLogger(__package__)
 BLOCK_TYPES = "[$*XxFf]"
 POEM_TYPES = "[Pp]"
 ALL_BLOCKS_REG = f"[{re.escape('#$*FILPXCR')}]"
+QUOTE_APOS_REG = "[[:alpha:]]’[[:alpha:]]"
 
 
 def tool_save() -> bool:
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
Things like
```
‘SWEATED teacher’s INDUSTRY’
AND THE’
‘MINIMUM WAGE
```
should only report the single quotes that are not matched with their pair. Without this PR, the close quote at the end of "INDUSTRY" is reported as unmatched, because it thinks the quote in "teacher's" matched the one at the start of "SWEATED".

The "algorithm" for considering a quote to be an apostrophe (& therefore ignoring it) is that it must be a close quote, with an alphabetic character on both sides. This does not trap `teachers'` of course, but you don't know if that was a close quote after "teachers" or an apostrophe without more context that a simple algorithm can handle.

windymilla:
Rebased to include previous WF bug fixes

srjfoo:
I took my test file, added your suggested change, and then went through and changed all of the single straight quotes to (hopefully) the correct curl. The file has a _**lot**_ of plural possessives (`s’`), all of which are flagged. I'm assuming that this is just luck of the draw -- there's no good way to tell the difference between `s’ ` or `s’.` as a possessive or end of phrase/sentence, so the PPer gets to decide. But no harm in asking.

[projectID6582f65017dd2-apos.txt.zip](https://github.com/user-attachments/files/16117759/projectID6582f65017dd2-apos.txt.zip)

windymilla:
> I took my test file, added your suggested change, and then went through and changed all of the single straight quotes to (hopefully) the correct curl. The file has a _**lot**_ of plural possessives (`s’`), all of which are flagged. I'm assuming that this is just luck of the draw -- there's no good way to tell the difference between `s’ ` or `s’.` as a possessive or end of phrase/sentence, so the PPer gets to decide. But no harm in asking.

No harm indeed, but as you say - no way to tell, so sounds like it's working as expected?

srjfoo (approved):

PR #321, opened 2024-07-10
Implement "busy" notification

### LINKED ISSUE #299: Implement "busy" state
Either a busy cursor, like GG1

Or, a busy widget somewhere (status bar I guess) that either changes color, or becomes animated when a long operation starts.

NTS: Use after_idle to restore "non-busy" status

windymilla:
1. "Working" label shows in bottom right corner.
2. Cursor changes to "watch" image while busy.

Fixes #299

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -22,7 +22,7 @@ from guiguts.utilities import (
     is_mac,
     sing_plur,
 )
-from guiguts.widgets import ToplevelDialog, TlDlg, mouse_bind
+from guiguts.widgets import ToplevelDialog, TlDlg, mouse_bind, Busy
 
 MARK_ENTRY_TO_SELECT = "MarkEntryToSelect"
 HILITE_TAG_NAME = "chk_hilite"
@@ -253,6 +253,7 @@ class CheckerDialog(ToplevelDialog):
             args: Optional args to pass to dialog constructor.
             kwargs: Optional kwargs to pass to dialog constructor.
         """
+        Busy.busy()
         return super().show_dialog(title, destroy, **kwargs)
 
     @classmethod
@@ -436,6 +437,7 @@ class CheckerDialog(ToplevelDialog):
         """Display all the stored entries in the dialog according to
         the sort setting."""
 
+        Busy.busy()
         sort_key: Callable[[CheckerEntry], tuple]
         if (
             preferences.get(PrefKey.CHECKERDIALOG_SORT_TYPE)
@@ -508,6 +510,7 @@ class CheckerDialog(ToplevelDialog):
                     self.select_entry_by_index(index)
                     break
         self.update_count_label()
+        Busy.unbusy()
 
     def showing_suspects_only(self) -> bool:
         """Return whether dialog is showing Suspects Only.
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -28,6 +28,7 @@ from guiguts.widgets import (
     ToolTip,
     themed_style,
     theme_set_tk_widget_colors,
+    Busy,
 )
 
 logger = logging.getLogger(__package__)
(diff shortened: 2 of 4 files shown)


srjfoo:
The "working..." message shows very timely, but the clock doesn't show up until right as the job finishes, if then. For the 300-page project I tried, running pptxt took long enough that the macOS system rotating beachball came up, but I didn't see the clock (yes, I did see `working...`). The clock also doesn't appear to come up if the cursor is not hovering over the widget that's working.

srjfoo:
Followup: After thinking about it, I'm guessing the clock works just fine on Windows. The notes above are about macOS. I'd appreciate a sanity check on my macOS findings from @tangledhelix 

After quitting the`busy` branch and bringing it up again, I can confirm that the 700+-page project loaded much more quickly, though not instantaneously, when loaded into a newly opened application.

tangledhelix:
> The "working..." message shows very timely, but the clock doesn't show up until right as the job finishes, if then. For the 300-page project I tried, running pptxt took long enough that the macOS system rotating beachball came up, but I didn't see the clock (yes, I did see `working...`). The clock also doesn't appear to come up if the cursor is not hovering over the widget that's working.
> 
> * `working...` was displayed in a timely fashion
> * Instead of the clock, I saw the system busy indicator until just as everything was loaded and `working...` disappeared.

I had a similar experience:

- "Working" seems to work just as I'd expect
- For an operation only lasting a few seconds or less, I saw the watch cursor
- For longer-blocking operations it seemed to switch from watch to beachball (and in my test this happened a lot as I used a 6MB input file)

I think the current situation is likely fine. Mac users are used to seeing the beachball sometimes while a thing is working, so I don't think it will be confusing or any issue. With my input file, WF re-run was pretty slow, but I used an unrealistically large file (or at least a very atypically-large one).

srjfoo:
> > The "working..." message shows very timely, but the clock doesn't show up until right as the job finishes, if then. For the 300-page project I tried, running pptxt took long enough that the macOS system rotating beachball came up, but I didn't see the clock (yes, I did see `working...`). The clock also doesn't appear to come up if the cursor is not hovering over the widget that's working.
> > 
> > * `working...` was displayed in a timely fashion
> > * Instead of the clock, I saw the system busy indicator until just as everything was loaded and `working...` disappeared.
> 
> I had a similar experience:
> 
>     * "Working" seems to work just as I'd expect
> 
>     * For an operation only lasting a few seconds or less, I saw the watch cursor
> 
>     * For longer-blocking operations it seemed to switch from watch to beachball (and in my test this happened a lot as I used a 6MB input file)
> 
> 
> I think the current situation is likely fine. Mac users are used to seeing the beachball sometimes while a thing is working, so I don't think it will be confusing or any issue. With my input file, WF re-run was pretty slow, but I used an unrealistically large file (or at least a very atypically-large one).

Interesting -- I guess my file just wasn't large enough to see the switching back and forth between the watch and the beah ball. I only saw the watch at the end (if then), when the wait was finishing. If you're not concerned about it -- then it probably is just the system overriding the visible busy indicator. Thanks!

srjfoo (approved):

windymilla:
Thanks @tangledhelix & @srjfoo. It sets the cursor to the "watch" shape at the start of the process, and restores it at the end. It does that for every widget (bar a few exceptions) in the tree hierarchy from the root downwards. It also forces an update (allegedly) on each of those widgets at the start, so they should in theory display the cursor correctly. We'll accept it as-is for now, with the "Working" label as a backup notification for systems where the cursor doesn't change when it should.

tangledhelix:
FWIW, on Linux it seems to work perfectly fine. Instead of the "watch" icon on Mac, it uses a small spinner icon, but it works just like I would expect, and the "working" label in the status bar also works correctly. Tested with a 5MB file to ensure some wait times would come up, and no problems there.

srjfoo:
I'm fine with it the way it is on macOS -- in fact, I would have been fine with just the red `working` in the status bar (but I guess some iconic (😁) representation is expected?

PR #322, opened 2024-07-10
Show image from page label config dialog

### LINKED ISSUE #295: "Configure Page Labels" needs an option to display the selected image
Could possibly work by going to the start of that image and using AutoImg? At least as a first attempt.

windymilla:
When user clicks on a row in the dialog, jump to the page in the file, so user can see the text there to aid their decision about which type of page it is. If user has Auto Img turned on, the image will show in the image viewer.

Fixes #295

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -644,7 +644,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         the_statusbar.add(
             "img",
             tooltip="Click: Go to image",
-            update=lambda: "Img: " + self.file.get_current_image_name(),
+            update=lambda: "Img: " + maintext().get_current_image_name(),
         )
         the_statusbar.add_binding("img", "ButtonRelease-1", self.file.goto_image)
 
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -479,42 +479,6 @@ class File:
         """
         return maintext().page_mark_next("1.0") != ""
 
-    def get_current_page_mark(self) -> str:
-        """Find page mark corresponding to where the insert cursor is.
-
-        Returns:
-            Name of preceding mark. Empty string if none found.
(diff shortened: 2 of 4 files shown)


srjfoo:
Testing on macOS, clicking on the image name in the Configure Page Labels widget takes you to the image, but, while using the arrow to move up and down in the list does actually move the row highlighted, it does not navigate between images. I.e. only clicking will do that. Is it possible for using the arrows to move up and down the list can also navigate to the correct image?

windymilla:
> Is it possible for using the arrows to move up and down the list can also navigate to the correct image?

I'll look into it - should be possible

windymilla:
Two commits:
1. Make arrow key selection show the correct page (& image if Auto Img is on), not just mouse selection.
2. When the dialog pops, select the current page's settings in the dialog, and scroll the list so the current page is just below the top of the dialog.

srjfoo:
> Two commits:
> 1. Make arrow key selection show the correct page (& image if Auto Img is on), not just mouse selection.
> 2. When the dialog pops, select the current page's settings in the dialog, and scroll the list so the current page is just below the top of the dialog.

I must be missing something, because this is not working any differently for me than the previous version. (macOS)

srjfoo:
> > Two commits:
> > 
> > 1. Make arrow key selection show the correct page (& image if Auto Img is on), not just mouse selection.
> > 2. When the dialog pops, select the current page's settings in the dialog, and scroll the list so the current page is just below the top of the dialog.
> 
> ~I must be missing something, because this is not working any differently for me than the previous version. (macOS)~

Update -- I decided to check one other way -- for the previous test, I thought I'd updated my existing branch. Apparently I hadn't. I deleted that old branch and tried again, and it worked just fine, thank you!

srjfoo (approved):

PR #329, opened 2024-07-17
Fix maximize function on Mac

### LINKED ISSUE #323: Starting maximized on Linux, menus are invisible
On Ubuntu 22.04 LTS, if Guiguts is started in maximized mode, the menus are not visible. That is, when Guiguts was maximized, and then closed, and reopened, so it starts in maximized mode.

Nothing interesting was logged by `-d` flag. Un-maximizing returns the menus to their proper place (even without a restart).

Screenshot:

Screenshot 2024-07-11 at 10 46 00 AM

tangledhelix:
Improvements to make dialogs work properly on Mac. Since this is based on prior work of @windymilla it should also fix the Linux menu bar issues in #323.

I tested this on Linux, where it seems to work fine and just as you'd expect.

@windymilla if you would, please give it a try on Windows.

On Mac, it works with the following caveats:
- The dialogs are marked transient, but only when in fullscreen. So if you are in fullscreen and then open dialogs, they are always-on-top.
- This is better than the "tab of the main window" that was happening before, IMO, by far.
- If you open dialogs while not in fullscreen, they don't have the always-on-top behavior.
- One edge case I found: if you have dialogs already open, then go fullscreen, the main window is converted to a desktop of its own and the dialogs are left behind in the desktop where they already lived. On reflection I think this is actually the usual behavior when you fullscreen an app with multiple windows or with open dialogs. E.g., if you open the Safari browser, then open its preferences (which is a dialog), and then fullscreen the Safari window, the exact same thing will happen. So this is actually normal behavior on this platform.

Fixes #323 
Closes #325 

(Added close of 325 because this is rebased onto that branch already)

### DIFF
--- a/src/guiguts/root.py
+++ b/src/guiguts/root.py
@@ -10,7 +10,7 @@ from types import TracebackType
 from typing import Any
 
 from guiguts.preferences import preferences, PrefKey
-from guiguts.utilities import is_x11, is_mac
+from guiguts.utilities import is_x11
 
 logger = logging.getLogger(__package__)
 
@@ -78,9 +78,6 @@ class Root(tk.Tk):
         By setting flag now, and queuing calls to _save_config,
         we ensure the flag will be true for the first call to
         _save_config when process becomes idle."""
-        # Full Screen behaves oddly on Macs, so if user tries, revert immediately
-        if is_mac() and root().wm_attributes("-fullscreen"):
-            root().wm_attributes("-fullscreen", False)
         self.save_config = True
         self.after_idle(self._save_config)
 
--- a/src/guiguts/root.py
+++ b/src/guiguts/root.py
@@ -1,5 +1,6 @@
 """Handle Tk root window"""
 
+
 from enum import StrEnum, auto
 import logging
 import traceback
--- a/src/guiguts/root.py
+++ b/src/guiguts/root.py
@@ -1,5 +1,7 @@
 """Handle Tk root window"""
 
+
+from enum import StrEnum, auto
 import logging
 import traceback
 import tkinter as tk
@@ -9,11 +11,20 @@ from typing import Any
 
 from guiguts.maintext import maintext
 from guiguts.preferences import preferences, PrefKey
+from guiguts.utilities import is_x11
 from guiguts.widgets import grab_focus
 
 logger = logging.getLogger(__package__)
 
 _the_root = None  # pylint: disable=invalid-name
+
+
+class RootWindowState(StrEnum):
+    """Enum class to store root window states."""
+
+    NORMAL = auto()
+    ZOOMED = auto()
+    FULLSCREEN = auto()
 
 
 class Root(tk.Tk):
@@ -26,7 +37,17 @@ class Root(tk.Tk):
 
         super().__init__(**kwargs)
         self.geometry(preferences.get(PrefKey.ROOT_GEOMETRY))
-        self.state(preferences.get(PrefKey.ROOT_GEOMETRY_STATE))
+
+        # Set zoomed/fullscreen state appropriately for platform
+        state = preferences.get(PrefKey.ROOT_GEOMETRY_STATE)
+        if state == RootWindowState.ZOOMED:
+            if is_x11():
+                root().wm_attributes("-zoomed", True)
+            else:
+                self.state("zoomed")
+        elif state == RootWindowState.FULLSCREEN:
+            self.wm_attributes("-fullscreen", True)
+
         self.option_add("*tearOff", preferences.get(PrefKey.TEAROFF_MENUS))
         self.rowconfigure(0, weight=1)
         self.columnconfigure(0, weight=1)
@@ -74,13 +95,18 @@ class Root(tk.Tk):
         root dialog creation and resizing. Only the first will actually
         do a save, because the flag will only be true on the first call."""
         if self.save_config:
-            # Bug in maximized geometry leads to size being full screen size,
-            # but top-left is non-maximized top left, i.e. mid-screen.
-            # So, if maximized, don't save geometry, just save "normal/zoomed" state.
-            # Then when de-maximize happens, you have the correct size AND top-left.
-            if self.state() != "zoomed":
+            zoomed = (
+                root().wm_attributes("-zoomed")
+                if is_x11()
+                else (self.state() == "zoomed")
+            )
+            state = RootWindowState.ZOOMED if zoomed else RootWindowState.NORMAL
+            if root().wm_attributes("-fullscreen"):
+                state = RootWindowState.FULLSCREEN
+            # Only save geometry if "normal". Then de-maximize should restore correct size and top-left.
+            if state == RootWindowState.NORMAL:
                 preferences.set(PrefKey.ROOT_GEOMETRY, self.geometry())
-            preferences.set(PrefKey.ROOT_GEOMETRY_STATE, self.state())
+            preferences.set(PrefKey.ROOT_GEOMETRY_STATE, state)
 
 
 def root() -> Root:
--- a/src/guiguts/root.py
+++ b/src/guiguts/root.py
@@ -38,15 +38,11 @@ class Root(tk.Tk):
         super().__init__(**kwargs)
         self.geometry(preferences.get(PrefKey.ROOT_GEOMETRY))
 
-        # Set zoomed/fullscreen state appropriately for platform
-        state = preferences.get(PrefKey.ROOT_GEOMETRY_STATE)
-        if state == RootWindowState.ZOOMED:
-            if is_x11():
-                root().wm_attributes("-zoomed", True)
-            else:
-                self.state("zoomed")
-        elif state == RootWindowState.FULLSCREEN:
-            self.wm_attributes("-fullscreen", True)
+        self.full_screen_var = tk.BooleanVar(
+            value=preferences.get(PrefKey.ROOT_GEOMETRY_STATE)
+            == RootWindowState.FULLSCREEN
+        )
+        self.allow_config_saves = False
 
         self.option_add("*tearOff", preferences.get(PrefKey.TEAROFF_MENUS))
         self.rowconfigure(0, weight=1)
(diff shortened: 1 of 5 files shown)


src/guiguts/root.py
```diff
@@ -73,15 +86,45 @@ def _save_config(self) -> None:
...
-        do a save, because the flag will only be true on the first call."""
-        if self.save_config:
-            # Bug in maximized geometry leads to size being full screen size,
-            # but top-left is non-maximized top left, i.e. mid-screen.
-            # So, if maximized, don't save geometry, just save "normal/zoomed" state.
-            # Then when de-maximize happens, you have the correct size AND top-left.
-            if self.state() != "zoomed":
+        do a save, because the flag will only be true on the first call.
+
+        Will do nothing until enabled via a call to set_zoom_fullscreen."""
```
srjfoo, line 91: I realize that the word `zoom`/`zoomed` was already there. I'm mostly commenting here because I think the word is being misused here. Zooming in, to me, means choosing a specific spot and doing the equivalent of looking at it via a magnifying glass. That is, whether it's an image or text, the area being zoomed in on is being enlarged.

However, if y'all think that it's an appropriate use case, I'll defer to you. I just wanted to raise awareness because we will eventually need to deal with image zoom.

srjfoo (approved):
@windymilla, if this works for you, is there anything else you wanted to do in #325 that needs testing. Since @tangledhelix just adapted your code for Macs, I started by reviewing this first.

I agree that if users use full screen on Macs, they're going to have to put up with the dialogs on top unless they want to migrate them to different desktops/screens.

I double-checked and confirmed that double-clicking in the title bar does maximize the window without fullscreening it.

All the different combinations I tried worked as expected.

src/guiguts/root.py
windymilla: I agree that "zoom" normally means something else. In this case though, the state reported by the window is actually the word "zoomed", so it makes sense to use that within the code.

windymilla:
> @windymilla, if this works for you, is there anything else you wanted to do in #325 that needs testing.

@tangledhelix or @srjfoo: please could you check what happens when you are fullscreen, and you click in the status bar on the first (row/col) button to jump to a line number, or on the Lang button to set the language. It just occurred to me that the dialogs used for that are tkinter simpledialogs, not our TopLevel dialog, so it's possible they may do the weird fullscreen/tab thing. There are also calls to messagebox, e.g. if you try to quit without saving after editing a file. I also wonder about the File Load/Save dialogs (filedialog), though I guess one of you would have noticed them if they weren't working.

If they go wrong, then I guess we will have to inherit our own versions from them, and add the withdraw, transient & deiconify hacks to the init method, Dan?

Thanks

windymilla:
> @windymilla if you would, please give it a try on Windows.

Thanks Dan - I didn't find any problems on Windows, but had the thoughts above about other types of dialog.

tangledhelix:
> @tangledhelix or @srjfoo: please could you check what happens when you are fullscreen, and you click in the status bar on the first (row/col) button to jump to a line number, or on the Lang button to set the language. It just occurred to me that the dialogs used for that are tkinter simpledialogs, not our TopLevel dialog, so it's possible they may do the weird fullscreen/tab thing. There are also calls to messagebox, e.g. if you try to quit without saving after editing a file. I also wonder about the File Load/Save dialogs (filedialog), though I guess one of you would have noticed them if they weren't working.

- Line number, Lang, etc bring up a floating dialog, no weird fullscreen behavior.
- File open dialog: looks normal
- "Save changes?" on closing: looks normal

All good!

srjfoo:
> > @tangledhelix or @srjfoo: please could you check what happens when you are fullscreen, and you click in the status bar on the first (row/col) button to jump to a line number, or on the Lang button to set the language. It just occurred to me that the dialogs used for that are tkinter simpledialogs, not our TopLevel dialog, so it's possible they may do the weird fullscreen/tab thing. There are also calls to messagebox, e.g. if you try to quit without saving after editing a file. I also wonder about the File Load/Save dialogs (filedialog), though I guess one of you would have noticed them if they weren't working.
> 
>     * Line number, Lang, etc bring up a floating dialog, no weird fullscreen behavior.
> 
>     * File open dialog: looks normal
> 
>     * "Save changes?" on closing: looks normal
> 
> 
> All good!

This may be an effect of the fact that I moved my macOS doc from the (default) bottom of the screen to the right (I think it was initially because at the time I was doing a lot of Photoshop work, and the doc shortened the available height enough to be annoying (where have I heard that before? 😁), and I've never been motivated to put it back at the bottom of the screen.

Anyway, if, in full-screen mode, I go to the bottom of the screen, the gg2 window disappears. No, it doesn't kill the window, it just puts GG2 into the background.

If I move to the second monitor and go full screen, this doesn't happen, but I need to double-check how some of those options are supposed to work, and I don't have time right now -- I'll get back to this later.

srjfoo:
Okay, finally had a chance to do some further testing. What I'm seeing on my main laptop screen (i.e. where the dock is) as far as the toolbar is concerned when in full-screen mode, if I move my mouse to the bottom of the screen, there's at least one point where even just hovering over the toolbar will cause the GG window to go into the background, as mentioned above. In addition to my default dock location (right, hidden), I tried
* bottom, visible, second monitor attached
* bottom, hidden, second monitor attached
* bottom (don't remember whether hidden or visible) with second monitor detached.

This does not happen if the window is maximized instead of fullscreened.

@tangledhelix, It may very well be some setting on my computer that I've forgotten setting, but the more that I play with this, the more convinced I am that Mac users are better off maximizing rather than fullscreening.

tangledhelix:
After more discussion and tinkering tonight, @srjfoo and I have figured this one out (at least, we figured out what's going on -- how to fix it is a problem for a different day).

Also, I'm coming to agree with @srjfoo that Mac should have fullscreen disabled if we can find a way, and preferably override the green button to maximize (not fullscreen) the window.

This is the explanation for why Guiguts "disappears":
- Going to fullscreen on Mac creates a new desktop in the virtual desktop system ("Mission Control").
- That desktop sits alongside all the other desktops, but from the user's POV unless they know how to look at the desktop overview, all they see is Guiguts consuming the entire screen. Many Mac users are familiar with Mission Control, but it's likely that many more aren't.
- Mousing down to the bottom of the screen caused Guiguts to "disappear". I was able to reproduce this.
- In reality, Guiguts isn't disappearing -- what happens is that the user is flipped back to the main desktop. The fullscreen app is still there, in its own desktop, but this happens quickly and from the user POV it's just gone away.
- If you go to the Mission Control view (or use any of the gestures or whatnot that let you navigate desktops...) you can go back and forth between Guiguts and the normal desktop.

Having reproduced the observed behavior and determining that Guiguts wasn't actually closing the window or similar, we then looked into more detail about the trigger that causes the behavior. @srjfoo had theorized it was caused by moving the mouse pointer to the bottom, thus triggering the Dock to un-hide. However, further experiments showed that the behavior was not triggered by going to the bottom of the screen. Moving very very slowly, it turned out that it happens when you get about halfway through the GG status bar. And that's when I realized - it's the tooltips. Since they are actually just very small windows, triggering their display seems to cause the app to navigate back to the main desktop for some reason.

It seems like managing all of these things against Mac's fullscreen approach is a tar pit. So I'm going to see if I can find a way to override the green button. Either to cause it to only maximize the window - or to disable it somehow. I'm not sure whether it's possible. It could be our best option will be something like a dialog pop if you go fullscreen that warns you "things will be weird", and adding a note in the documentation recommending against fullscreen mode.

windymilla:
> And that's when I realized - it's the tooltips. Since they are actually just very small windows, triggering their display seems to cause the app to navigate back to the main desktop for some reason.

Tooltips inherit from the TopLevel class as well, so theoretically shouldn't cause a problem. Two ideas for edits to try in widgets.py:
 1. Comment out Line 388 (sets geometry to -1000, -1000) - don't think it will be that, because that is during creation (which would be when the status bar button is created, not when the mouse moves in). The tooltip isn't actually displayed until later in `_show`.
 2. In `_show`, Line 419, change `if self.use_pointer_pos:` to `if True:` to see whether it's to do with things like `winfo_rootx` returning weird values for fullscreen/tabbed things.
 3. In `_show` line 417, just abandon the "showing" of the tooltip if currently fullscreen. Something like adding
 `or is_mac() and preferences.get(PrefKey.ROOT_GEOMETRY_STATE) == RootWindowState.FULLSCREEN`
like you added previously in the `transient` edit.

The documentation can say "you won't get tooltips on a Mac if in fullscreen", assuming neither 1. nor 2. above works, or we can't get a good solution.

Separate idea: since the green gumdrop method of going fullscreen should still call root's `_save_config` method, could you intercept it at that point, and have code that says `if is_mac() and root().wm_attributes("-fullscreen"): turn off fullscreen and do zoomed instead`? Combined with removing the View menu Fullscreen option on Macs, that would stop people working in full screen mode.

I'll leave the decision to you both to make the final call on what we do for Mac users: should they have the View menu Fullscreen option; should we detect attempts to go fullscreen and subvert them to make them zoomed instead; should we put more development effort into trying to make fullscreen work perfectly on a Mac?

windymilla:
I wasn't sure how to get your changes back into my original PR, so in order to show you my suggested edit, I'll commit a small change to this PR - hope that's OK with you, @tangledhelix 

I've made it remove the "full screen" checkbox from the View Menu on Macs. I've also added to the function that gets called when the configuration of the main window is changed - if it detects that you have used the green gumdrop to go "full screen", it reverts immediately back to non-full screen.

Perhaps you and/or @srjfoo wouldn't mind testing it on Macs, and saying if you think it is an acceptable solution (at least for now). Thanks

If it's no good, we can always revert the commit, and try something else.

windymilla:
Also rebased, since I think black (at least) is a later version than when this PR was first created

srjfoo:
> I've made it remove the "full screen" checkbox from the View Menu on Macs. I've also added to the function that gets called when the configuration of the main window is changed - if it detects that you have used the green gumdrop to go "full screen", it reverts immediately back to non-full screen.

Is there a way that clicking on the green gumdrop can be made to either "do nothing", or "maximize" (as opposed to full-screen)? Mind you, it probably doesn't matter a lot from my perspective, because I'll rarely be trying to go full screen, anyway, but I found that the window going from quarter-screen to full-screen, only to return to quarter-screen to be a bit disconcerting visually. However, if @tangledhelix is okay with it, I'll defer to what y'all will think best.

windymilla:
OK - worked out how to get @tangledhelix's fixes into my branch (I knew but didn't realize I knew :) - cherry pick commits in Github Desktop)
So, I've reverted this PR to its previous state without my attempted fix in. I'll continue working in my PR (with Dan's commit included).

windymilla:
Now that #325 is merged, including @tangledhelix's fixes, I will close this shortly, unless anyone objects.

windymilla:
Handled by #325

PR #330, opened 2024-07-20
Add page label info to page marker flags

windymilla:
Add the info necessary to construct the labels when in GG1. Format is `[ImgNNN|Arabic|+1]`
"Arabic" can be "Roman"
"+1" can be "No Count" or a page number

This version must be used in conjunction with a version of GG1 that has the added info - it won't work with older versions of GG1, like 1.6.2.

[GG1 PR#1308](https://github.com/DistributedProofreaders/guiguts/pull/1308)

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -15,7 +15,6 @@ from guiguts.maintext import (
     PAGE_FLAG_TAG,
     PAGEMARK_PIN,
     BOOKMARK_TAG,
-    PAGEMARK_PREFIX,
     img_from_page_mark,
     page_mark_from_img,
 )
@@ -23,6 +22,7 @@ from guiguts.page_details import (
     PageDetail,
     PageDetails,
     PAGE_LABEL_PREFIX,
+    NUMBER_INCREMENT,
     STYLE_ARABIC,
     STYLE_DITTO,
 )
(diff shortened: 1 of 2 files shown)


windymilla:
Two advantages of this version:
1. You can delete the page flags as soon as you load into GG1 and work on it as much as you like, then add them back just before saving to transfer back to using GG2.
2. The page numbering info is all transferred, not just the page break location.

srjfoo:
Discovered during cross-testing that GG1 and GG2 do not join words split across page boundaries the same. GG1 places the page marker at the end of the word that was split. GG2 places the page marker where the split occurred. My understanding is that GG1's behavior is the expected behavior.

One other difference that I don't know whether it matters: when I add page marker flags, they're added at the end of the line in GG1, and at the beginning of the line in GG2.

windymilla:
> Discovered during cross-testing that GG1 and GG2 do not join words split across page boundaries the same. GG1 places the page marker at the end of the word that was split. GG2 places the page marker where the split occurred. My understanding is that GG1's behavior is the expected behavior.

Added issue #331 to tackle this - thanks!

windymilla:
> One other difference that I don't know whether it matters: when I add page marker flags, they're added at the end of the line in GG1, and at the beginning of the line in GG2.

I believe this is caused by GG1 moving page marks to end of line during Page Sep Fixup, rather than any fault in GG1 or GG2 Page Marker Flag addition. No action necessary IMHO.

srjfoo (approved):

PR #333, opened 2024-07-21
Stop Page Sep Fixup leaving page breaks mid-word

windymilla:
Instead of joining the lines and then insert a newline after the end of the joined word, cut the second part of the word and paste it on the end of the first part. This preserves the page break at the start of the
second line.


srjfoo:
On this one, I've been stepping through a separator at a time, so I actually spotted this. I expect that it's this way in #330 also, so I don't know if you want to fix it there, or punt it (or tell me, "No, no, Sharon, you did it all wrong!").

In this situation, where there are two consecutive page separators because the only thing on that page was an illustration mid-para that had already been moved:

image

This is what happens:

image

windymilla:
> where there are two consecutive page separators...

I've added [an issue](https://github.com/DistributedProofreaders/guiguts-py/issues/336) for this as it's not related - GG1 has a related bug, although it (just) copes with this exact situation. I'm not going to change GG1, but I will make GG2 cope with consecutive page seps.

If you don't have any other concerns, this can be approved, and I will tackle the other issue separately - thanks.

srjfoo (approved):

PR #334, opened 2024-07-22
Restrict width of file path in title bar

windymilla:
If file path is too long, truncate it to the left, leaving the filename still visible.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -103,6 +103,7 @@ class Guiguts:
 
         maintext().focus_set()
         maintext().add_modified_callback(self.update_title)
+        maintext().add_config_callback(self.update_title)
 
         # Known tkinter issue - must call this before any dialogs can get created,
         # or focus will not return to maintext on Windows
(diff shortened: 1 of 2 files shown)


windymilla:
Does this look useful?

srjfoo (approved):

srjfoo:
> Does this look useful?

I went ahead and approved it, but won't have my feelings hurt if you opt not to merge it.

From a testing standpoint, I leave my test project files named with the projectid, so the name of the folder that contains the text file is 22 characters long, the file itself in its raw state is 26 characters long. More if I add suffixes to identify different things that are being tested. Add other outer folder or two (in addition to the home path), and yes, you exceed the line length of the main window pretty quickly.

This may not be as much of an issue for active PPers, if they rename the text file to something significantly shorter, they may have no trouble seeing the actual filename, even with multiple folder labels.

windymilla:
It seems pretty harmless, so I'll merge it. It would be easy to revert later if we wanted to for any reason.

PR #341, opened 2024-07-27
Improve scan directory handling

### LINKED ISSUE #328: Can we have the internal viewer remember the image file directory?
When using the Auto Img internal viewer, I must select a directory whenever I open a file. In contrast, GG1 automatically opens file directories ending in "...images" from the same folder as the text file I'm editing. Additionally, GG1 only prompts me once for clarification if there's no "...images" directory.

windymilla:
1. Use "Folder" on macOS, not "Directory" in menu buttons
2. Use `projectID0123456789abc_comments.html` file to extract project ID, and save it in the project json (bin) file.
3. When user turns on Auto Img or tries to load an image for the first time for that project, first look if there is a `pngs` subdir in the project dir. If not, is there a subdir `projectID0123456789abc_images`?

Fixes #328

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -55,7 +55,7 @@ from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
 from guiguts.tools.pptxt import pptxt
 from guiguts.tools.jeebies import jeebies_check, JeebiesParanoiaLevel
-from guiguts.utilities import is_mac, is_windows
+from guiguts.utilities import is_mac, folder_dir_str
 from guiguts.widgets import themed_style, theme_name_internal_from_user
 from guiguts.word_frequency import word_frequency, WFDisplayType, WFSortType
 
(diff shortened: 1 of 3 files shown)


srjfoo:
"Directory" changed to "Folder" for the one instance I saw, so that looks good, but I'm still getting a prompt for the location of the proofing images if the pngs folder is not in the same folder with the text file.

I'm happy to put in some diagnostic prints, but I need to know what to use and where to put it.

srjfoo:
Followup. I started with a clean GGprefs directory, to see what happened with a newly-created `json` file. And got the same results -- prompt for where to look for the pngs. The only things in the file are

```json
{
  "text_font_family": "DP Sans Mono",
  "root_geometry": "1148x764+5+43",
  "auto_image": true,
  "recent_files": [
    "/Users/sharonjoiner/Desktop/guiguts-testing/_projects_for_testing/blunderland2/projectID620c5a7c98ea1/projectID620c5a7c98ea1.txt"
  ]
}
```

So, in the brand new `json` file, without saving the text file, the path to the text file is added to the json as a recent file, but I don't see anything about the projectid, except for as mentioned in the file I used (which could easily have been changed to something shorter). Mind you, I don't know how to read the code well enough to know if it's supposed to be, but without something in the log file, I don't know how to figure out what the problem is. And if it's working okay on Windows, it's probably some weird difference between the systems.

srjfoo (approved):
After a discussion in Slack with @windymilla on just what was supposed to happen, it turned out that I was misunderstanding what was supposed to be happening.

PR #348, opened 2024-07-29
Improve placing of GGprefs on Windows

windymilla:
User's "documents" folder is not necessarily `~/Documents`. This may be because the user has set its location, but is most likely because Windows has sneakily placed the documents on OneDrive and the user has given up
battling against it.

So, interrogate Windows to find the actual location of the Documents folder (which may not even be called `Documents`) and use that.

On all platforms, if the GGprefs folder doesn't exist, create it, and also create any parent folders that are necessary, e.g. if user's documents folder is
`~/Documents` but that folder doesn't exist, then
create `Documents` first, then the GGPrefs folder
within that. Should never happen, but just in case! Also, if folders can't be created, or the Prefs file can't be saved, output an error, rather than crashing with an exception.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -707,13 +707,13 @@ files = [
 
 [[package]]
 name = "pytest"
-version = "8.3.1"
+version = "8.3.2"
 description = "pytest: simple powerful testing with Python"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "pytest-8.3.1-py3-none-any.whl", hash = "sha256:e9600ccf4f563976e2c99fa02c7624ab938296551f280835ee6516df8bc4ae8c"},
-    {file = "pytest-8.3.1.tar.gz", hash = "sha256:7e8e5c5abd6e93cb1cc151f23e57adc31fcf8cfd2a3ff2da63e23f732de35db6"},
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: 
1. On non-Windows platforms, the only change should be that if the folder `~/Documents` (macOS) or `~` (Linux) doesn't exist, then it will be created first, then the `GGPrefs` (macOS) or `.ggpreferences` (Linux) will be created in it.
2. I've tested the following on Windows: I configured Windows to treat `~/docs2` as my Documents folder (scary - I didn't want to lose everything in my Documents folder), and checked that GG still ran successfully, creating a new GGPrefs folder in my new Documents folder. 
3. Once this is merged and a new alpha release made, then any user for whom Windows treats a OneDrive folder as their Documents folder, will find that GG2's GGprefs will begin to be stored in that OneDrive folder. Although this may be annoying, that's the way Windows works, and if you don't want it, you have to configure Windows to use a local Documents folder instead. This seems preferable to forcing the user to have a local Documents folder under their home folder, especially because Windows handles those "known" folders in a strange way. For example, in my testing of point 2 above, I ended up with two folders in my home folder, both of which were called `Documents` (although one of them was actually `~/docs2`) and with the special Documents icon, and I could only distinguish them by looking at their creation date, or by double clicking them and viewing the contents.

srjfoo (approved):
LGTM -- I assume from my standpoint, all I needed to do was confirm that if I delete the macOS `~/Documents/GGprefs/` folder, it's still created.

PR #350, opened 2024-07-29
Make highlight in checker dialogs clearer

### LINKED ISSUE #349: Issues with Tools -> Unmatched -> Block Markup
1. "/F" is currently flagged as unmatched block markup even when it appears mid-line, as in the following example:

`Badly formed markup: say, there it is printed in Roman, i. e.[**without space?/F1] in`

It might be helpful to limit error highlighting to instances where "/F" occurs at the beginning of a line or within a few characters of a new line.

2. The current green highlighting for "/F" in error messages is difficult to discern. I recommend using red highlighting for better visibility.

windymilla:
Use the colors of selected text in the main window.

Fixes #349 (the problem raised in point 1 in that issue is already handled by the tool).

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -224,7 +224,11 @@ class CheckerDialog(ToplevelDialog):
         self.process_command = process_command
         self.rowcol_key = sort_key_rowcol or CheckerDialog.sort_key_rowcol
         self.alpha_key = sort_key_alpha or CheckerDialog.sort_key_alpha
-        self.text.tag_configure(HILITE_TAG_NAME, foreground="#2197ff")
+        self.text.tag_configure(
+            HILITE_TAG_NAME,
+            background=maintext()["selectbackground"],
+            foreground=maintext()["selectforeground"],
+        )
         self.text.tag_configure(ERROR_PREFIX_TAG_NAME, foreground="red")
         self.count_linked_entries = 0  # Not the same as len(self.entries)
         self.count_suspects = 0


srjfoo (approved):
Observation -- since WF doesn't belong to this class, the list of results isn't highlighted. It really isn't necessary, since it's just a list of "words" as defined by WF.

However, the spelling checker *does* have the individual words highlighted, and it too is a list consisting of one word on each line, not one word or phrase that needs to be highlighted within a given line.

I've gone ahead and approved the PR, as it does what's on the label, but thought I should raise the issue of whether in the cases where the results in the widget are a single word of phrase, all of which is considered a search term, it needs to be highlighted (i.e. there's nothing else on the line in the search results). Or should WF match what the widgets do?

Screenshot comparing WF results and Spelling checker results:

image

windymilla:
Good question. Although they are superficially the same, in that they have a list of words, I think it's OK for them to be different for the following reasons:
1. Behavior is different - in "checker" dialogs (spell, jeebies, etc.), you click to go to the line number the message refers to, Ctrl/Cmd click to "fix", right-click to remove, etc. In WF, you click multiple times to cycle through the places where that word occurs, and Cmd/Ctrl click to pop the S/R dialog.
2. The look is different - "checker" dialogs begin with the line/col numbers, then the message". The spell checker doesn't just have the word, it has
`25.6 badspeeling (3)`
so the word is "hidden" between two other bits of text.
In WF, the words are all aligned in a column with nothing after them.
3. I definitely don't think that Spell check should omit the highlighting (since the highlighting & behavior are consistent across "checkers"), and I think highlighting all the vertically aligned words in WF would look odd and unnecessary.

PR #351, opened 2024-07-30
Don't add newline when rewrapping

### LINKED ISSUE #347: Tools->Rewrap All and Tools->Rewrap Selection add a blank line after every rewrapped paragraph
Selected this section:

```
The articles on Church architecture have been carefully revised by the Rev.
G. A. Poole, M. A., vicar of Welford.
 
The Law articles have been revised, partly by the Rev. James Brogden,
A. M., of Trinity College, Cambridge, and partly by William Johnston, of
Gray's Inn, Esq., barrister-at-law.
```

Tools->Rewrap Selection or Tools->Rewrap All result:

```
The articles on Church architecture have been carefully revised by the
Rev. G. A. Poole, M. A., vicar of Welford.


The Law articles have been revised, partly by the Rev. James Brogden,
A. M., of Trinity College, Cambridge, and partly by William Johnston,
of Gray's Inn, Esq., barrister-at-law.
```

GG1 does not add extra blank lines when rewrapping entire paragraphs.

windymilla:
If the paragraph input to the text wrapper didn't have a newline, then the output wrapped paragraph shouldn't have one either.

Fixes #347

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -759,6 +759,7 @@ class File:
         Args:
             section_range: Range of text to be wrapped.
         """
+        maintext().selection_ranges_store_with_marks()
         maintext().undo_block_begin()
 
         # Dummy insert & delete, so that if user undoes the wrap, the insert
@@ -772,6 +773,7 @@ class File:
         maintext().rewrap_section(
             section_range, lambda: self.unpin_page_marks(mark_list)
         )
+        maintext().selection_ranges_restore_from_marks()
 
     def pin_page_marks(self) -> list[str]:
         """Pin the page marks to locations in the text, by inserting a special
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -36,6 +36,8 @@ PAGEMARK_PIN = "\x7f"  # Temp char to pin page mark locations
 BOOKMARK_TAG = "Bookmark"
 PAGEMARK_PREFIX = "Pg"
 REPLACE_END_MARK = "ReplaceEnd"
+SELECTION_MARK_START = "SelectionMarkStart"
+SELECTION_MARK_END = "SelectionMarkEnd"
 
 
 class FindMatch:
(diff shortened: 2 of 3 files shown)


windymilla:
Testing notes: See Rick's issue #347 for example text.

Sorry for letting this slip through - Quentin had coded the Knuth Plass algorithm correctly, but the way it meshed with my previous code that used the python wrapper meant that a newline was added to every paragraph.

srjfoo:
Two paragraphs wasn't quite enough for the testing I did, so I duplicated Rick's test paragraphs as shown in issue #347. 

I noticed when I was testing just the two paragraphs that, even though the selected text/all the text in the file did not have a trailing newline, a trailing newline was added, so part of what I was testing was whether, if I selected a couple of paragraphs in the middle, carefully omitting the newline, would a newline be added. The answer is no, it isn't. A line is only added at the end of the file, and then only if there isn't one already.

However, post-rewrap-selection, the selected area changes.

* For one paragraph, the selection is cleared completely
* For two paragraphs, the blank line in between the paragraphs is selected
* For three or more paragraphs, the first and last paragraphs are deselected, and everything in between remains selected.
* For a multi-paragraph selection that includes the leading and trailing blank lines, everything remains selected
* For a multi-paragraph (and probably a single para that included the end-of-file) selection that includes the end of file and starts with a blank line, the selection remains intact.
* For a multi-paragraph selection that starts at the beginning of the paragraph and includes end of file where there is not a return at the end of the file, 
* * The beginning and ending paragraphs are removed from the selection
* * A blank line is added to the end of the file
* * The cursor is located on the blank line that was added.

The screenshot below shows an example of the last bullet. I changed the source text slightly (spelling the abbreviation "Rev." out so that the difference between unwrapped and wrapped text would be more clearly visible. It shows that the rewrapped text starts with the 4th paragraph, but the 4th and 8th paragraphs are no longer selected. The cursor is on line 28 (which did not exist before the rewrap).

I'm guessing that the addition of a blank line at the end of the file is not a problem. Nothing I did added an extra blank line between paragraphs, so that's good. The reason I was so long-winded here is that I wanted to make sure that you're aware of the weirdness associated with post-rewrap selection area.

image

windymilla:
> A line is only added at the end of the file, and then only if there isn't one already.

That is what I would expect. Although it's possible to have a file without a newline at the end of the last line, it appears as though Tk (and some other programs) prefer that. I think it's harmless, and maybe beneficial to have a final newline.

> the weirdness associated with post-rewrap selection area.

Thanks - I'll have a look at that.

windymilla:
OK @srjfoo - that should fix the weirdness relating to post-rewrap selection areas.

windymilla:
@G4OEU - to download the new commit I just added to this PR so it appears in your local branch using Github Desktop, you should be able to use the "fetch" button next to the branch selection button, or Repository-->Fetch.

G4OEU:
When I tested the rewrite to Python of the Knuth-Plass rewrapping algorithm, I compared the wrapped output against the same text wrapped in GG1 (by the Perl version of that algorithm). The tests produced identical results on the paragraphs I was using.

I have just now compared the wrapping of a 8,600 line book file by GG1 and GG2. The input file has all the usual markup you would expect at the point you do the first rewrap. Almost all paragraphs wrap identically which bears out my experience when testing. However in the case of GG2, it has managed to fit an extra word onto some lines so some wrapped paragraphs show differences.

A consequence is that some of those paragraphs rewrap more neatly in GG2 while others are perhaps less neat than the GG1 wrapped version of the paragraph; that is to be expected and my judgement is subjective. Another consequence is that the length of the rewrapped file is eight lines shorter than its GG1 comparator.

I have used the default rewrap margins as displayed in the Preferences menu of each version. As far as I can see I am using the same values in each version of GG although their description may differ in the GG2 Preferences menu.

I recall a comment by Nigel some time ago about interpreting margins when implementing them in GG2. I cannot remember if that meant there are some differences in this regard between GG1 and GG2 that might account for what I see. More likely it is a subtle difference in the behaviour of the new Python Knuth-Plass algorithm.

In both GG1 and GG2, the wrapped output is within the margins specified in the preferences file. That is, if 'Right Margin' is set to 72 then no line wraps to a length greater than 72 characters.

srjfoo (approved):
Looks good to me

G4OEU (approved):
Running the changes shows that they are doing what is expected. Looks good.

G4OEU:
I said in a previous post that "[i]n both GG1 and GG2, the wrapped output is within the margins specified in the Preferences file. That is, if 'Right Margin' is set to 72 then no line wraps to a length greater than 72 characters."

Since both GG1 and GG2 will in some cases wrap a line to the full width allowed and display a line of 72 characters (from col. 0 to col. 71 inclusive), I think that implies that my Python implementation of the Knuth-Plass Python wrapper is responsible for the differences I'm seeing, NOT the way margins are interpreted/used in GG2.

It's not obvious though whether the Perl wrapper is correct and the Python wrapper incorrect or vice versa!

It would be nice if I could explore that question by wrapping the same text with a third tool that also uses Knuth-Plass and can wrap to specified margins. Any suggestions?

windymilla:
> It would be nice if I could explore that question by wrapping the same text with a third tool that also uses Knuth-Plass and can wrap to specified margins. Any suggestions?

In my opinion, @G4OEU, that is only of academic interest, so by all means investigate if it interests you, but I don't think it's necessary for the development of GG2. You've created a wrapper that uses Knuth-Plass, and the GG1 & GG2 versions are obviously almost identical. Even if they aren't, it isn't clear which it better, so there wouldn't be any benefit in adjusting GG2 to match GG1 exactly in all cases, even if you managed to work out which was "correct".

PR #354, opened 2024-07-30
Add `--nohome` argument

windymilla:
When `--nohome` is specified, prefs are not loaded from the prefs file, and also not saved if preferences are changed.


srjfoo (approved):

srjfoo:
Thanks -- there may not be a lot of times that we need this for testing, but there've been a few.

Issue #353, opened 2024-07-30
Improve accessibility, particularly relating to color


windymilla:
At the moment, things like the key bit of text in a checker dialog error message are shown highlighted in the same foreground and background colors as the OS scheme "selected text" (e.g. Windows default is white text on blue background).

Some may not find colors distinguish the text sufficiently.

Therefore:
1. Consider how many of the colors used need to be user configurable
2. Consider the optional use of underline, borders, bold, etc., in place of, or in addition to the colors.

srjfoo:
One comment I've seen over and over again about accessibility is that we shouldn't be depending just on color.

windymilla:
Additional note re colors: At the moment, something like highlighting ("spotlighting" in the code to distinguish from other highlights) is set to have an orange background, and either black or white text depending on if the theme is light or dark. However, in the "default" theme case, the current code assumes it's "light" and sets the text color to be black on orange. If, as is possible on a Mac, the default theme follows the system theme, the default may actually be dark with white text. The highlighted text is then black on orange when it should really be white on orange.

One solution would be to leave the Text foreground color unchanged, so it matches the rest of the text, but with an orange background. However, suppose your Mac theme had orange text - it would then be unreadable against an orange background. So, it appears as though it is necessary to set the fg as well as the bg color.
Another solution, given that we have chosen bg colors where either white or black text would work, would be to inspect the main text fg color (which will usually be black/close to black/white/close to white) and force the fg color for highlighting to be either black or white depending on which is closer the main text fg color. Then on a light default theme (with dark text) the highlight text would be black, and on a dark default theme, the highlight text would be white. This would be perfect for most themes, and would still be acceptable for unusual themes, e.g. with orange text.

All of the above may be irrelevant depending on if/how/how many colors we make user-configurable.

srjfoo:
A couple of comments:

My impression (and I don't remember where I read/heard this) is that if you're setting a background color, you should always set the text color, also. Mind you, this is mainly as applied to HTML/CSS, but I see a parallel here.

I think, as long as you set both bg and fg colors to be readable, it doesn't really matter what theme the user has set, it's still readable. I certainly don't object to having black text on a lighter highlight/spotlight if the spotlight needs to be light.

srjfoo:
I know we're not quite ready for beta releases yet, but I would not delay the first beta release for this, even if it is tagged as a core feature. I think that ultimately, we need input from the community, especially those who have accessibility needs.

[...]

PR #355, opened 2024-07-31
Spell check

windymilla:
1. If user has selected text, spell check will just check the selected words (or part words if word is not wholly within selection). Note that GG1 Spell **Query** does not have this feature, but Spell **Check** does.

2. If word appears more times in the file than the threshold value, it is assumed to be a good spelling and so is not reported - this is the same as GG1's Spell **Query** but Spell **Check** does not have this.

.

### DIFF
--- a/src/guiguts/highlight.py
+++ b/src/guiguts/highlight.py
@@ -1,15 +1,16 @@
-"""Highlight functionality"""
+"""Highlight functionality."""
 
 from enum import auto
 
 from guiguts.maintext import maintext
 from guiguts.preferences import preferences, PrefKey
+from guiguts.utilities import IndexRange
 
 
 class Highlight:
-    """Global highlight settings"""
+    """Global highlight settings."""
 
-    TAG_QUOTEMARK = auto()
+    TAG_QUOTEMARK = str(auto())
 
     # Possible future enhancement:
     #
@@ -25,6 +26,14 @@ class Highlight:
         "Light": {"bg": "#a08dfc", "fg": "black"},
         "Dark": {"bg": "#a08dfc", "fg": "white"},
         "Default": {"bg": "#a08dfc", "fg": "black"},
+    }
+
+    TAG_SPOTLIGHT = str(auto())
+
+    COLORS_SPOTLIGHT = {
+        "Light": {"bg": "orange", "fg": "black"},
+        "Dark": {"bg": "orange", "fg": "white"},
+        "Default": {"bg": "orange", "fg": "black"},
     }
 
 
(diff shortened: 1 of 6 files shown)


windymilla:
Question for @srjfoo:

If you run the spell checker in GG2, it lists the spelling errors, and chooses the first error (or the user can choose any error) which then selects that word in the main text window. If you decide you want to adjust the threshold and re-run, then now it only rechecks that one word (because it is selected and we have just made spell check only check the selection if there is one. 
That seems a bit annoying (at least it was to me) because it's almost certainly not what you intended. If you had previously selected a region to spell check, then you probably want to re-check that region. If you previously spell checked the whole file, you probably want to re-check the whole file.

Any good ideas of how to make it more helpful?

My best idea so far (which would need to wait until #351 is merged, with its "restore selection" feature) - probably more accurate to describe the comments below as a brain dump / random thoughts:
Save the selection (if any) immediately before doing the spell check. When using "re-run" do a "restore selection" immediately before re-running. The re-run would then work on the same section of text as the initial run. 
Disadvantage: Consider this case - user does spell check on whole file, then realizes they meant to only spell check a selection, so they make a selection and re-run. The reverse case (accidentally only spellchecking a selection, then attempting to re-run with no selection to spell check everything) would also go wrong. With the above idea, re-running would override their selection/de-selection.
Fix? If user makes a selection (including selecting nothing), then clear the #351 saved selection

windymilla:
@srjfoo - I've pushed another commit that uses the "alternative selection"/"highlighting" idea instead of the main selection mechanism when highlighting a word due to a click in a checker tool (also in WF). The color (orange) is the same as GG1.
This fixes the "spell check a selection, then re-run" problem.

Question: Would it be better if the selected words in the dialog were also in orange? I tried it, and having lots of orange was a bit overwhelming, so I've currently left them in the default selection color. Here's a screen shot of all orange compared with leaving it blue.

image

image

srjfoo:
For what it's worth, this is what I see, (I assume) because my selection is still active. So, somewhat different from what you see. However, I agree with you. I'd rather the highlighting in the spellcheck results stay the user's system default. (believe it or not, I'm not really all that fond of pink 😁)

image

---

I don't know if you're seeing this on Windows, but I am on macOS: When re-running spell check (whether for a changed threshold or a different selection), the count in the upper-left-hand corner is set to "0 Entries" between the old results display and the new. I doubt that anyone (after the first time) will misinterpret it, because the spell check does finish and you have a final count.

My first thought was to leave that area blank while the script is working, but it probably makes more sense to either display something like a `...` or the `working` message that's displayed in the main window. My thinking is that the PPer is probably going to be focused on the spell check widget, not on the bottom right-hand corner of the main text window. (You are, of course, welcome to ignore both suggestions -- or punt on the issue. 😁)

srjfoo (approved):

src/guiguts/highlight.py
```diff
@@ -27,6 +28,14 @@ class Highlight:
         "Default": {"bg": "#a08dfc", "fg": "black"},
     }
 
+    TAG_SPOTLIGHT = str(auto())
+
+    COLORS_SPOTLIGHT = {
+        "Light": {"bg": "orange", "fg": "black"},
+        "Dark": {"bg": "orange", "fg": "white"},
+        "Default": {"bg": "orange", "fg": "black"},
```
srjfoo, line 36: Aaah, this is why I see black text on orange in the main text window. :D 

Black looks fine on orange, which is a good reason to choose a color that black and white are equally readable on, because otherwise, fancy footwork would be needed with the default theme.

windymilla: See additional color notes in #353 relating to "fancy footwork" for default theme.

windymilla:
> the count in the upper-left-hand corner is set to "0 Entries" between the old results display and the new

I've added #364 to tackle this

PR #360, opened 2024-08-02
Improve user feedback after fraction convert

### LINKED ISSUE #358: Tools->Convert Fractions does not give any indication that anything happened
In GG1, using any of the 'Tools -> Convert Fractions' options positioned the cursor at the last converted fraction. If no fractions were found, the cursor remained unchanged.

GG2 currently doesn't move the cursor regardless of whether fractions were converted or not. Providing some visual feedback to confirm the action's completion would be beneficial.

Low priority.  Or alternatively, mention in the manual that Ctrl-z could be used to confirm something happened.

windymilla:
If fraction convert is done on a selection, make sure the selection is correctly preserved, using the new feature added by PR #351.

Also position insert cursor directly after last fraction converted.

Fixes #358

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -1037,11 +1037,13 @@ def fraction_convert(conversion_type: FractionConvertType) -> None:
     subscripts = str.maketrans("0123456789", "₀₁₂₃₄₅₆₇₈₉")
 
     maintext().undo_block_begin()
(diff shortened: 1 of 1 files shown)


windymilla:
> Is that the intended result?

Neither intended nor unintended as I hadn't thought of it. I agree it's better to position after last converted fraction. New commit should do that.

srjfoo (approved):
👍

Issue #177, opened 2024-03-23
Consider options for interplay between cursor position and selection


windymilla:
In the review of PR #176, the subject of where the insert cursor is when a selection is made. In GG1 (& currently GG2), it is either at the start or end of the selection, depending on how the selection is made. In some apps, e.g. Notepad++, VS Code on Windows, this is also the case. However, in some other apps, e.g. Chrome address bar on Windows, and most/all Mac apps, the insert cursor is not visible when some text is selected.

Things to consider for GG2 (which would make it different to GG1, but maybe an improvement):

1. Do we want the cursor to show if there's a selection? Is it even possible to hide it in a Text Widget?
2. If the user presses left/right arrow when there is a selection, in all the apps we tested the selection was cancelled and the insert cursor moved to the start/end of the previously selected text. This was true, even in apps like Notepad++, so the cursor might be flashing at the end of the selected text, but when left arrow is pressed, it jumps to the start. Do we want that behavior in GG2?
3. The status bar shows the "current" character's Unicode ordinal and name. In GG1 and GG2 at the moment, "current" means the one immediately after the insert cursor. If the user selects a single character, then the name shown will depend on if the selection was done left-to-right or right-to-left, because it depends on the cursor not the selection. Could this be confusing? Do we want to display the name of the single selected character if there is one? Also, what if more than one character is selected? (Edited to add: suggestion is to display nothing)
4. Similar questions to 3 relate to the reported row.column numbers in the status bar. They currently depend on the cursor position, and are unaffected by selection. Is that desirable.

Edited to add:
5. Note that column selection is particularly unexpected in that it leaves the insertion cursor where it was before selection, which may not even be within the selection.

tangledhelix:
I had so much to say on #176, I feel I should say something here as well 😃 

Item 1 - I don't feel it's necessary to hide the cursor. Taking VS Code as an example implementation, it does not hide the cursor in selection mode, rather behaving similar to GG2.

Item 2 - I will note VS Code again here - actually as a potential outlier? It seems to behave the same as GG2 in this situation. Unlike what you wrote, it will de-select the previously selected text, but instead of jumping to the beginning or end of the prior selection, it will move the cursor in the same way GG2 is doing today. Which behavior to mimic is probably a matter of opinion, but I would be inclined to leave it as-is since GG1 users are accustomed to that (and it's zero work to implement 😄 )

Item 3 - I said my piece in #176 so will just refer readers there...

[...]

PR #366, opened 2024-08-04
Highlight line number of the current line

windymilla:
Also addresses point 5 of issue #177, i.e. the insert cursor is left in a sensible location after column select.


windymilla:
Should update correctly when a key is pressed, or a mouse button clicked, including column select.
Note that previously the insert cursor location wasn't updated after a column select, but now it is.

srjfoo (approved):

PR #370, opened 2024-08-05
Cursor position, selection and status bar

windymilla:
1. When there is a selection, typical behavior from other apps if the user presses the left/right arrows is to deselect and position the cursor at the start/end of the selected range.
2. If exactly one character is selected, show that in the character name field in the status bar. If nothing selected, show the character after the cursor. Otherwise show nothing.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -750,8 +750,18 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             preferences.set(PrefKey.ORDINAL_NAMES, force)
 
         def ordinal_str() -> str:
-            """Format ordinal of char at current insert index for statusbar."""
-            char = maintext().get(maintext().get_insert_index().index())
+            """Format ordinal of single selected char, or char at current insert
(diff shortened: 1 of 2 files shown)


windymilla:
I think this handles everything we can do from #177

srjfoo:
Right and left arrows work as expected (✅ )
No character shown in the status bar if multiple characters selected (✅ )

**_However_**, when the arrow key is pressed to cancel the selection, the line with the cursor jumps to the middle of the page, instead of staying where it is. Is there something that can be done about that? Ideally, the cursor shouldn't move from where it is in this case, I think?

If anyone comments about the behavior of the up and down arrows, maybe document that it depends on the direction that the selection was made in, and has been determined to not have a standard among word processors/text editors.

windymilla:
Now doesn't scroll window to show cursor, unless cursor is off-screen, e.g. if you select a big chunk of text, then scroll down so the top of the selection is off-screen, then press left-arrow.

srjfoo (approved):

Issue #373, opened 2024-08-07
Tools->Word Frequency Control-left click Search & Replace highlighting


okrick:
When using Hyphens mode, clicking a word or phrase with Ctrl+Left Click opens the "Search & Replace" dialog. The selected word appears highlighted in the "Search" field, mimicking selection. However, pressing Ctrl+C doesn't copy the term. You need to re-select it manually before pasting it into the "Replace" field.

Ideally, the highlighted term should be automatically selected, allowing immediate copying with Ctrl+C. If this can't be achieved, perhaps removing the highlight altogether when opening the dialog via Ctrl+Left Click in Word Frequency would be a better solution.

Interestingly, when a word or phrase is selected in the editing area and Ctrl+F is used, the search term is highlighted as expected. Unlike the behavior in WF, the highlighted term can be copied directly using Ctrl+C without needing to reselect it.

windymilla:
Thanks for that. 

Slight clarification to the description after investigating...
1. There is definitely a bug, which needs fixing.
2. The bug doesn't show on the very first use of this instance of the dialog, i.e. when the S/R dialog has never been used, or has been dismissed, and the Ctrl-click in WF is re-popping it.
3. The problem occurs when the focus in the S/R dialog that pops up is not in the Search box, but elsewhere (actually nowhere) in the dialog. Using the tab key will show the focus swap between the Search & the Replace, but when the dialog gets used for a second time (without being dismissed), it appears as though the focus is in neither search nor replace. This means that although the search term is selected, the Ctrl-C does not go to the Search field, so it doesn't copy the text.
4. Should be fairly easy to fix - when Ctrl-click pops the S/R dialog, it should make sure the focus is in the Search field after it fills the field and selects it.
5. NTS - check to see what happens when S/R is triggered from the main window with Ctrl-F on a selected word, but the S/R dialog was already visible, and focus was in the Replace field.

PR #378, opened 2024-08-10
Ensure search box has focus when SR dialog popped

windymilla:
See issue #373 for detailed description of problem

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -225,6 +225,7 @@ class SearchDialog(ToplevelDialog):
         self.search_box.set(search_string)
         self.search_box.select_range(0, tk.END)
         self.search_box.icursor(tk.END)
+        self.search_box.focus()
 
     def search_clicked(self, opposite_dir: bool = False) -> str:
         """Search for the string in the search box.


windymilla:
Testing notes: you are trying to engineer a situation where the S/R dialog is showing, and the Search field is "selected" (highlighted), but because the focus is not in the Search field (e.g. the Replace field or nowhere), when you press Cmd+C to copy the selected text, you don't get it.

There are lots of ways to do this, but one short way is to use WF, All Words. 
Then Cmd+mouse-1 on a word in the WF list. 
The S/R dialog should pop with the word selected in the Search field.
The focus **is** in Search field, so if you do Cmd+C, it copies the word into your copy/paste buffer. 
Then, without dismissing the S/R dialog, do Cmd+mouse-1 again in the WF dialog. 
This time, the word is selected in the Search field, but the focus (and insert cursor) are not.
Cmd+C now will not copy the apparently selected word from the Search field.
In this PR, the focus should be forced to the search field each time a word in the Search field is selected, so Cmd+C will behave as expected.

srjfoo:
I tried the testing notes, but `master`, another random branch, and `sr-dialog` all behaved the same:

* If GG2 was in the foreground, cmd-left-click on the WF word would fill in the word in the sr dialog, and select it. However, note that it doesn't move the highlighting to any subsequent word after the first one that is clicked without the cmd key being used in conjunction. That is if I left-click on "Amphidolops", and then cmd-left-click on "Anacodon", "Anacodon" is highlighted in s/r search field, and "Amphidolops" remains highlighted in the main text window (whether or not I had caused it to be inserted in the search field in the sr dialog).
* I discovered, however, that if I was using the text input field for github comments as a storage place for the selected words, and forgot to force the active app to be GG2 (i.e. Python), but cmd-left-clicked on words in the WF list, it would cheerfully fill in the search field in the SR dialog, and cheerfully move the highlighted word in WF, but would not bring Python/GG2 to the foreground, would not change the highlighting in the main text window, and would not highlight the word in the search field in the SR dialog. (I hope all this makes sense.)

Short version -- with this particular test, if I'm doing it right, I saw no difference among the three branches I tested in.

srjfoo (approved):

PR #379, opened 2024-08-11
Improve navigation in WF dialog

### LINKED ISSUE #374: Tools->Word Frequency window does not allow easy navigation
GuiGuts 1 offers multiple navigation options within any Word Frequency window, including:

-     Moving to the end or beginning of a list with End and Home keys.
-     Paging up or down with Page Up and Page Down keys.
-     Jumping to the first word starting with a specific letter by pressing that letter.

In contrast, GuiGuts 2 is limited to using the up and down arrow keys plus the scroll bars for navigation.

windymilla:
1. Make focus default to list of words, except when user is typing in the threshold or regex field
2. Home & End keys go to first/last entries, like GG1
3. Arrows and Page Up/Down should scroll list
4. Typing a character jumps to the first word in the list that begins with that character, like GG1

Fixes #374

### DIFF
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -27,6 +27,7 @@ from guiguts.utilities import (
     sound_bell,
     process_accel,
     cmd_ctrl_string,
+    is_mac,
 )
 from guiguts.widgets import (
     ToplevelDialog,
@@ -394,6 +395,13 @@ class WordFrequencyDialog(ToplevelDialog):
         self.text.bind("<Key>", self.goto_word_by_letter)
         self.text.bind("<Home>", lambda _e: self.goto_word(0))
         self.text.bind("<End>", lambda _e: self.goto_word(len(self.entries) - 1))
+        # Bind same keys as main window uses for top/bottom on Mac.
+        # Above bindings work already for Windows
+        if is_mac():
+            self.text.bind("<Command-Up>", lambda _e: self.goto_word(0))
+            self.text.bind(
+                "<Command-Down>", lambda _e: self.goto_word(len(self.entries) - 1)
+            )
         ToolTip(
             self.text,
             "\n".join(
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -214,9 +214,9 @@ class WordFrequencyDialog(ToplevelDialog):
         # Re-run buttons
         rerun_frame = ttk.Frame(header_frame, borderwidth=1, relief=tk.GROOVE)
         rerun_frame.grid(row=0, column=1, rowspan=2, padx=(0, 15))
-        ttk.Button(rerun_frame, text="Re-run", command=word_frequency).grid(
-            row=0, column=0, sticky="NSEW", padx=5, pady=2
-        )
+        ttk.Button(
+            rerun_frame, text="Re-run", command=word_frequency, takefocus=False
+        ).grid(row=0, column=0, sticky="NSEW", padx=5, pady=2)
 
         def change_ignore_case() -> None:
             """Handle changing of ignore case flag - enable/disable buttons & re-run tool."""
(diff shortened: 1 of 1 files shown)


srjfoo:
> 2\. Home & End keys go to first/last entries, like GG1

Other than the extended keyboard (which is not available on Mac laptops), Mac keyboards do not have Home & End keys.

Currently, we're using `Cmd`+`up-arrow` for Home and `Cmd`+`down-arrow` for end in the main text window. That does not work for WF navigation.

... however ... I googled Home and End, and found that there are actual official shortcuts: `Fn`+`left-arrow` and `Fn`+`right-arrow` work for Home and End, respectively, in WF navigation.

I'd appreciate @tangledhelix's opinion on this, as it's a Mac issue. Should we re-think the Mac shortcuts? Personally, I prefer  `Cmd`+`up-arrow` for Home and `Cmd`+`down-arrow` for End, as what works (for me) in WF actually can map to use the `Cmd` key instead of the `Fn` key, and would cause `Cmd`+`left` or `right-arrow` to map to Home and End in WF, and "beginning" or "end of line" in the main text window. Which I don't think is a good idea.

So, my personal preference is to re-map the keys for WF, also, but I don't know if that's the right thing to do, or if I should just learn new shortcuts.

srjfoo:
> 2\. Home & End keys go to first/last entries, like GG1

Other than the extended keyboard (which is not available on Mac laptops), Mac keyboards do not have Home & End keys.

Currently, we're using `Cmd`+`up-arrow` for Home and `Cmd`+`down-arrow` for end in the main text window. That does not work for WF navigation.

... however ... I googled Home and End, and found that there are actual official shortcuts: `Fn`+`left-arrow` and `Fn`+`right-arrow` work for Home and End, respectively, in WF navigation.

I'd appreciate @tangledhelix's opinion on this, as it's a Mac issue. Should we re-think the Mac shortcuts? Personally, I prefer  `Cmd`+`up-arrow` for Home and `Cmd`+`down-arrow` for End, as what works (for me) in WF actually can map to use the `Cmd` key instead of the `Fn` key, and would cause `Cmd`+`left` or `right-arrow` to map to Home and End in WF, and "beginning" or "end of line" in the main text window. Which I don't think is a good idea. (My testing has not been exhaustive.)

So, my personal preference is to re-map the keys for WF, also, but I don't know if that's the right thing to do, or if I should just learn new shortcuts.

windymilla:
@srjfoo - I suspect most Windows apps are similar to the description of the Mac behavior. 
Strictly, Home/End should do beginning & end of line (which matches with the Fn+left/right idea) and Ctrl+Home/End should do top/bottom of file (corresponding more closely with up/down arrows). That's what happens in the main window.
We could say that in WF (and possibly Checker dialogs too) it's the key that does top/bottom of main window that you use for top/bottom of list of errors. I think that makes sense and is more consistent, and it's only GG1 compatibility that made me use Home/End. In fact, on Windows, you **can** actually use Ctrl+Home/End, because if you haven't bound "Ctrl+Home", then when you press that key combo, the Ctrl is ignored, and it does the same as "Home". So, by binding to Home/End, I actually made it compatible with GG1 and also made it consistent with main window behavior. However that doesn't work for the Mac.  

Proposal: Keep it bound to Home/End (for GG1 compatibility and also makes Ctrl+Home/End work, and in fact Fn+Left/Right on Macs) but also add Cmd+Up/Down on Macs. Then a Mac use could use either Fn+Left/Right (if they know that means "Home") or Cmd+Up/Down (if they know that behavior in the main window). I don't see any disadvantage in doing that, so I may do that anyway in a few minutes.

srjfoo:
> [*] The [Human Interface Guidelines][1] note that this combo switches the keyboard layout - and I feel like that's come up before and it's something bound by default if multiple keyboards are installed, and @srjfoo uses it? I don't have multiple keyboards installed so I will let her discuss it ...

Hm. It may be default for multiple keyboards, but I always change it to `Cmd`+`spacebar` (which was the original default shortcut for rotating among keyboards until they repurposed it for Spotlight, which I don't use).

I keep left/right arrow with `Cmd` for beginning and end of line. (Mini-test: `fn`+`left` or `right-arrow` does nothing in this draft message. `Cmd`+ same go to beginning and end of line. I no longer know, I guess, what the defaults would do, so let's keep what we have until and unless we have complaints.

srjfoo (approved):

Issue #367, opened 2024-08-05
Fix Jeebies "Check Level" radio button behaviour


G4OEU:
Radio buttons should normally just set an option then separately clicking the "Re-run" button would apply that option and update the display.

The "Check Level" radio buttons of Jeebies currently set the check level required (Tolerant, Normal or Paranoid) and then they do a 'rerun' to apply the change.

Fix the radio button behaviours so that they just set the check level.

windymilla:
Thanks @G4OEU. I previously mentioned Word Frequency as maybe having the same behavior, but having looked at it, I think WF is OK as it is.

G4OEU:
I would note there is at least one exception to the expected behaviour of radio buttons described above. This is the Sort ('Line & Col' or 'Alpha/Type') radio buttons on tool windows which on clicking have immediate effect.

windymilla:
I think the Sort buttons are a little different - they are not re-running the tool, just ordering the results that are already in the dialog.

G4OEU:
OK, I would agree with that more elegant way of distinguishing the behaviour of the Sort radio buttons from other radio buttons.

windymilla:
Fixed by #380

PR #380, opened 2024-08-12
 Fix Jeebies "Check Level" radio button behaviour

G4OEU:
Fixes issue #367. The "Check Level"
buttons no longer re-run Jeebies when
clicked. They just set the check level
and the user then clicks "Re-run" to
apply the Jeebies check at that level.


windymilla (approved):
Look good to me - thanks!

srjfoo (approved):

PR #381, opened 2024-08-12
Levenshtein edit distance check tool

G4OEU:
Provides a tool that runs a Levenshtein
edit distance check on the opened file.
The edit distance defaults to 1 but the
user can alternatively chose an edit
distance of 2 then re-run the tool.


G4OEU:
Closing this PR due to pylint error and consequent need to update pyproject.toml to add Levenshtein to list of modules required by everyone.

G4OEU:
Re-opened this PR and will try more commits to fix the pylint issue.

src/guiguts/tools/levenshtein.py
```diff
@@ -0,0 +1,683 @@
...
+
+    global _the_levenshtein_checker
+
+    if not tool_save():
+        return
+
+    if _the_levenshtein_checker is None:
+        try:
+            _the_levenshtein_checker = LevenshteinChecker()
+        except DictionaryNotFoundError as exc:
```
windymilla, line 679: When the `LevenshteinChecker` object is created here, it doesn't do any spelling, so can never raise the `DictionaryNotFoundError` exception, so there's not point having the try/except.
However, during running, where the Spellcheck object is created, the exception can be raised. See comment there for more details.

src/guiguts/tools/levenshtein.py
```diff
@@ -0,0 +1,683 @@
...
+            # These must already exist (see enclosing function) before they are
+            # used here. They are the lists and dictionaries into which the results
+            # of this function are 'returned'.
+            nonlocal good_words
+            nonlocal suspect_words
+            nonlocal word_to_lines_map
+            nonlocal all_words_case_map, all_words_counts
+
+            # Will check spelling using SpellChecker methods from spell.py
+            spelling = SpellChecker()
```
windymilla, line 360: This can end up raising SpellChecker's `DictionaryNotFoundError` exception, so either needs to be caught here, or possibly the SpellChecker could be created higher up and passed as an argument.
If it was created during the `init` and stored as `self.spellchecker`, then the try/except at line 679 could remain, though it should use SpellChecker's exception, not the one created at the top of this file.

src/guiguts/tools/levenshtein.py
```diff
@@ -0,0 +1,683 @@
...
+
+
+class LevenshteinEditDistance(StrEnum):
+    """Enum class to store Levenshtein Edit Distance."""
+
+    ONE = auto()
+    TWO = auto()
+
+
+class DictionaryNotFoundError(Exception):
```
windymilla, line 56: Probably should import this exception from `guiguts.spell` rather than define a new class here - it is SpellChecker that ends up raising this.

src/guiguts/tools/levenshtein.py
windymilla: Forgot to say - an easy way to trigger this is to change the language to `xyz` before running Levenshtein.

G4OEU:
Note that the standard 'Sort' button on the Levenshtein tool continues to work as on the other GG2 tools.

windymilla (changes requested):
Looks good and worked fine for me. 
Minor changes to the Exception handling needed, as commented elsewhere

windymilla:
Testing notes (not quite ready for testing, since Quentin has a couple of changes and speed-ups he's adding):
The Levenshtein distance check (Tools menu) finds words that are not in the spelling dictionary that are similar to words that are, for example, if your file has "chesnut" and "chestnut", that should be flagged, because "chesnut" is not in the dictionary.
Some "words" are filtered out, such as those which are just digits, or are only 1 character long.
Once you have the list, you can click in the usual way to visit those locations in the file. All the "bad" locations are listed, but only the first 2 "good" locations to avoid what could be a very long list.

G4OEU:
In regard to windymilla review comments above about exception handling of missing spelling dictionaries, I have moved this check
to before the Levenshtein distance checking commences. Doing it when the SpellChecker object is created is too late as the tool window has already been opened. Furthermore, there is no point in invoking the Levenshtein distance checking if the ability to check spelling of words is not available.

I would be grateful if particular attention in the review can be directed at how I have corrected that exception handling. It now works exactly as it should but to achieve that required three lines of code to be commented with '# type: ignore[attr-defined]' otherwise the static check done my mypy complains about lines 431, 456 and 483. For completeness, see also related code lines 36 and 37 and the handful of code lines from 718. Thanks.

srjfoo:
> Some "words" are filtered out, such as those which are just digits, or are only 1 character long.

When I changed the edit distance to 2, "words" that were only digits were no longer filtered out.

Just out of curiosity, should words that include digits be filtered out?

My test project is [A history of land mammals in the western hemisphere [1913]](https://www.pgdp.net/c/project.php?id=projectID631f616441fff). 

Screenshot of typical numeric results, as that may better help answering my question above. (The quoted bit mentions "[words] that are just digits", which got me curious to know what "words" might legitimately include digits.)

image

---

One thing that I may have been peripherally aware of before, but which (for no apparent reason) really jumped out at me when I checked the edit distance = 2, was the list layout. Visually, the `-----` stands out more than a blank line, so my eyes tend to associate the text between blank lines as a group. As part of my investigation of this, I looked at the output of PPtxt, also, and I get the same feeling there, as well. Is there an equivalent in GG1 for this layout? I looked, but didn't spot anything immediately.

windymilla:
> Visually, the `-----` stands out more than a blank line, so my eyes tend to associate the text between blank lines as a group

Thanks Sharon. I think we'll put off consideration of exact layout for now and tackle it, and maybe pptxt, in a new issue.

Since they are based on the PPWB tools, it may be a slightly different layout would suit the guiguts context. However it might need some extra capability in the Checker dialogs, like underline, highlighting, font size, etc.

windymilla:
> When I changed the edit distance to 2, "words" that were only digits were no longer filtered out.

My guess is it's not actually the edit distance that causes that, but it's because the filtering of all-digit words is only applied to the suspect words whereas an all-digit word is valid as far as the dictionary is concerned. You have a suspect a0100, which is 2 edits away from 000.

We probably don't want to exclude all words  that contain digits, as we want to catch "d0g" (zero for o scanno) for example.

srjfoo:
> We probably don't want to exclude all words that contain digits, as we want to catch "d0g" (zero for o scanno) for example.

Totally fair. And I probably shouldn't have chosen a CTF to run the checks on.

srjfoo (approved):

windymilla (approved):
Thanks - code structure and behavior looks great

PR #386, opened 2024-08-15
Use python regexes in Search/Replace

### LINKED ISSUE #266: Tcl regex bug
Decided this ought to have an issue of its own...

There are [known bugs in Tcl.tk text regex search](https://www.tcl.tk/man/tcl8.6/TkCmd/text.htm#M196) relating to searching across multiple lines. I think my first attempt at the above example hits one of those.
Search for `\nabc` instead of `^\nabc` and the above isn't found (except when it is - try searching for longer strings, like `\n\n\nabc` or `xyz\n\n\nabc`, then leave the selection as it is, but delete chars from the beginning of the regex - it's possible to get `\nabc` to match, but if you start again from the top of the file, it won't find it.

Solutions?

- GG1 solution if regex contains `\n` is to load whole file into a string & search in that, then convert back to rowcol
- https://www.tcl.tk/man/tcl8.6/TclCmd/regexp.htm#M9 lists `(?p)` and `(?w)` that can affect handing of newlines- didn't seem to help in a quick experiment

### LINKED ISSUE #297: Some regex replaces aren't working as expected
Reported via Slack:

> Tried a search, \b' & replace ’ on a section of dialect.
 "I'd like to
know who'd 'a benn 'er! I wouldn't 'a benn 'er; Ed wouldn't
'a benn 'er; Joe wouldn't 'a benn 'er, and I'd like to know who
would 'a benn 'er!"
I noticed something strange. Originally, the count was 10 leading left curly quotes, but after the change, none were converted to right curly quotes. Even more surprising, the quote count dropped to 0! To double-check, I copied and pasted the section into GG 1.6. My eyes weren't playing tricks on me (the text might be a bit small in GG2, but I assume there will eventually a way for the user to adjust that ).

Also

> I'm trying to replace (\n{3}(?=[A-Z]{2})) with (\n\n.sp 2\n.h3\n). Unfortunately, the search and replace isn't making any changes.

windymilla:
To avoid bugs and limitations in Tcl regexes, slurp the relevant part of the file into a python string and do the searches there before converting the match locations back into line/col format as needed.

Also fix bug in Replace where word boundaries, and look-ahead caused Replace to do nothing even though the match had been found.

This affects, Search, Replace, Search & Replace, Count, Find All and Replace All.
Could potentially affect WF, where you click a word and it takes you to the location in the file.

Fixes #266 
Fixes #297

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -661,7 +661,6 @@ class File:
             search_range,
             nocase=False,
             regexp=True,
-            wholeword=False,
             backwards=False,
         ):
             maintext().delete(
@@ -684,7 +683,6 @@ class File:
             search_range,
             nocase=False,
             regexp=True,
-            wholeword=False,
         )
         if not flag_matches:
             return False
--- a/src/guiguts/highlight.py
+++ b/src/guiguts/highlight.py
@@ -42,7 +42,6 @@ def highlight_selection(
     tag_name: str,
     nocase: bool = False,
     regexp: bool = False,
-    wholeword: bool = False,
 ) -> None:
     """Highlight matches in the current selection.
     Args:
@@ -51,16 +50,13 @@ def highlight_selection(
     Optional keyword args:
         nocase (default False): set True for case-insensitivity
         regexp (default False): whether to assume 's' is a regexp
-        wholeword (defalut False): whether to match only whole words
     """
 
     if not (ranges := maintext().selected_ranges()):
         return
 
     for _range in ranges:
-        matches = maintext().find_matches(
-            pat, _range, nocase=nocase, regexp=regexp, wholeword=wholeword
-        )
+        matches = maintext().find_matches(pat, _range, nocase=nocase, regexp=regexp)
         for match in matches:
             maintext().tag_add(
                 tag_name, match.rowcol.index(), match.rowcol.index() + "+1c"
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -207,7 +207,7 @@ class MainText(tk.Text):
         self.bind_event("<KeyRelease>", self._on_change, add=True, force_break=False)
         # Add mouse event here after column selection bindings above
         self.bind_event("<ButtonRelease>", self._on_change, add=True, force_break=False)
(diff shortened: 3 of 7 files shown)


windymilla:
Speed seems acceptable: about 3000 per second, whether counting, finding or replacing

windymilla:
### Testing notes:
Try regex and non-regex Search, Replace, Replace All, Find All, Count

### Timing/speed:
Because the Perl/Tk text widget doesn't support multiline regex matches, it fired up special code if your regex contained the `\n` character. Depending on how many things would match the regex you were searching for, this could lead to a substantial delay before it was ready to do the first search for that regex (then faster for subsequent searches). This version doesn't treat those regexes as special, and essentially does the same for all searches. This should mean that searching time is more predictable for the user.

### Changes to supported regexes

The following 2 constructions work for searching in master:

1. Word boundaries, e.g. `\babc` should find `abcdef` but not `xabcdef`
2. Look ahead, e.g. `srj(?=foo)` matches the `srj` in `srjfoo` but not in `srjzoo` (because it is/isn't followed by `foo`)

However, when you try to replace those strings, the replace fails. That should be fixed in this PR.

Similarly negative lookahead, e.g. `srj(?!foo)` matches the `srj` in `srjzoo` but not in `srjfoo` (because it isn't/is followed by `foo`)
This works in master, and replace should work too. So, no change in this PR (hopefully)

Lookbehind does not work in master at all - regex will fail to compile since it's not valid in Tcl regexes.
Example: `(?<=srjf)oo` matches the `oo` in `srjfoo` but not in `srjzoo` (because it is/isn't preceded by `srjf`)
Similarly for negative lookbehind
Example: `(?<!srjf)oo` matches the `oo` in `srzfoo` but not in `srjfoo` (because it isn't/is preceded by `srjf`)

Other interesting things now supported are Unicode characters and properties, e.g. `\p{L}` for a "letter", and `\X` which is "the unicode version of `.`" in that it matches something like `o` followed by `combining tilde` as one character rather than two. There are lots of `\p` properties - see https://www.regular-expressions.info/unicode.html.

srjfoo (approved):

PR #387, opened 2024-08-15
Highlight current line always

windymilla:
If the code moved the insert cursor, but the window didn't need to scroll because of it, it didn't trigger the update code.

Fixed by overloading the mark_set method and catching the case where it's the insert cursor.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -338,9 +338,17 @@ class MainText(tk.Text):
         self._on_change()
 
     def replace(self, index1: Any, index2: Any, chars: str, *args: Any) -> None:
-        """Override method to ensure line numbers are updated."""
+        """Override method to ensure line numbers are updated.
+
+        Also preserve pagemark locations within the replacement."""
         self.replace_preserving_pagemarks(index1, index2, chars, *args)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: Use S/R dialog to search within the current page so it doesn't scroll.

srjfoo (approved):
This may be a Window and possibly a Linux problem. Or maybe it's the Macs that have the problem. In `master`, I'm not seeing the issue as described, because literally every time I do a search, whether using the "search" button or the list from "Find all", if the search term appears multiple times on a line, the line isn't repositioned, but if the search term is even one word down, the line with the cursor on it is centered top to bottom. So, it moves and is updated. (Is that supposed to happen? I thought it actually got fixed somewhere, but if so, it wasn't in the Search widget.

Anyway, I can't see any difference in behavior from before, but the PR didn't break anything! 😁 

Edit: And yes the search term appeared multiple times on the same page. I chose a _**very**_ common word that appeared lots of times on the page.

Edit 2: @windymilla explained that this was only an issue in very short files, where the lines can't be centered, and so don't move. I was then able to duplicate the problem, and that this PR fixes it.

PR #388, opened 2024-08-15
Delete selection when inserting compose sequence

### LINKED ISSUE #383: GG2 Tools->Unicode>Compose Sequence does not replace the selected character(s).
When text is selected in GG1, the generated character substitutes the entire selection. However, in GG2, the generated character is inserted before the selected text, leaving the original text intact. 

Typically, applications overwrite selected text with new input, but insertion occurs only when no text is highlighted.

windymilla:
If characters are selected, delete them before inserting the character created by the compose sequence.

Fixes #383

### DIFF
--- a/src/guiguts/widgets.py
+++ b/src/guiguts/widgets.py
@@ -633,6 +633,16 @@ def insert_in_focus_widget(string: str) -> None:
     if _text_focus_widget is None or not _text_focus_widget.winfo_exists():
         return
     assert isinstance(_text_focus_widget, (tk.Entry, tk.Text))
+
+    if isinstance(_text_focus_widget, tk.Text):
+        sel_ranges = _text_focus_widget.tag_ranges("sel")
+        if sel_ranges:
+            _text_focus_widget.mark_set(tk.INSERT, sel_ranges[0])
+            _text_focus_widget.delete(sel_ranges[0], sel_ranges[1])
+            _text_focus_widget.tag_remove("sel", "1.0", tk.END)
+    elif isinstance(_text_focus_widget, tk.Entry):
+        if _text_focus_widget.selection_present():
+            _text_focus_widget.delete("sel.first", "sel.last")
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: 
To replace `xxx` with `é`, select the `xxx`, then use Cmd+i to pop the Compose Sequence dialog, and type `e` then `/`.
In master, `xxx` will not be deleted, which makes it inconsistent with the behavior when you use standard copy/paste to insert the new character.

If (for some bizarre reason) you have a column selection, then just the first line of the column selection will be replaced with the inserted character. I don't think it's worth trying to do anything more complex than that.

srjfoo (approved):
LGTM!

> If (for some bizarre reason) you have a column selection, ...

I have been known to do a column select with the intention of replacing the characters in each row of the column with the selected content on the clipboard, or with the intention of typing in duplicate text on each line. N++,  BBEdit, Sublime Text and VSCode  all do this. I'm not saying that this should be true of GG2 (in fact, column-select in GG2 doesn't work that way; the only way you get column action on the column select other than deleting/cutting/copying it is to use `Edit->Column Paste`.

Would it be a good idea to document that it doesn't work as some of the popular text editors do, though. I can imagine it being a challenging feature to implement, and very far down on the feature list if you decide it's a good idea.

windymilla:
> I'm not saying that this should be true of GG2 (in fact, column-select in GG2 doesn't work that way; the only way you get column action on the column select other than deleting/cutting/copying it is to use `Edit->Column Paste`.

If I understand correctly, then I don't think this is true.
If I put "dog" into my copy/paste buffer, then column select a few lines, then hit Ctrl-V (Cmd-V) it does "smart paste" which detects that you have a column selection, and pastes the word "dog" in on every line that you had column selected, overwriting the words that were selected.

However, that doesn't work for inserting a compose sequence - it could be made to, but probably not top priority. The user will have to use compose sequence to insert the character, then Cut it to their paste buffer, then column select & paste.

PR #390, opened 2024-08-18
Use main text font for S/R fields

### LINKED ISSUE #389: Allow S/R fields (at least) to have configurable font

windymilla:
Since these fields typically have the same type of text as the main window, let them share the font.

Fixes #389


windymilla:
Testing notes: 
Pop S/R dialog & type something in the Search & Replace fields. Font should be same as main window, and should be reconfigured via the prefs dialog.

srjfoo (approved):

PR #391, opened 2024-08-19
Improve page marker positions for S/R

### LINKED ISSUE #282: Improve S/R page break positioning algorithm
If the search/replace strings have the same number of lines (which would almost always be true) then treat each line separately, so page breaks would stay on the same line. Within a line, it would use length scaling as it currently does.

windymilla:
If Search & Replace strings have same number of lines, then replace line-by-line so that page break marks remain on the correct line.

Fixes #282


windymilla:
Testing notes: see issue #282 for example of the kind of S/R this should improve.

PR #392, opened 2024-08-20
Implement `\C...\E` to execute python code

windymilla:
Also, trap bad regexps in replace strings


windymilla:
Testing notes:
1. This should replace any captured string with an uppercase version. The python syntax is `string.upper` so we enclose `\1` in double quotes so that it is a string, and run the `upper` function on it.
S: `the (captured string)`
R: `the \C"\1".upper()\E`

2. This should add 5 to any number:
S: `(\d+)`
R: `\C\1+5\E`

srjfoo (approved):

PR #399, opened 2024-08-23
Regex bugs

### LINKED ISSUE #395: Search->Search & Replace--R & S selects but doesn't highlight next find
I used a regular expression `\[\*([^\]\[]+?)\]` to search for `[** Proofers Notes]`. I left the replacement field empty then enabled both 'Regex' and 'Wrap-around' options. 

The first proofreader note was found and highlighted after clicking search. I next clicked 'Replace & Search.' The 1st note was correctly deleted. However, the next proofreader note found was unhighlighted. Despite not being highlighted, clicking 'Replace' deleted the 2nd note, indicating that the 2nd find was selected despite the lack of highlighting.

### LINKED ISSUE #396: Search->Search & Replace Count doesn't count some things
I used the regular expression `([\*#])/\n(\-[^\n]+[\-\\]\n+)/\1\n` to search for a specific pattern and replace it with `\2`. With 'Wrap-around' and 'Regex' options enabled, the 'Count' function reported `Found: 0 matches in entire file.` However, the correct number of matches should have been 3.
[fan-src.zip](https://github.com/user-attachments/files/16715567/fan-src.zip)

### LINKED ISSUE #398: Consider if anything needs doing about backward searches with backreferences and/or lookbehind/ahead
[Comment from author of regex module](https://github.com/mrabarnett/mrab-regex/issues/542#issuecomment-2305683404):

> It's not just searching backwards, it's matching backwards.
> It needs to capture before it can refer to what it has captured, so when matching forwards it needs to be r"(.)\1", but when matching backwards it needs to be r"\1(.)".
> This is also true within lookbehinds.

Possible solutions for GG2:
1. Document the above
2. Edit the regex if the above circumstances are detected, to swap the positions of the capture group and the backreference
3. Do reverse search by doing repeated forward search and returning the last match found (probably only if there are backreferences or lookahead/behinds)

windymilla:
1. Highlight next match correctly when using Replace & Search
2. Count regexes with backreferences (or lookarounds) correctly
3. Replace All with backreferences (or lookarounds) correctly
4. Search backward for regexes with backreferences (or lookarounds)  correctly

Fixes #395
Fixes #396 
Fixes #398

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1465,10 +1465,17 @@ class MainText(tk.Text):
             and None if no match; also the index into the slurp text of the match start, which is
             needed for iterated use with the same slurp text, such as Replace All
         """
+        slurp_newline_adjustment = 0
         if regexp:
             # Since "^" matches start of string (when not escaped with "\"), and we want it
             # to match start of line, replace it with lookbehind for newline.
-            search_string = re.sub(r"(?<![\[\\])\^", r"(?<=\\n)", search_string)
+            if re.search(r"(?<![\[\\])\^", search_string):
+                search_string = re.sub(r"(?<![\[\\])\^", r"(?<=\\n)", search_string)
+                # Need to make sure there is a newline before start of string
+                # if string starts at the beginning of a line
+                if slurp_start.col == 0:
+                    slurp_text = "\n" + slurp_text
+                    slurp_newline_adjustment = 1
         else:
             search_string = re.escape(search_string)
         if wholeword:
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes:
1. Use regexes with backrefs, e.g. `(s.j)foo \1zoo` should match `srjfoo srjzoo` but not `srjfoo sxjzoo`.
2. Use regexes with lookahead/behind, as suggested in #386 
3. Use the regexes with Count, Replace All, Find All, and **reverse** Search and R&S.
4. In master, regexes with backrefs will fail to do those operations correctly, and some with lookaround may fail, though I haven't managed to make it happen.
5. In this PR, all those regexes should work.

srjfoo (approved):

PR #400, opened 2024-08-23
Investigate narrow S/R dialog

### LINKED ISSUE #393: S/R dialog is narrow each time it pops up on Macs
Other dialogs don't have the problem.
Things to look at/try in search.py:
1. Remove `kwargs["resize_y"] = False` in SearchDialog `init`
2. Remove `self.minsize(400, 100)` or change the min width to be bigger
3. Change Combobox widths to something in pixels, like 150, instead of 30 (chars).

windymilla:
@srjfoo - please see if this fixes #393 
It's not going to be merged as it is - it's just for testing purposes.

I commented out two lines. If they have worked, you could uncomment each in turn and check which is the one that fixes it.

Note, this PR does not have the latest backref regex work in it.

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -57,9 +57,10 @@ class SearchDialog(ToplevelDialog):
         except AttributeError:
             SearchDialog.selection = tk.BooleanVar(value=False)
 
-        # kwargs["resize_y"] = False
+        if not is_mac():
+            kwargs["resize_y"] = False
         super().__init__("Search & Replace", *args, **kwargs)
-        # self.minsize(400, 100)
+        self.minsize(400, 100)
 
         # Frames
         self.top_frame.columnconfigure(0, weight=1)
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -57,9 +57,10 @@ class SearchDialog(ToplevelDialog):
         except AttributeError:
             SearchDialog.selection = tk.BooleanVar(value=False)
 
-        # kwargs["resize_y"] = False
+        self.enable_geometry_saving = False
+        kwargs["resize_y"] = False
         super().__init__("Search & Replace", *args, **kwargs)
-        # self.minsize(400, 100)
+        self.minsize(400, 100)
 
         # Frames
         self.top_frame.columnconfigure(0, weight=1)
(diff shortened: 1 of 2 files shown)


src/guiguts/search.py
```diff
@@ -57,9 +57,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         except AttributeError:
             SearchDialog.selection = tk.BooleanVar(value=False)
 
-        kwargs["resize_y"] = False
+        # kwargs["resize_y"] = False
```
srjfoo, line 60: 🤣  When I uncomment line 60, the search box looks like this:

image

It  flashes larger, then reduces to the thin line you see. It's resizeable -- just hover over the line and you can drag it open.

---

src/guiguts/search.py
```diff
@@ -57,9 +57,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
         super().__init__("Search & Replace", *args, **kwargs)
-        self.minsize(400, 100)
+        # self.minsize(400, 100)
```
srjfoo, line 62: Re-commenting line 60 and uncommenting line 62 has the same effect as having both of the commented out. That is, the S&R dialog retains its size and placement from the last time it was opened.

Note that in either case (i.e. either with both 60 and 62 commented out, or with only 60 commented out, the y-axis is now modifiable.

src/guiguts/search.py
windymilla: Was that with `--nohome` or possibly using a setting from the Prefs file that is not valid now it's resizable?

srjfoo:
Apologies for the delay in re-testing.

Used `guiguts -d --nohome` to bring up guiguts. The first time I invoked the s/r dialog, it was nicely sized. Once closed, reopening it would flash to the proper size, but immediately resize to the smaller version that started this investigation in the first place.

windymilla:
Reverted previous attempt, and committed a hopefully working version (based on previous testing).
Note that the y-size will be modifiable on Macs - may be the best we can do for now.

srjfoo (approved):
Yes, this fixes it. Or, at least ...
* When using `--nohome`, changed width is remembered within the session
* When using GGprefs, changed width is remembered across sessions

And I think that's what we were going for.

Adjustable height doesn't bother me -- this is at least workable.

PR #407, opened 2024-08-27
Add Unicode Blocks dialog

windymilla:
In Tools->Unicode menu.
Also doubles as "Commonly-used characters" dialog, by using the first entry in the dropdown. It should remember which block was shown the last time the dialog was shown. Clicking a character should insert it. Hovering over a button should show a tooltip with the character ordinal & name.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -35,6 +35,7 @@ from guiguts.misc_dialogs import (
     PreferencesDialog,
     ComposeSequenceDialog,
     ComposeHelpDialog,
+    UnicodeBlockDialog,
 )
 from guiguts.misc_tools import (
     basic_fixup_check,
@@ -374,6 +375,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         preferences.set_default(PrefKey.SPELL_THRESHOLD, 3)
         preferences.set_default(PrefKey.UNMATCHED_NESTABLE, False)
+        preferences.set_default(
+            PrefKey.UNICODE_BLOCK, UnicodeBlockDialog.commonly_used_characters_name
+        )
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -605,6 +609,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             lambda: fraction_convert(FractionConvertType.SUPSUB),
         )
         unicode_menu = Menu(menu_tools, "~Unicode")
+        unicode_menu.add_button("Unicode ~Blocks", UnicodeBlockDialog.show_dialog)
         unicode_menu.add_button(
             "~Normalize Selected Characters",
             unicode_normalize,
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -6,6 +6,7 @@ import unicodedata
 
 import regex as re
 
+from guiguts.maintext import maintext
 from guiguts.preferences import (
     PrefKey,
     PersistentBoolean,
@@ -279,9 +280,6 @@ class ComposeSequenceDialog(OkApplyCancelDialog):
             elif match := re.fullmatch(r"#(\d{2,})", sequence):
                 # Or specify in decimal following '#' character
                 char = chr(int(match[1]))
-        # Unprintable characters shouldn't be inserted
-        if not char or not char.isprintable():
-            return
         insert_in_focus_widget(char)
         self.entry.add_to_history(self.string.get())
         if not force:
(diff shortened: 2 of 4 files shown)


srjfoo:
Comments:
* Maybe make the character buttons slightly taller. At one point, when looking at the Combining Half Marks block, the group of marks that were low disappeared from the buttons. Actually, also let the buttons size dynamically.
image

* Let tooltips show for characters not on the user's computer, if possible.
image

* If I resize the text while the Unicode Block dialog is in front (actually, any widget, including the Settings widget), the text in the text window is resized immediately, as are the line numbers, and the main window line height is adjusted, but the line height for the line numbers is not adjusted until the main window is brought to the front. Unicode block characters on the buttons don't change size until the window is brought to the front (actually, all of the widgets will adjust at the same time -- if I resize and go to s/r and type something in and hit return, the unicode block sizes will adjust too.

* I'm seeing some characters that are truncated. Not entirely sure how to tell what causes it; certainly some of it must be width; for some, it's obvious when you look at them, but others not so. For instance, if the tail of a left-pointing arrow is cut off, you might not notice, but if the point of a right-facing arrow is cut off, you're more likely to. This screenshot shows some cut-off emoji, but they're far from the only sorts of characters that are cut off.
image

* It takes a long time to display any of the blocks.

The resizing issue is probably unrelated to the Unicode block, but does.

windymilla:
- Needless to say, the buttons are a perfectly sensible height on Windows - the height is not set in the code, so should default to "1 character high". Looks like MacOS is allergic to sensible tkinter widget sizing.
- It should show tooltips for all "printable" characters - is there an example of one that you think should have a tooltip but doesn't? I can also do try doing a minimal tooltip (with just the U+ABCD number) for those where there is no name in the Unicode database.
- The cut-off characters are weird - the button is plenty wide enough isn't it? I can see similar on Windows. Given that the font is allegedly monospaced, it's very strange.
- I'm punting the resize-delay issues as not-related, and also not too serious. Someone should add a separate issue for that. Again, it doesn't happen on Windows of course.
- The time delay displaying the blocks (I assume you mean populating the dialog with buttons) is substantially down to the creation of Tooltips - I'll look into that.

windymilla:
Font size issue added: #408 
Time delay fixed by simple change to Tooltip class

srjfoo:
> * Needless to say, the buttons are a perfectly sensible height on Windows - the height is not set in the code, so should default to "1 character high". Looks like MacOS is allergic to sensible tkinter widget sizing.

Yeah, after our discussion in Slack, it appears that the "default" theme is particularly annoying here. <sigh>

> * It should show tooltips for all "printable" characters - is there an example of one that you think should have a tooltip but doesn't? I can also do try doing a minimal tooltip (with just the U+ABCD number) for those where there is no name in the Unicode database.

It's now loading at least the code point number. Last night, it wasn't loading anything, though.

> * The cut-off characters are weird - the button is plenty wide enough isn't it? I can see similar on Windows. Given that the font is allegedly monospaced, it's very strange.

That appears to be fixed with your latest commit, too.

> * I'm punting the font resize-delay issues as not-related, and also not too serious. Someone should add a separate issue for that. Again, it doesn't happen on Windows of course.

Thanks for adding the issue. Yes, it's entirely usable as-is, just a little less polished than we'd like. Definitely worth punting on.

> * The time delay displaying the blocks (I assume you mean populating the dialog with buttons) is substantially down to the creation of Tooltips - I'll look into that.

Looks good now.

After looking at the Unicode block in the other two themes, I'm wondering if we ought to just suggest that Mac users learn how to use the macOS Character Viewer (via "Show Emoji & Symbols"). Or (reluctantly, on my part, because I think that the color scheme for the default theme is better designed) recommend that they use either the Light or the Dark theme and avoid the default theme. Seems that I really need to be switching themes back and forth during testing, when it's a display issue.

windymilla:
> It's now loading at least the code point number. Last night, it wasn't loading anything, though.
Just to confirm, that's what I would expect. If it can't get the Unicode name for the character, the tooltip will now just show the code point number

srjfoo:
Turns out that this (and maybe other stuff that seems weird on Macs that works on Windows) may be related to the theme. Apparently, the default theme on Windows is behaviorally much closer to the Light and Dark themes than it is on Macs (though apparently "default" maps to light on Windows, even if the system is set to dark, and it maps to a different set of dark colors on macOS when the system is dark).

Current iteration (which will change), invisible characters cannot be detected in dark or light themes, and can only maybe be detected (at least on Macs) by a very slight color change for the row where they are in the default theme (possible solution is to hopefully add a thin outline). I think this is primarily a problem for the Variation Selectors block.

Scrollbars allow one to scroll to see all of a large block without resizing the widget window. I've seen some weirdness in where short blocks are placed in the windows. With scrolling turned on, even without a scrollbar, the block may (or may not) move vertically within the display pane.

If possible, when a user changes to a new block, the scroll-bar position should default to the top (I just switched from a long block, bottom of the block, to a short block, and everything in the block was hidden because the scrollbar was still at the bottom.

Unicode Block widget may or may not resize upon quiz and re-start after changing theme and/or font size (that probably should include font, but I haven't tested that). I thought it did, but just had an example where it didn't.

srjfoo (approved):

PR #411, opened 2024-08-29
Dock/Undock Image Window without errors

windymilla:
Checkbox used for Dock/Undock was initialized with a bool instead of a tk.Boolean.

Also removed use of literal strings for settings.


windymilla:
Testing notes - Use View-->Dock Image
Should now be no error.

srjfoo (approved):

PR #414, opened 2024-08-30
Trap bad regex in S/R dialog

windymilla:
1. Change text color to red if current string is invalid regex.
2. Display regex compilation error if user still tries to use a red (bad) regex.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2108,6 +2108,13 @@ class MainText(tk.Text):
             return ""
         return img_from_page_mark(mark)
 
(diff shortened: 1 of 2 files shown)


srjfoo:
This shows the red on the default theme in macOS, when the OS is in dark mode (I checked the light theme, but not default with a light OS. (GG2 font size 13.)
image

Would a darker red work on the default theme for Windows? (It probably wouldn't for the Dark theme.) I'm willing to experiment, but time to stop for now. 😁

srjfoo (approved):

PR #417, opened 2024-09-02
Tidy title bar when no filename

windymilla:
Previous edit meant the preceding hyphen was still shown, even if there was no filename.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -224,7 +224,8 @@ class Guiguts:
             max_width = int(maintext().winfo_width() / maintext().font.measure("0"))
             if len_title > max_width:
                 filetitle = "..." + filetitle[len_title - max_width :]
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: 
1. Start GG2
2. Look in title bar. 
3. In this PR it should say "Guiguts 2.0.0a6". 
4. In master, it would have said "Guiguts 2.0.0a6 - " with a trailing hyphen.
5. Load a file - the hyphen and filename should now be displayed
6. File-->Close - the filename and hyphen should disappear

srjfoo (approved):
Noted in testing: as soon as you start typing in a new document, "edited" appears. Would it be more appropriate to do what at least some apps do on macOS, and use "untitled document" or simply "untitled" in place of the title of an unsaved, initially blank document?

Issue #421, opened 2024-09-06
Update the "supported" Unicode blocks list


windymilla:
Use http://unicode.org/Public/UNIDATA/Blocks.txt
combined with top answer here: https://stackoverflow.com/questions/243831/unicode-block-of-a-character-in-python
or possibly the bottom similar answer.

This will give ALL unicode blocks. Do we want to hide some from the Unicode Block dialog and/or the Unicode Search dialog?

windymilla:
Also potentially useful is the 3.2 blocks list: https://www.unicode.org/Public/3.2-Update/Blocks-3.2.0.txt

srjfoo:
> This will give ALL unicode blocks. Do we want to hide some from the Unicode Block dialog and/or the Unicode Search dialog?

The list doesn't look nearly as overwhelming in this form in GG2 as it does to me in GG1, so I have no strong feelings one way or the other (though there are a lot there that don't look useful). One thing that many apps have (I don't know if it would be easy to implement or whether it would be worth it in a different PR) is the ability to navigate open menus by typing the first two or three letters. This does not appear to be default behavior for all menus, though. Although it works in the app level menus (i.e. the menubar) for me for GG2, it doesn't work for the widget.

Issue #270, opened 2024-06-02
Unicode character entry tools


windymilla:
Maybe combine/rationalize what is in GG1 which is:

1. Unicode menu with each block as an entry - pops a  Unicode Block Dialog containing a block of characters (with pulldown to change blocks) - click a character to insert it
2. Tools->Character Tools "Commonly Used Chars"
3. Tools->Character Tools "Unicode Character Entry" - type the dec/hex ordinal and the character is shown & can be inserted in text
4. Tools->Character Tools "Unicode Character Search" - type words from the char name, e.g. "capital greek" and all matching characters are shown. You click one, and it pops the Unicode Block Dialog showing the block (go to point 2)
5. Tools->Character Tools "Compose Sequence" - type keystrokes (or a dec/hex ordinal) to enter any character (e.g. `<"` for left-curly-double-quote)

windymilla:
Point 5 partly done by #280

tangledhelix:
Suggestions:

Item 1 still seems useful but item 2 could just be one page (perhaps the first page in the list?) in that window, collapsing them to one dialog/widget.

Items 3 and 4 could collapse to one dialog with a modal input, e.g.:
- radio button (string, hex, decimal)
- text input box

Depending which option is selected, either behave like item 4 (string) or item 3 (hex, decimal). You could even forego the radio button and try to return valid results for all 3 cases based on whatever the user inputs.

PR #425, opened 2024-09-07
Add Unicode Search dialog

windymilla:
User can either type words (e.g. "capital latin") and find all Unicode characters that contain all those words in their name. Or they can type a hex code point number, e.g. "a9" for the copyright sign.

Clicking on the character/name in the list will insert the character; clicking on the block name will open the Unicode block dialog.

Note: The warning symbol ⚠ (non-emoji version) will appear next to a character's name if the character was added to Unicode after version 3.2 (March 2002)

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -612,10 +612,10 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             lambda: fraction_convert(FractionConvertType.SUPSUB),
         )
         unicode_menu = Menu(menu_tools, "~Unicode")
-        unicode_menu.add_button("Unicode ~Blocks", UnicodeBlockDialog.show_dialog)
         unicode_menu.add_button(
             "Unicode ~Search/Entry", UnicodeSearchDialog.show_dialog
         )
+        unicode_menu.add_button("Unicode ~Blocks", UnicodeBlockDialog.show_dialog)
         unicode_menu.add_button(
             "~Normalize Selected Characters",
             unicode_normalize,
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -36,6 +36,7 @@ from guiguts.misc_dialogs import (
     ComposeSequenceDialog,
     ComposeHelpDialog,
     UnicodeBlockDialog,
+    UnicodeSearchDialog,
 )
 from guiguts.misc_tools import (
     basic_fixup_check,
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes: I'm aware that some characters apparently have no block name. This is due to the point raised in #421 and will be addressed separately

srjfoo:
> warning symbol ⚠ (non-emoji version)

I appreciate the attempt, but apparently python-macOS ignores the `VS-15`. (🙄)

image

---

Is it worth mentioning that if you see a question mark enclosed in a square, that means your computer doesn't have a font on it with that character, or is that pretty obvious?

---

Perhaps allow code points to be preceded by either `U+` or `X` or `x`. I found one case where I thought I was typing in a code point, and that hex designation was actually part of a couple of characters' Unicode names 🤣 

---

And finally, it may not be obvious, and it may not be of use to anyone but me during testing (if that's so, ignore this), but a couple of things that I thought of to do via this dialog that I couldn't do were
* type an actual character in the search form, which would identify the block, and then
* be able to click on the block, itself, to see all the associated characters in the block.

srjfoo:
> * type an actual character in the search form, which would identify the block, and then
>       ~* be able to click on the block, itself, to see all the associated characters in the block.~
> 
> 
> Edit: struck out that second line because it actually is implemented -- it just didn't seem to work the first time I tried it.

Hm ... I looked closer, and (at least for me right now) it brings up the unicode block window with the block name in it, but doesn't change the block that is displayed. It works the first time (based on restarting GG2), but thereafter stays where it is.

image

windymilla:
> > warning symbol ⚠ (non-emoji version)
> 
> I appreciate the attempt, but apparently python-macOS ignores the `VS-15`. (🙄)

Despite what I wrote (and *Windows* behavior) the Mac is correct. I was trying for the emoji version with VS-16, not the non-emoji version with VS-15. I'll also add VS-16 to the tooltip.

> Is it worth mentioning that if you see a question mark enclosed in a square, that means your computer doesn't have a font on it with that character, or is that pretty obvious?

Obvious enough and can happen in many places, I think
 
> Perhaps allow code points to be preceded by either `U+` or `X` or `x`. I found one case where I thought I was typing in a code point, and that hex designation was actually part of a couple of characters' Unicode names 🤣

Good idea 

> * type an actual character in the search form, which would identify the block, and then

I'll do this too, if it drops through the keyword check and the code point check. I considered allowing you to type several characters and displaying info on all of them, but that would be confusing, e.g. if you mistyped "ACUET" for "ACUTE", it would then display info on all the letters in "ACUET" which would surprise the user. So, I'll limit it to the single non-blank character case you suggested.

windymilla:
@srjfoo, I think all the points are addressed in the latest commit.

windymilla:
This completes the work described in issue #270

srjfoo (approved):

PR #426, opened 2024-09-09
Add fixup tools for Illustrations and Sidenotes

G4OEU:
Adds 'Illustration Fixup' and 'Sidenote
Fixup' to the Tools menu.

The tools share the same module.
There are two functions used to move
a tag up and two similar functions to
move a tag down.

The movements allowed for a tag in
either direction conform to the
current Guiguts Manual for GG1.

The tools assume the file being
checked have Illustration and Sidenote
tags that conform to the rules set out
in the Formatting Guidelines. If the file
does not conform then some tags will
not be allowed to move if there is a
risk of incorrect deletions after a tag
is moved.

The placement of a tag after a move
always conforms to the Formatting
Guidelines; that is, it will be separated
from other file content by a blank
line before and after the tag line(s).

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -577,14 +577,14 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_tools.add_button("PP~txt...", lambda: pptxt(self.file.project_dict))
         menu_tools.add_button("~Jeebies...", jeebies_check)
         menu_tools.add_button(
-            "W~ord Distance Check...", lambda: levenshtein_check(self.file.project_dict)
+            "Word ~Distance Check...", lambda: levenshtein_check(self.file.project_dict)
         )
         menu_tools.add_separator()
         menu_tools.add_button("~Page Separator Fixup...", page_separator_fixup)
         menu_tools.add_button("~Footnote Fixup...", footnote_check)
-        menu_tools.add_button("Sidenote Fixup...", lambda: illosn_check("Sidenote"))
+        menu_tools.add_button("Side~note Fixup...", lambda: illosn_check("Sidenote"))
         menu_tools.add_button(
-            "Illustration Fixup...", lambda: illosn_check("Illustration")
+            "~Illustration Fixup...", lambda: illosn_check("Illustration")
         )
         menu_tools.add_separator()
         menu_tools.add_button("~Rewrap All", self.file.rewrap_all)
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -36,6 +36,8 @@ from guiguts.misc_dialogs import (
     PreferencesDialog,
     ComposeSequenceDialog,
     ComposeHelpDialog,
+    UnicodeBlockDialog,
+    UnicodeSearchDialog,
 )
 from guiguts.misc_tools import (
     basic_fixup_check,
@@ -51,7 +53,7 @@ from guiguts.misc_tools import (
 )
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
-from guiguts.root import root
+from guiguts.root import root, ImageWindowState
 from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
 from guiguts.tools.pptxt import pptxt
@@ -224,7 +226,8 @@ class Guiguts:
             max_width = int(maintext().winfo_width() / maintext().font.measure("0"))
             if len_title > max_width:
                 filetitle = "..." + filetitle[len_title - max_width :]
-            filetitle = " - " + filetitle
+            if filetitle:
+                filetitle = " - " + filetitle
         root().title(f"Guiguts {version('guiguts')}" + modtitle + filetitle)
         if is_mac():
             root().wm_attributes("-modified", maintext().is_modified())
@@ -310,7 +313,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_callback(PrefKey.AUTO_IMAGE, self.auto_image_callback)
         preferences.set_default(PrefKey.BELL_AUDIBLE, True)
         preferences.set_default(PrefKey.BELL_VISUAL, True)
-        preferences.set_default(PrefKey.IMAGE_WINDOW, "Docked")
+        preferences.set_default(PrefKey.IMAGE_WINDOW, ImageWindowState.DOCKED)
         preferences.set_default(PrefKey.RECENT_FILES, [])
         preferences.set_default(PrefKey.LINE_NUMBERS, True)
         preferences.set_default(PrefKey.ORDINAL_NAMES, True)
@@ -324,6 +327,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(PrefKey.SEARCHDIALOG_WHOLE_WORD, False)
         preferences.set_default(PrefKey.SEARCHDIALOG_WRAP, True)
         preferences.set_default(PrefKey.SEARCHDIALOG_REGEX, False)
+        preferences.set_default(PrefKey.SEARCHDIALOG_MULTI_REPLACE, False)
         preferences.set_default(PrefKey.DIALOG_GEOMETRY, {})
         preferences.set_default(PrefKey.ROOT_GEOMETRY, "800x400")
         preferences.set_default(PrefKey.ROOT_GEOMETRY_STATE, "normal")
@@ -375,6 +379,12 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         preferences.set_default(PrefKey.SPELL_THRESHOLD, 3)
         preferences.set_default(PrefKey.UNMATCHED_NESTABLE, False)
+        preferences.set_default(
+            PrefKey.UNICODE_BLOCK, UnicodeBlockDialog.commonly_used_characters_name
+        )
+        preferences.set_default(PrefKey.UNICODE_SEARCH_HISTORY, [])
+        preferences.set_default(PrefKey.SPLIT_TEXT_WINDOW, False)
+        preferences.set_default(PrefKey.SPLIT_TEXT_SASH_COORD, 0)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -611,6 +621,10 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         unicode_menu = Menu(menu_tools, "~Unicode")
         unicode_menu.add_button(
+            "Unicode ~Search/Entry", UnicodeSearchDialog.show_dialog
+        )
+        unicode_menu.add_button("Unicode ~Blocks", UnicodeBlockDialog.show_dialog)
+        unicode_menu.add_button(
             "~Normalize Selected Characters",
             unicode_normalize,
         )
@@ -624,10 +638,16 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         """Create the View menu."""
         menu_view = Menu(menubar(), "~View")
         menu_view.add_checkbox(
+            "Split ~Text Window",
+            lambda: maintext().show_peer(),
+            lambda: maintext().hide_peer(),
+            root().split_text_window,
+        )
+        menu_view.add_checkbox(
             "~Dock Image",
             self.mainwindow.dock_image,
             self.mainwindow.float_image,
-            preferences.get(PrefKey.IMAGE_WINDOW) == "Docked",
+            root().image_window_state,
         )
         menu_view.add_button("~Show Image", self.show_image)
         menu_view.add_button("~Hide Image", self.hide_image)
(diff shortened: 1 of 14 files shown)


windymilla:
1. Difference from GG1: if illo/SN does not have a blank line after it, you can't move it. This situation should not happen during normal PPing.
2. Difference from GG1: If there are several blank lines, e.g. a chapter start, then if you are moving a SN/illo into it, it puts it at the bottom of the 4 blank lines, and if you are moving up, it puts it at the top. In GG2, if you are moving down into a 4 blank line space, it positions the SN/illo at the top of the 4 blank lines, and each successive attempt to move it, just moves 1 line down until it gets to the bottom, when it then moves on to the next available blank space. Or, to put it another way, if in the (unlikely) event that you want to move an illo past a chapter header, you have to do quite a lot of clicks in GG2 compared to GG1.
3. Just checking, moving is only currently implemented by clicking the buttons, not via Shift or Control clicks on the items in the dialog?

src/guiguts/application.py
```diff
@@ -581,6 +582,10 @@ def init_tools_menu(self) -> None:
         menu_tools.add_separator()
         menu_tools.add_button("~Page Separator Fixup...", page_separator_fixup)
         menu_tools.add_button("~Footnote Fixup...", footnote_check)
+        menu_tools.add_button("Sidenote Fixup...", lambda: illosn_check("Sidenote"))
```
windymilla, line 585: The new menu entries don't have keyboard shortcuts (using `~`)

src/guiguts/illo_sn_fixup.py
```diff
@@ -0,0 +1,578 @@
...
+        """
+        self.text = text
+        self.start = start
+        self.end = end
+        self.hilite_start = hilite_start
+        self.hilite_end = hilite_end
+
+
+class IlloSNChecker:
+    """Find, check & record footnotes."""
```
windymilla, line 49: Class docstring should also say "and sidenotes"

windymilla (changes requested):
Generally looks good - a few comments made.
I agree with the choice to have separate functions for up and down - it would involve too much complication in the code (and comments!) to try to combine these into one common function.

We should consider (not in this PR) whether the use of "system bookmarks" in GG1, so that the user can quickly jump between the current and previous position of the latest moved illo/SN, is something we should also provide in GG2, and if so, whether we want it to be exactly the same or if it could be improved upon.

windymilla:
Code is definitely shorter, and looks fine. Moving up and down generally appears to work fine.

I did find a couple of things that I'd like your opinion on please, @G4OEU 

I think my comment 2 previously is still the case:

> Difference from GG1: If there are several blank lines, e.g. a chapter start, then if you are moving a SN/illo into it, it puts it at the bottom of the 4 blank lines, and if you are moving up, it puts it at the top. In GG2, if you are moving down into a 4 blank line space, it positions the SN/illo at the top of the 4 blank lines, and each successive attempt to move it, just moves 1 line down until it gets to the bottom, when it then moves on to the next available blank space. Or, to put it another way, if in the (unlikely) event that you want to move an illo past a chapter header, you have to do quite a lot of clicks in GG2 compared to GG1.

Also, as mentioned before:

> The new menu entries don't have keyboard shortcuts (using ~)

Finally, I found that a mid paragraph sidenote is not listed as such in the SN fixup dialog. An example is line 7816 in the attached file: [projectID625088a2986be-sidenotes.txt](https://github.com/user-attachments/files/17062701/projectID625088a2986be-sidenotes.txt)
It is not reported as being mid-paragraph, and if I move that sidenote, it leaves a blank line behind.

G4OEU:
All the issues @windymilla raised, bar the keyboard shortcuts, have been fixed.

- moving the selected Illustration or Sidenote record UP/DOWN into a block of blank lines now moves it to the top/bottom of the block as for GG1.
- anomalous Illustration and Sidenote tags such as the mid-paragraph Sidenote at line 7816 are now flagged as 'MIDPARAGRAPH' too. This fix has identified and flagged a second anomalous mid-paragraph tag at line 8721.
- moving anomalous Illustration and Sidenote records closes up the line(s) they were moved from.

The advantage of using the UP/DOWN buttons to move the selected Illo and SN records is that it is all done one-handed via the mouse. You don' have to use the other hand to hold down keys AND click with the mouse on the selected tag with the other hand. It is also seems in keeping with my understanding of other tools in GG2 which use buttons rather than keyboard shortcuts to invoke their actions.

If keyboard shortcuts (Shift/Ctrl+Shift in GG1) are to be added as an alternative to the UP/DOWN buttons then can you please point me at another tool which I can use as a model to implement those keyboard shortcuts?

Once I've completed testing the fixes I'll issue another commit. This will be in another day or two.

G4OEU:
Oops! I think I misunderstood the "The new menu entries don't have keyboard shortcuts (using ~)" issue.

I'll fix this before the commit. It is a minor edit to application.py once I find two unused letters.

windymilla (approved):
Thanks - I'm happy with the code and the behavior now.

srjfoo:
Note that testing was done on macOS Sonoma.

Possibly as a future enhancement, may I suggest an "Undo" button? Page Sep Fixup has one. (For that matter, it might not be a bad idea for Footnote Fixup, too?)

I don't know if there's anything that can be done about this, but it's disconcerting for the spotlight on the illustration/sidenote tag to flash to the first tag in the file when "Move UP or "Move DOWN" is clicked, before settling back where it's supposed to be. Is there a way that the jump can be suppressed? 

My preference would be for UP and DOWN to just be Title Case -- words in all caps is the equivalent of shouting in terms of netiquette (particularly in a UI). This doesn't have an equivalent in GG1 that needs to be matched. And along those lines, is it likely to cause complaints that this doesn't behave as GG1 does?

Is there a reason for including a return before the tag in the results list? For instance, as an indication to the PPer that it's properly preceded by a blank line?

Question: The Sidenote Fixup is only Manual, there is no Automatic version, as in GG1?

---

I did not do exhaustive testing, looking for weird edge-cases (I don't think that it's needed, actually). The current code does do the job of moving the sidenote and illo tags up. as far as I can tell. I'm going to go ahead and approve the PR as-is, but I'd especially like to see thae bouncing around of the spotlighted tag resolved before we do another alpha release.

The preference for title case in the buttons is personal, and I'll defer to @windymilla and @G4OEU to sort that out. And also on whether an "Undo" button would be useful.

srjfoo (approved):

G4OEU:
Thank you @srjfoo for your feedback. I have fixed two bugs that you identified:

1. There is no longer a Return-character displayed before the text of the tag in the dialog;
2. There is no longer the disconcerting jump in the refreshed dialog between the highlighting of the first tag in the file and the highlighting of the selected tag after it has been moved. No tag is highlighted in the dialog now until one is selected by the user to be moved. After the move and refresh of the dialog, the highlighting remains on that same tag until the user selects another tag.

WRT the other issues you raised:

1. I have changed the button label text to use Title Case;
2. There are no current plans to add an 'Undo' button but that will be kept under review. I understand you have had some correspondence with @windymilla about 'Undo' buttons on GG2 applications;
3. There are no current plans to add automated Sidenote fixup as in GG1. The Manual Fixup for Sidenotes was a recent (2023) addition to GG1. It is considered more useful than Automatic Sidenote Fixup which moved the selected Sidenote _up_ to the first paragraph break (blank line), leaving behind the asterisk. It did this because an automatic procedure cannot be sure whether or not the Sidenote was mid-paragraph. Manual intervention was still needed after Automatic Sidenote Fixup completed since the user had then to delete the asterisk along with the line they occupied (when they are mid-paragraph), or delete them but keep the blank lines when paragraph breaks are needed. The comprehensive checking offered by the manual fixup tools in GG2 for Illustrations and Sidenotes makes it easier and safer to reposition them correctly.

srjfoo:
Thanks, @G4OEU!


> 1. There is no longer a Return-character displayed before the text of the tag in the dialog;

👍 

> 2. There is no longer the disconcerting jump in the refreshed dialog between the highlighting of the first tag in the file and the highlighting of the selected tag after it has been moved. No tag is highlighted in the dialog now until one is selected by the user to be moved. After the move and refresh of the dialog, the highlighting remains on that same tag until the user selects another tag.

This is a definite improvement, thanks.

Further comment about this, though: there are occasions where it's appropriate to move up or down more than one paragraph's-worth, and even without the visible jump, moving something takes quite a while (and only half the time, is the "working..." message showing). Until it's done, one cannot move the sidenote up or down another notch. There may be nothing that can be done to speed things up, but I wanted to mention it. (And if it is possible, it needn't be done in this PR.)

> WRT the other issues you raised:
> 
>> 1. I have changed the button label text to use Title Case;

👍 -- thanks

>> 2. There are no current plans to add an 'Undo' button but that will be kept under review. I understand you have had some correspondence with @windymilla about 'Undo' buttons on GG2 applications;

Yes, that's something that @windymilla and I have discussed. I mentioned it primarily because there *is* an undo for Page Separator Fixup.

>> 3. There are no current plans to add automated Sidenote fixup as in GG1. ...

That's fine with me. Again, I mentioned it because it's in GG1, and I know that @windymilla is concerned about keeping GG2 similar enough to GG1 that it won't be as much of a burden for existing users to switch to GG2 when the time comes. I have no problem at all with leaving out some of the GG1 functionality if y'all are okay with it. And given that if automatic were used, one would still need to go through it manually, I don't see much use for the automatic version.

If y'all are happy with it, let's go with it. Nothing that I've mentioned above needs to be done in this PR.

srjfoo (approved):

G4OEU:
Update to the 'Undo' button enhancement suggested by @srjfoo. As long as a a user knows how to 'undo' in the normal way with Ctrl+Z, they now have the the same facility as an 'Undo' button due to the following software change: All the inserts/deletes that happen during the move of a selected tag are now grouped so that they can be undone in one go with a single keyboard 'undo' (Ctrl+Z). @windymilla has observed that this improvement will mean that if the user moves a tag up 3 paragraphs with 3 consecutive clicks of the 'Move Selection Up' button, they would have to 'undo' 3 times to restore the file to its previous state. However he also notes that this is OK, because quite likely, they moved it 3, and then realised they only wanted it moved up 2 paragraphs, so a single undo will suffice to correct things to the desired state.

srjfoo:
> Update to the 'Undo' button enhancement suggested by @srjfoo. As long as a a user knows how to 'undo' in the normal way with Ctrl+Z, they now have the the same facility as an 'Undo' button due to the following software change: All the inserts/deletes that happen during the move of a selected tag are now grouped so that they can be undone in one go with a single keyboard 'undo' (Ctrl+Z). @windymilla has observed that this improvement will mean that if the user moves a tag up 3 paragraphs with 3 consecutive clicks of the 'Move Selection Up' button, they would have to 'undo' 3 times to restore the file to its previous state. However he also notes that this is OK, because quite likely, they moved it 3, and then realised they only wanted it moved up 2 paragraphs, so a single undo will suffice to correct things to the desired state.

Just a comment: The composite `ctrl`/`cmd`+`z` does work, but it doesn't restore page number or mid-para marking in the sidenote/illo check results unless `re-run` is done. That's probably okay, since I'm sure the multi-step undo also behaved that way, but should probably be documented? Should also be documented that the new composite undo using `ctrl`/`cmd`+`z` works only from the main text window, not from the fixup dialogs? That's true of undo in general, so probably not.

srjfoo:
Hm ... it just occurred to me that that may be a Mac-specific thing. I.e. that clicking on something in one of the fixup dialogs doesn't bring the main text window into focus, but leaves the focus in the dialog? Enough has happened since we last were looking at that that I've forgotten. If so, it's something that needs to be documented for Mac users, but not so much for others.

windymilla:
> the new composite undo using `ctrl`/`cmd`+`z` works only from the main text window, not from the fixup dialogs

I'll add a separate issue for that - as you say, it's unrelated to this work - probably the keystroke is being swallowed by the readonly text the messages are displayed in - https://github.com/DistributedProofreaders/guiguts-py/issues/446

windymilla:
> The composite `ctrl`/`cmd`+`z` does work, but it doesn't restore page number or mid-para marking in the sidenote/illo check results unless `re-run` is done.

If we decide this is a problem, then in a future PR, we would either have to add an Undo button which would do a regular "undo" followed by a re-run of the tool. Even without an Undo button, if we decide it's an issue, we could possibly override the meaning of Ctrl+z when focus is in the dialog to do the re-run. We'll leave it for now though.

windymilla:
> moving something takes quite a while (and only half the time, is the "working..." message showing)

I'm just looking at this with @G4OEU, who has a large sample file we can test with. The delay is probably just caused by the size of the file and the number of sidenotes/illos, since the dialog does a re-run every time you move a SN/illo.

If speed is too much of a problem, we can look at a way to refresh the dialog without a full re-run, since we know that more or less all that has happened is the start point of the SN/illo has moved up or down a few lines. This could also be relevant to the FN dialog, though the only feature that currently has is "join to previous", for which it just removes the message entry for the continuation part of the footnote. This would mean that the line numbers of SN/illos further down the file may be offset by 1 or 2 lines. However, I think I coded the checker dialogs so they cope with that and will still jump to the correct line even if the target has moved.

PR #427, opened 2024-09-09
Add soft wrapping

windymilla:
Soft Wrap: None/Char/Word in Prefs dialog.

Note that column select doesn't necessarily work well if you're trying to select stuff in the wrapped part. That's no surprise, given how column select works, and that column numbers won't line up if word-wrapped. I think it would need to be documented, or the whole idea abandoned - it's not obvious how to fix it.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -7,6 +7,7 @@ import logging
 import importlib.resources
 from importlib.metadata import version
 import os.path
+import tkinter as tk
 from tkinter import messagebox
 from typing import Optional
 import unicodedata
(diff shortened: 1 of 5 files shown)


windymilla:
Testing notes: This is really for you to test to see if it's worth having in its current form and/or what issues there are. It may be that it's not worth the complication and non-working features (like column select) that will ensue from having soft wrapping.

srjfoo:
It works as I'd expect, but it's questionable whether it would be useful to PPers. Both comments that I've seen so far have have been negative. I realize I suggested it as a possibility in the first place, but after thinking about it, I don't really think we need it.

PR #428, opened 2024-09-10
Trap python errors in `\C...\E` more neatly

### LINKED ISSUE #423: `\C bad python \E` causes traceback
It would be better to capture this and display the error more cleanly, since it's easy to make a typo when inserting the python code

windymilla:
Output the error in an error box and the message log without a full stack trace.

Fixes #423

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -3,6 +3,7 @@
 import logging
 import tkinter as tk
 from tkinter import ttk
+import traceback
 from typing import Any, Tuple, Optional
 
 import regex as re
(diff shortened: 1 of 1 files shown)


windymilla:
Testing comments:
1. Generate errors by putting invalid python in `\C...\E` replace string, e.g. `\Cthisisbadpython\E`, or `\C 1 / 0 \E` (divide by zero), or `\C x = 5 \E` (it's an assignment not an expression).
2. Things you can do (but probably shouldn't if you don't know what you're doing): `\Cexec("import math") or math.sqrt(2)\E` or `\C__import__('math').sqrt(2)\E` (import math module and get square root of 2. Try misspelling `sqrt` or `math`. 
3. Definitely shouldn't do without a safety belt: `\C__import__('subprocess').check_output('ls')\E` - executes an OS command (ls) and uses the results as the replace string.

srjfoo (approved):

PR #432, opened 2024-09-11
Fix highlight bug if footnote has no `:`

### LINKED ISSUE #431: Make one-line change to 'Footnote Fixup' code
There is a bug in 'footnotes.py' which has a simple, one-line, fix. It is in the block of code from lines 157-168 that finds the colon position in the tag, or uses end of the word "Footnote".

The bug doesn't manifest itself in 'Footnote Fixup' because all footnote tags that it processes have a ':' at some point after the string 'Footnote; that is, you should only see the forms '[Footnote 9: text.]' or '*[Footnote: text.]' when processing a file. This means the faulty code is never exercised.

However, when you use the same code in another module to parse an '[Illustration]' tag (note no ':' in the tag), the string 'Illustration' is not highlighted in the dialog because the calculation to find the end of the string 'Footnote' or 'Illustration', etc., is incorrect. The fix below corrects it. 

Current Code:
```
            start = beg_match.rowcol
            # Find colon position, or use end of the word "Footnote"
            colon_match = maintext().find_match(
                ":", IndexRange(start, maintext().rowcol(f"{start.row}.end"))
            )
            if colon_match is None:
                colon_pos = maintext().rowcol(
                    f"{start.index()}+{beg_match.count + 1}c wordend"
                )
            else:
                colon_pos = colon_match.rowcol
            end_point = colon_pos
```
Replace 8th line down (line 164) with
`                    f"{start.index()}+{beg_match.count}c"`

windymilla:
Too much was highlighted in this case - should never happen for footnotes, but could do theoretically.

Fixes #431

### DIFF
--- a/src/guiguts/footnotes.py
+++ b/src/guiguts/footnotes.py
@@ -160,9 +160,7 @@ class FootnoteChecker:
                 ":", IndexRange(start, maintext().rowcol(f"{start.row}.end"))
             )
             if colon_match is None:
-                colon_pos = maintext().rowcol(
-                    f"{start.index()}+{beg_match.count + 1}c wordend"
-                )
+                colon_pos = maintext().rowcol(f"{start.index()}+{beg_match.count}c")
             else:
                 colon_pos = colon_match.rowcol
             end_point = colon_pos


windymilla:
Testing notes: Remove the colon from a footnote, e.g. change
`[Footnote 7: seventh note]`
to
`[Footnote 7 seventh note]`
or
`[Footnote - seventh note]`
This PR should make just "Footnote" get highlighted when there isn't a colon.

windymilla:
On reflection, I'm closing this. After testing with and without this change, I think the code I had did what I intended, though when the same code was copied for use in the `[Illustration]` case (#426) it gave an undesirable result. More details in #431

PR #436, opened 2024-09-14
Try out split screen

windymilla:
I know this has wrong font, wrong, colors, no line numbers etc. This is just an experiment to see if it would be worth pursuing. Work so far is minimal - probably quite a bit more to get it to look smart and check what the bugs or limitations are, if any.
You can drag the split up & down, but obviously, it also needs an on/off option somewhere.
Splitting vertically would be possible instead, with a little more complication if you want the option of either a horizontal or vertical split.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -382,6 +382,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
         preferences.set_default(PrefKey.UNICODE_SEARCH_HISTORY, [])
         preferences.set_default(PrefKey.SPLIT_TEXT_WINDOW, False)
+        preferences.set_default(PrefKey.SPLIT_TEXT_SASH_COORD, 0)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -381,6 +381,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
             PrefKey.UNICODE_BLOCK, UnicodeBlockDialog.commonly_used_characters_name
         )
         preferences.set_default(PrefKey.UNICODE_SEARCH_HISTORY, [])
+        preferences.set_default(PrefKey.SPLIT_TEXT_WINDOW, False)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -629,6 +630,12 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
     def init_view_menu(self) -> None:
         """Create the View menu."""
         menu_view = Menu(menubar(), "~View")
+        menu_view.add_checkbox(
+            "Split ~Text Window",
+            lambda: maintext().show_peer(),
+            lambda: maintext().hide_peer(),
+            root().split_text_window,
+        )
         menu_view.add_checkbox(
             "~Dock Image",
             self.mainwindow.dock_image,
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -48,6 +48,7 @@ BINFILE_SUFFIX = ".json"
 BINFILE_KEY_MD5CHECKSUM: Final = "md5checksum"
 BINFILE_KEY_PAGEDETAILS: Final = "pagedetails"
 BINFILE_KEY_INSERTPOS: Final = "insertpos"
+BINFILE_KEY_INSERTPOSPEER: Final = "insertpospeer"
 BINFILE_KEY_IMAGEDIR: Final = "imagedir"
 BINFILE_KEY_PROJECTID: Final = "projectid"
 BINFILE_KEY_LANGUAGES: Final = "languages"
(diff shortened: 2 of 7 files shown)


tangledhelix:
A very quick test shows it works for me (although as you say, fonts, line numbers, etc). Dragging to resize is working, and updating text in one pane is faithfully reproduced in the other, which is good.

I know this is incomplete, but just in case you hadn't considered this already, you can add to the list of items to look at: if you are focused in the bottom pane and cmd-F to search, it will navigate you to the result in the top pane, which would be unexpected. Probably similar concern in other places, e.g. set/go-to bookmarks and other things, would need to be updated to be aware of which pane the user is focused in.

If you're taking input on the idea itself, I think this is something I would use and have wished for a few times in GG1. It might be worth asking in the forums to gauge interest before putting a ton of work into a finished version, though? I might be an outlier, after all.

tangledhelix:
A few further testing notes (Mac). This may all be expected, but noting it to add to the list of what would need implementation later.

- Navigation hotkeys aren't bound in the bottom pane currently (cmd-up/down for example)
- Some keys do work (e.g., cmd-A to select all)
- Status bar stuff doesn't update (cursor position, current character, selection)
- Right-click menu doesn't come up in bottom pane

tangledhelix:
Everything I wrote above is also true in Ubuntu.

windymilla:
Thanks @tangledhelix - all very helpful. I'll try moving the current hack into the proper place for it, where I hope it will be easier to make it look and behave more like it should without too much effort.

srjfoo:
Observations and questions.

* I have to tell this on myself. I brought it up, and looked in the View menu, then looked at the Settings pane ... and only _**then**_ noticed that it opened in split-screen by default. 🤣 -- how to turn it on and off has been discussed, but not implemented yet.
* When the bottom pane is selected, it has a highlight around the text area. When the top pane is selected, there's no obvious indication, except that the bottom pane **_isn't_** highlighted.
* Just looking, in BBEdit, the only indication which pane you're in is that in the active pane, the line where the cursor is, is highlighted. In GG2, the line numbers for the cursor for each pane is highlighted, but the one that is active is the only one with a blinking cursor. Bouncing back up here from talking about select all, below, it occurs to me that if there's a way to change the line number highlight from the OS highlight color to a darker/lighter grey (depending on theme) that might help visually differentiate which pane is currently active.
* You're right, no need to have an indication in the status bar about which pane is active. The row/column should just reflect the position of the cursor in the currently active pane.
* I don't know if there's anything you can do about this, but here's a screenshot of a comparison between the drag-bar for BBEdit (left) and GG2 (right). I had several tries before I was able to actually grab the GG2 one (not sure why -- it *looks* big enough, but apparently the trackpad precision isn't fine enough, even if it looks like it is).
image
* In BBEdit, if I select all (looking at @tangledhelix's comments) in one pane, the select all is applied to the whole document, but it's not apparent until you switch to the other pane, and the selection flashes before disappearing because you have to click in the pane, which automatically positions the cursor and cancels the selection. In GG2, if I select all in the top pane and switch to the bottom, the top pane stays highlighted. Same for bottom to top, but the highlighting in the former example uses and retains the Mac default highlight color (which is currently a pinkish color). In the latter case, when I switch from bottom to top, the highlight retained goes to grey.
* In both cases, if I select all and delete, the text disappears form both panes (which is as it should be). 
* In bottom pane, select a word and cmd/ctrl-f does not auto-fill. Not surprising, given @tangledhelix's comments above (which I have confirmed), but worth a mention as something to check after some additional debugging.

I think I'll stop here for now. It's looking really good!

windymilla:
Details of latest changes in commit message:

> 1. Add "Split Text Window" to the View menu - store in the Prefs file, etc.
> 2. Set highlight color and thickness so that there is a highlight border round the active of the two split windows in all themes.
> 3. Make line numbers update more reliably in peer window.
> 4. Track which of the windows last had focus, in preparation for future work like Search/Replace
> 5. Make splitter sash wider and add "handle" to drag with

@srjfoo's comment:
> Same for bottom to top, but the highlighting in the former example uses and retains the Mac default highlight color (which is currently a pinkish color). In the latter case, when I switch from bottom to top, the highlight retained goes to grey.

Right back when first implementing S/R, I made this change in the main text to force it to show the selection in the selected color (your pink) even when the text window doesn't have focus. Default behavior (as you see in the peer) is for the selection to be shown in "inactiveselect" color. On Macs (with your theme) this is gray. On Windows, no highlighting is shown when the widget loses focus (though it remembers what is selected and restores it when focus returns). This is what GG1 does, though in other apps, I think I've seen it gray, like you describe on macOS.
Shall I change it to do gray on the main text, and we'll see what that looks like? On macOS, I can let it do default behavior (which means it would be more likely to be the same gray as other apps use in similar situations), and on other platforms, I can pick darker/lighter gray depending on theme.

srjfoo:
> Shall I change it to do gray on the main text, and we'll see what that looks like? On macOS, I can let it do default behavior (which means it would be more likely to be the same gray as other apps use in similar situations), and on other platforms, I can pick darker/lighter gray depending on theme.

What I think makes the most sense to me is for the active pane of the active window to have the "active" highlighting, and for the inactive pane to either have the "inactiveselect", or, since you say Windows doesn't show anything, nothing. I think, though, that it's more important to get the other issues sorted first, and then deal with the highlighting?

srjfoo (approved):

Issue #144, opened 2024-02-18
Advanced Search features


windymilla:
1. Preserve page marker positions ☑
2. Multiple Replacement fields (like GG1)
3. Ability to execute arbitrary code (like `\C...\E`)☑
4. Other regex customizations from GG1, like `\G`, etc. ❌(not doing this at the moment) 
5. Show bad regexes in red as they are typed.☑
6. Incremental search?❌(not doing this at the moment) 
7. Handle multiline regexes - probably needs same treatment as in GG1☑
8. Catch bad regexes if user tries to use them (despite 5.)☑

windymilla:
See also #266 - may be OK if point 7 above is completed.

windymilla:
1, 3 & 7 are done so far

windymilla:
5 and 8 will be addressed by #414

PR #439, opened 2024-09-20
Implement multi-replace

windymilla:
Allow user to toggle just one replace box & buttons or 3 replace boxes & buttons, using checkbox in the S/R dialog.

Dialog resizes to take account of additional buttons.


windymilla:
Testing notes: 
1. Pop S/R dialog
2. Check Multi-replace checkbox
3. Type different replacement strings in the different boxes
4. Check that Repl, Repl & Search, and Repl All buttons use the replacements from the correct box.
5. Uncheck multi-replace to check extra boxes/buttons go away
6. Try with `--nohome` to see start up situation

I've also tried to improve the way the box is laid out so that it copes better with the 3 different themes.

windymilla:
This fixes the last feature from #144 that will be implemented at the moment.

srjfoo (approved):

PR #440, opened 2024-09-21
Hide Suspects only when not relevant in WF

windymilla:
Some modes don't have Suspects, so the "Suspects Only" checkbox was confusing.


windymilla:
Testing notes:
I first tried disabling the Suspects Only checkbox, but that means you can end up with a disabled checkbox that is checked for All Words, for example. This is still confusing because you might think it is only showing suspects, but you can't turn it off because it is disabled. It therefore makes more sense to hide the checkbox for the irrelevant modes.

srjfoo (approved):

PR #442, opened 2024-09-21
Return focus to main text when peer is closed

windymilla:
If focus was in peer, then a dialog was popped, then the Split Screen was turned off, the peer was hidden, but was still the "active" window, causing odd behavior. Not sets the focus, and hence active, window to be maintext.


windymilla:
Testing notes:
1. Turn on Split Window
2. Put focus in peer
3. Pop a dialog (e.g. a tool, or S/R)
4. Turn off Split Window (but without clicking in the peer or main text)
5. Continue to use the dialog, and it will work on the phantom peer window, e.g. searches, or clicking on a tool message.

srjfoo (approved):

PR #456, opened 2024-09-27
Improve image viewer

windymilla:
1. Zoom in & out
2. Fit to image width/height
3. Invert checkbox reverses colors in image
4. Set background of canvas to button background color
5. Use better scaling algorithm to improve image quality
6. Remove lots of complication


srjfoo (approved):
This is a pretty awesome improvement. I'm sure that there are other tweaks that can be done in the future, but I think it's great as-is.

PR #461, opened 2024-09-29
Make undo/redo work from checker dialogs

### LINKED ISSUE #446: Keystrokes being ignored from dialogs
Keystrokes like `Ctrl+s` for save are supposed to work even when the focus is in a dialog.
`Ctrl+z` doesn't work when focus is in a checker dialog.

Check out other dialogs/keystrokes.

windymilla:
Redirect the undo/redo events from the read-only text list in checker dialogs to the main text window.

Fixes #446

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -564,9 +564,16 @@ class ScrolledReadOnlyText(tk.Text):
         # Also on creation, so it's correct for the current theme
         theme_set_tk_widget_colors(self)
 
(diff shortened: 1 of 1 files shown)


windymilla:
NTS: Redo not working on macOS - switch to binding keystroke instead of (or in addition to) redirecting event?

srjfoo (approved):
I did not test every tool, but I did test several, and undo/redo worked for all of them (noting here that this was a Mac-specific check; the original version apparently worked as originally coded for Windows).

PR #462, opened 2024-09-29
Add "Save a Copy As" menu option

### LINKED ISSUE #459: Add 'Save a Copy As' to File Menu
'Save a Copy As' is used by PPers who work with GG1 and they would expect to find it in GG2.

Notepad++ is another text editor that has the same command.

windymilla:
Saves a copy of the current file, but does not change the current filename, nor the "edited" flag.
Name of saved copy does get inserted in the "Recent Documents" menu.

Fixes #459

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -426,6 +426,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         self.menu_file.add_button(
             "Save ~As...", self.file.save_as_file, "Cmd/Ctrl+Shift+S"
         )
+        self.menu_file.add_button("Sa~ve a Copy As...", self.file.save_copy_as_file)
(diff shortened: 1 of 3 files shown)


srjfoo (approved):
Going ahead and approving this, as it works as advertised. My main concern, though, is that I'm not sure that it's appropriate for it to be saved in the `r1` position. If I'm doing a "save a copy as", I'm not likely to want to open it as `r1` the next time I open GG2. 

That's just my opinion, though. In the long run, if we wind up with an encapsulated install, where users don't invoke gg2 from the command line, position in the list won't matter, just that it be in the list.

windymilla:
It's the same as GG1, but you could definitely make an argument for putting it in the r2 position

srjfoo (approved):

PR #493, opened 2024-10-08
Ensure all features are accessible through menus

### LINKED ISSUE #474: Ensure the only way to access a feature isn't by some obscure click in the status bar
One current example is Configure Page Labels - it should go in a menu, even if it's an relatively hidden one like File-->Project.
Also check other status bar clicks to make sure there aren't any others. Having them in menus helps people find them, and also helps documentation.

windymilla:
Some were previously only accessible through clicking on status bar buttons.

Fixes #474

See #474 for list of changes

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -54,7 +54,7 @@ from guiguts.misc_tools import (
 )
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
-from guiguts.root import root, ImageWindowState
+from guiguts.root import root
 from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
 from guiguts.tools.pptxt import pptxt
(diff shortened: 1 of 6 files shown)


srjfoo:
Everything worked as it should, but I have a few comments/questions. Particularly item 5, which I didn't understand 😐 

1. Go to line --> Search menu? ✅ (General comment about the Goto options -- should there be a Goto submenu? Without keyboard shortcuts, using the menu items for some of these is going to be a real slog, so I imagine people will be more likely to use the status bar options. Having a submenu will force those who do use the menu down another level, but will shorten the menu for general use)

2. Toggle line numbers --> Prefs? ✅

3. Go to image --> Search menu? ✅ (General comment: given three-digit pngs, both 44 and 044 work, but 044.png does not. Should it? Also, having two "Go to Page" title bars is somewhat confusing, especially when they mean different things. Should this one (and the item in the menu) actually say "Go to Image"?)

4. Go to prev image --> Search menu? ✅

5. Wording of See Image (tooltip)/Show Image (menu button) -- not sure what this means?

6. Toggle Auto Image - move from Prefs to View menu? ✅

7. Go to next image --> Search menu? ✅

8. Go to page (label) --> Search menu? ✅

9. Configure page labels --> File/Project menu? ✅

10. Restore previous selection --> Edit menu? ✅ (note that this does not remember if the previous selection was a column selection -- should it?)

11. Toggle column/regular selection --> Edit menu? ✅

12. Set language(s) --> File/Project menu? ✅

13. Toggle ordinal/name display --> Prefs? ✅

srjfoo (approved):

PR #503, opened 2024-10-24
Full functionality 'Footnote Fixup' tool

G4OEU:
Does not include inline footnotes which will have a separate tool if that is required.

Implements the fast 2-pass version of 'Move FNs to Paragraphs' which uses marks.

Fixes bug caused by interaction with use of marks when last line of the file is a footnote.

Uses the elegant GG1 alpha() function to convert footnote labels to alphabetic form.

Unlike GG1 it correctly handles 'mid-paragraph' footnotes; that is, footnotes at the bottom of the page that are anchored in the paragraph above which continues on to the following page(s).

Footnote Fixup will work consistently before or after Page Marker removal.

### DIFF
--- a/src/guiguts/footnotes.py
+++ b/src/guiguts/footnotes.py
@@ -299,7 +299,7 @@ class FootnoteChecker:
 
         An issue arises if the last line of the file is a footnote.
         Inserting a blank line ("\n") programatically will place it
-        *before* the footnote's end "Checker" mark if tk.RIGHT gravity
+        before the footnote's end "Checker" mark if tk.RIGHT gravity
         was specified when the mark was set. We want the blank line
         placed after the mark otherwise the "\n" will be carried with
         the footnote if it is moved.
--- a/src/guiguts/footnotes.py
+++ b/src/guiguts/footnotes.py
@@ -1,13 +1,19 @@
 """Footnote checking, fixing and tidying functionality"""
 
 import logging
+from enum import StrEnum, auto
 from tkinter import ttk
 from typing import Optional
 import regex as re
+import roman  # type: ignore[import-untyped]
 
 from guiguts.checkers import CheckerDialog, CheckerEntry
 from guiguts.maintext import maintext
 from guiguts.misc_tools import tool_save
+from guiguts.preferences import (
+    PersistentString,
+    PrefKey,
+)
 from guiguts.utilities import (
     IndexRowCol,
     IndexRange,
@@ -19,6 +25,14 @@ logger = logging.getLogger(__package__)
 _the_footnote_checker: Optional["FootnoteChecker"] = (
     None  # pylint: disable=invalid-name
 )
+
+
+class FootnoteIndexStyle(StrEnum):
+    """Enum class to store FN index style types."""
+
+    NUMBER = auto()
+    LETTER = auto()
+    ROMAN = auto()
 
 
 class AnchorRecord:
@@ -89,6 +103,9 @@ class FootnoteChecker:
         self.fn_records: list[FootnoteRecord] = []
         self.an_records: list[AnchorRecord] = []
         self.checker_dialog: CheckerDialog = checker_dialog
+        self.fn_index_style = PersistentString(PrefKey.FOOTNOTE_INDEX_STYLE)
+        self.fn_check_errors = False
+        self.fns_have_been_moved = False
 
     def reset(self) -> None:
         """Reset FootnoteChecker."""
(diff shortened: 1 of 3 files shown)


srjfoo:
With a feature as complex as this, I like to test against several source files, and I'm not quite finished yet, but have a couple of comments.

I was surprised that even with chapter landing zones, I was required to re-index, so I went back and checked GG1, and it does not require re-indexing for chapter landing zones.

---

Also, based on the widget, GG1 also allows the PPer to customize individual footnotes -- they don't all have to be the same kind (numeric, alpha, Roman), which appears to be the case here.

windymilla:
> I was surprised that even with chapter landing zones, I was required to re-index, so I went back and checked GG1, and it does not require re-indexing for chapter landing zones.

I don't think I've worked on a book that needs separate FN numbering for each chapter, but @G4OEU has a wider range of PPing experience than me, and @srjfoo sees a lot of different books when marking books as posted & other tasks. If you two think it's a necessary feature, to permit the proofed FN numbers to remain, instead of reindexing, I'd be happy with that.

> Also, based on the widget, GG1 also allows the PPer to customize individual footnotes -- they don't all have to be the same kind (numeric, alpha, Roman), which appears to be the case here.

@G4OEU & I had a conversation about that feature during the development of Footnote Fixup. My opinion was that it would be fine for us not to have it, at least to begin with. Since it's a rare occurrence to need it, and you can get round it with a bit of manual S/R to just work on one style of footnotes at a time, maybe we could wait and see if there's a call for it?

srjfoo:
> > I was surprised that even with chapter landing zones, I was required to re-index, so I went back and checked GG1, and it does not require re-indexing for chapter landing zones.
> 
> I don't think I've worked on a book that needs separate FN numbering for each chapter, but @G4OEU has a wider range of PPing experience than me, and @srjfoo sees a lot of different books when marking books as posted & other tasks. If you two think it's a necessary feature, to permit the proofed FN numbers to remain, instead of reindexing, I'd be happy with that.

I'd like to see input from PPers who do that sort of project. 

I'd actually like to see the ability to reindex by chapter, so that if the PPer desires, the numbering can start over at 1 by chapter, for the sources where the numbering restarts by page, but I don't think that's any more useful than just reindexing the whole document. If I were actively PPing, I'd think that I'd want to be able to match the scan where it's possible to do so, though, so not forcing full-document reindexing if the renumbering starts by chapter in the original makes sense to me.

> 
> > Also, based on the widget, GG1 also allows the PPer to customize individual footnotes -- they don't all have to be the same kind (numeric, alpha, Roman), which appears to be the case here.
> 
> @G4OEU & I had a conversation about that feature during the development of Footnote Fixup. My opinion was that it would be fine for us not to have it, at least to begin with. Since it's a rare occurrence to need it, and you can get round it with a bit of manual S/R to just work on one style of footnotes at a time, maybe we could wait and see if there's a call for it?

I'm okay with that approach.

srjfoo:
After multiple distractions, finally getting back to this. Looking at the screenshot below, notice a couple of things:

image

* The "Re-run" button is off to the right all by itself. None of the other buttons extends below the re-run button.
* The longer button labels may be clipped, depending on how wide the results window is. This is just an fyi as something to keep an eye on until the dialog settles into a final form.

windymilla:
> * The "Re-run" button is off to the right all by itself. None of the other buttons extends below the re-run button.
> * The longer button labels may be clipped, depending on how wide the results window is. This is just an fyi as something to keep an eye on until the dialog settles into a final form.

I have a couple of ideas to help with these, unless you've already tackled them @G4OEU 
1. Since the rearrangement of the old multiple frames version, the Frame `frame` now only has one item inside it: `fixit_frame`. So you could get rid of `frame` and make `fixit_frame` a direct child of `header_frame` instead:
`fixit_frame = ttk.Frame(self.checker_dialog.header_frame)`
2. To make that frame extend underneath the Re-run button, you can add `columnspan=2` to its `grid` command (Note to self: I should really fix checkers.py, so that we don't need to add that to every new checker dialog we create, and get the padding sorted so it aligns with the RH edge of the Re-run button without needing to add extra padding every time - issue added.)
3. To make all the buttons get an equal proportion of any stretch or squeeze of the dialog width (and hence minimise the clipping @srjfoo saw) set the `weight=1` for all 4 columns - there are some near duplicate calls to `column_configure` and at least one of them sets the weight to zero. The duplicates can be removed - weight doesn't need setting before each button is added, just once for each column, not every row of every column.

G4OEU:
Have just committed changes and pushed them to origin but am getting some sort of failure message from GitHub (see above).

The list of changes in that commit omitted to say that buttons now extend under the 'Re-run' button and that, as part of that fix, 'weight=1' is set just once for the four button columns. The duplicated code that set weights has been removed.

To add some detail about the rare reindexing bug, I've modified the reindex() function to set temporary marks for the hilite start/end columns. They are removed at the end of the function. This fixed a rare bug that could occur when there are two or more anchors on the same line. If the label of an earlier anchor is lengthened on a line during reindexing then you cannot rely on the hilite start/end columns in the records of other anchors on that same line. I had dealt with that possibility and it worked fine except for a contrived example of anchor/footnote labelling which highlighted a bug. Using marks fixed the issue.

windymilla (approved):
I think all the points I'm aware of have now been addressed, either via changes to the code, or in new issues (#515, #526, #527)

src/guiguts/footnotes.py
```diff
@@ -122,28 +143,831 @@ def join_to_previous(self) -> None:
...
+        # Get index into dialog entries for this record.
+        dialog_entry_index = self.map_fn_record_to_dialog_index(prev_record)
+        if dialog_entry_index < 0:
+            logger.error("Unexpected return from 'join_to_previous()'")
+            return
+        # Make it the selected entry in the dialog.
+        self.checker_dialog.select_entry_by_index(dialog_entry_index)
+
+    def autoset_chapter_lz(self) -> None:
+        """Insert a 'FOOTNOTES:' LZ header line before each chapter break.
```
srjfoo, line 163: Discrepancy between GG2 and GG1 (moving footnotes to chapters).

In the first screenshot. Notice the position of fn 1 in the text file (split screen is irrelevant in this screenshot, ignore fn 2).

fn-lz1

GG1 does not place an LZ following the chapter title, but above the following chapter header.


I'm not sure what the difference is (I may need to look at the file more closely later), but notice that in this second screenshot, the footnote anchor is in the same place, but in this case, the footnote is more correctly moved to the end of the chapter:

image

---

A couple of comments/questions.

1. I understand that GG1 places the landing zones roughly the same way, but to me, looking at where the landing zones are placed, they're not where I would expect them to be. That is, yes, they're placed before the chapter header by four blank lines, but, if the chapters all start on a new page, that means that the footnotes for one chapter are actually on the same page as the beginning of the following chapter. Which seems to me to be problematical if a file has page numbers associated with it. Am I missing something that's a part of PP'ing?
2. I expected "Tidy Footnotes" to remove the unwanted "FOOTNOTES:" tags. It didn't for me in GG2. I checked GG1, and it did. Is this a Mac issue, or is it again something I should be doing that I'm not aware of?

srjfoo: > A couple of comments/questions.
> 
> 1. I understand that GG1 places the landing zones roughly the same way, but to me, looking at where the landing zones are placed, they're not where I would expect them to be. That is, yes, they're placed before the chapter header by four blank lines, but, if the chapters all start on a new page, that means that the footnotes for one chapter are actually on the same page as the beginning of the following chapter. Which seems to me to be problematical if a file has page numbers associated with it. Am I missing something that's a part of PP'ing?
> 
> 2. I expected "Tidy Footnotes" to remove the unwanted "FOOTNOTES:" tags. It didn't for me in GG2. I checked GG1, and it did. Is this a Mac issue, or is it again something I should be doing that I'm not aware of?

After touching base with @windymilla, I learned a couple of things:

* Regarding the unused LZ removal, that should happen when the footnotes are moved to LZs.

* Tidy Footnotes is only supposed to be used on the text file, after Text and HTML are split.

So, I was indeed misunderstanding. However, the unused LZs should be removed.

G4OEU: Re. removing unused LZ headers, the current (and new) Guiguts Manual agrees with you and reminds PPers that:

"If you used **Autoset Chap. LZ**, there may be unused Landing Zones at the ends of chapters with no footnotes. (Recall that Landing Zones are the word FOOTNOTES:). You probably will want to delete the unused Landing Zones. Similarly, if you moved all of the footnotes to the end of the document and later moved them to follow the paragraphs that referenced them, remember to delete FOOTNOTES: at the end of the document."

G4OEU:
@srjfoo has pointed out that in the case of moving FNs to a LZ at the end of each chapter that "yes, they're placed before the chapter header by four blank lines, but, if the chapters all start on a new page, that means that the footnotes for one chapter are actually on the same page as the beginning of the following chapter".

This should be fixed in the same way that is done when FNs are move to paragraph ends. In the latter case care is taken to ensure that where a paragraph ends at the bottom of a page, any FNs anchored in that paragraph are inserted before the page break so that they are placed on the same page as the paragraph.

windymilla:
> @srjfoo has pointed out that in the case of moving FNs to a LZ at the end of each chapter that "yes, they're placed before the chapter header by four blank lines, but, if the chapters all start on a new page, that means that the footnotes for one chapter are actually on the same page as the beginning of the following chapter".
> 
> This should be fixed in the same way that is done when FNs are move to paragraph ends. In the latter case care is taken to ensure that, where a paragraph ends at the bottom of a page, any FNs anchored in that paragraph are inserted before the page break so that they are placed on the same page as the paragraph.

From my brief trial, if we did that, it would be an improvement on GG1's behavior.

src/guiguts/footnotes.py
G4OEU: @srjfoo and you are right about this. I'll add that fix along with the other issue that @srjfoo identified.

srjfoo: @G4OEU -- many thanks. I just want to let you know that your effort is very much appreciated! Filling out Footnote Fixup is a major undertaking!

It's probably both an advantage and a disadvantage as a tester, that my perspective is that of someone who has done some post-processing, long ago, but not enough to remember the whole procedure and what did and didn't work in GG1 in the days of 0.2.4. I'll let @windymilla pass judgment on that (😁). It does encourage me to ask questions when something doesn't make sense to me, though!

windymilla (approved):
Looks good to me now - thanks!

src/guiguts/footnotes.py
```diff
@@ -122,28 +143,957 @@ def join_to_previous(self) -> None:
...
+
+        Before inserting the first LZ header, skip the first 200 lines of the
+        file if it is longer than 200 lines. This avoids inserting LZ headers
+        at pseudo paragraph breaks in front matter pages.
```
srjfoo, line 167: This is true of most projects with footnotes, but I feel pretty uncomfortable with this as a hard threshold (even if it's also true of GG1), without some way for those with projects where the footnotes start before that threshold to deal with it easily. One of the test projects I've downloaded has the first footnote anchor before line 200, even though the footnote itself is after; another has two footnotes (both anchor and footnote, before line 200). I browsed the PP pool, and found another that had first footnote and anchor preceding line 200.

windymilla: Since this is the same as GG1, I'd be happy for it to be merged with the 200 threshold. However, it occurs to me that we could maybe improve on GG1's algorithm by starting chapter LZs after the location of the first footnote, rather than after a fixed line number. As I said, happy for that to go into future footnote improvements, if not in this PR.

G4OEU: This should be simple to fix. I will add to this PR the @windymilla suggestion of starting chapter LZs after the first footnote rather than after the first 200 lines. Thank you @srjfoo for highlighting the issue.

srjfoo (changes requested):
I think this can be classified as an off-by-one problem, maybe. What appears to be happening is that the chapter LZs are being placed one line above the last line of the previous chapter. I've only checked a project with chapters that start on a new page, so I don't know what happens for chapters with mid-page chapter heads.

image

srjfoo:
Update -- I was able to quickly find a project that had a handful of footnotes, and chapters mid-page. The anomaly shown above did not happen for chapters mid-page.

srjfoo (approved):
It looks good to me, with one caveat that I haven't checked out fully, and have only seen the effect on chapter LZs.

The book that I tested this on had a Book 1 and a Book 2, with a half dozen-ish chapters in each book. If I added page markers before setting the chapter LZs, I got one LZ below the Book 2 header and one at the end of the book (which I think is automatic?). I took the page markers out (after backing out the LZs) and re-ran it and got the correct placement.

I think we should go ahead and get this PR in and then tag the issue (if y'all would, please, test and see if you get the same results, and I'll re-test later, as well).

windymilla:
> The book that I tested this on had a Book 1 and a Book 2, with a half dozen-ish chapters in each book. If I added page markers before setting the chapter LZs, I got one LZ below the Book 2 header and one at the end of the book (which I think is automatic?). I took the page markers out (after backing out the LZs) and re-ran it and got the correct placement.

Assuming by "added page markers", you mean File-->Project-->Add Page Marker Flags, then I don't think we need to do anything, apart from document in the manual that you shouldn't try to do anything (significant) when Page Marker Flags are visible. (Apart from expert usage) they are just for transfer to GG1 and back. I'll add a sentence to the manual now.

srjfoo:
> > The book that I tested this on had a Book 1 and a Book 2, with a half dozen-ish chapters in each book. If I added page markers before setting the chapter LZs, I got one LZ below the Book 2 header and one at the end of the book (which I think is automatic?). I took the page markers out (after backing out the LZs) and re-ran it and got the correct placement.
> 
> Assuming by "added page markers", you mean File-->Project-->Add Page Marker Flags, then I don't think we need to do anything, apart from document in the manual that you shouldn't try to do anything (significant) when Page Marker Flags are visible. (Apart from expert usage) they are just for transfer to GG1 and back. I'll add a sentence to the manual now. ETA: Done

Yes, that's what I meant, thanks. And, of course, I had what I thought was a good reason for having them visible, but, as you say, it was not what I'd consider a normal situation.

PR #504, opened 2024-10-24
Add Stealth Scannos feature

windymilla:
Two scannos files are included with the release:
regex.json and en-common.json, which are based
on the GG1 versions.

The user can also load their own scannos file. The recently loaded scannos files are in a dropdown at the top of the dialog.

The first scanno is shown, and the user can click Next or Prev. If there are no occurrences of the next/prev scanno, it autoadvances until it finds a scanno that does occur, or it hits the end/start of the scannos file.

All occurrences of the scanno are listed, and the
replace button (or Cmd/Ctrl left mouse click) will do the replacement if one is specified in the scannos file.

Scannos can also have an optional hint to explain what they do, which is displayed in the dialog.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -53,6 +53,10 @@ from guiguts.misc_tools import (
     proofer_comment_check,
     asterisk_check,
     TextMarkupConvertorDialog,
+    stealth_scannos,
+    DEFAULT_SCANNOS_DIR,
+    DEFAULT_REGEX_SCANNOS,
+    DEFAULT_STEALTH_SCANNOS,
 )
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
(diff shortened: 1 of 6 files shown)


srjfoo (approved):

srjfoo:
Random comments from testing. Overall looks good, these are just questions from a tester who is a relatively inexperienced practical user of GG.

* I don't know if this is relevant or not, but "Find an opening square bracket followed by anything but FSIG (or a digit)" also picks up footnote anchors that were formatted as a letter, and Music markup.

* Is this something that should be considered? I.e. more than two initials with spaces between?
image

* I don't suppose the hints can be wrapped?

* > Find a comma followed by a space and an upper-case character (except I) and replace with a period.

Finds things such as "I'm", "I've", "I'd"

* Should the "Next" and "Previous" buttons be visually disabled when one is at the first or last scanno regex?

* For what it's worth, two of the regexes had `<tb>` as a result. Probably doesn't matter, because I guess I'm assuming that thought breaks will have been resolved by the point in time where the scanno check is done.

---

Oh -- what about cmd up/down and home/end for navigating the lists?

windymilla:
> Random comments from testing.

Thanks - not changing the operation of specific regexes atm, just replicating the behavior of GG1. I would be very happy if an experienced PPer were to go through all the regexes and remove/edit those that are not as useful as they could be.

PR #507, opened 2024-10-25
Add "Convert text markup" feature

### LINKED ISSUE #465: Add Text Markup Conversion Dialog
This is to convert `<i>` to `_`, etc., as done in GG1

windymilla:
In new Text menu, button pops a dialog from where the user can convert `<i>`/`</i>`, etc. to the text file markup characters, e.g. `_`.
These characters are configurable and are stored in the Prefs file.
Can also convert small caps to ALLCAPS, and convert
thought breaks from `<tb>` to `       *       *...`

Fixes #465


src/guiguts/misc_tools.py
```diff
@@ -1224,3 +1230,111 @@ def asterisk_check() -> None:
             line, IndexRange(match.rowcol, end_rowcol), match.rowcol.col, end_rowcol.col
         )
     checker_dialog.display_entries()
+
```
srjfoo, line 1233: Extra blank line?

srjfoo (approved):
Works as advertised -- my only question is whether Undo should work if the widget is in the foreground. It doesn't on macOS; I have to switch back to the main text window.

src/guiguts/misc_tools.py
windymilla: No - the code prettifier forces 2 blank lines before the start of a new class (TextMarkupConvertorDialog), and 1 line between functions.

windymilla:
> Works as advertised -- my only question is whether Undo should work if the widget is in the foreground. It doesn't on macOS; I have to switch back to the main text window.

I've added an [issue ](https://github.com/DistributedProofreaders/guiguts-py/issues/511) to consider this problem more generally - I don't think it would be right to just make it work for this particular dialog without further consideration. (Summary: it currently works in checker dialogs and Page Sep Fixup, which has undo/redo buttons)

PR #508, opened 2024-10-25
Make S/R fields tall enough

### LINKED ISSUE #506: Subscript not showing in S/R dialog fields
Possibly depends font size. Stop the fields from growing & shrinking with the dialog.

windymilla:
Fixes #506

If font size was too large, and S/R dialog too short, then underscores wouldn't fit in the field because the fields resized vertically with the dialog. Now they don't.

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -3,6 +3,7 @@
 import logging
 import tkinter as tk
 from tkinter import ttk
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
With the latest commits, everything looks good as far as I can tell. Both in an instance using the home directory, or using `--nohome`.

PR #517, opened 2024-11-07
Convert to curly quotes & Check curly quotes

windymilla:
Convert to curly quotes tries to convert all straight quotes to curly where it can be confident of which curly is needed. This normally leaves unconverted all those that should be open single curly, but  does a good job of most others. This is also
true with the equivalent GG1 and ppsmq tools.

Check curly quotes pops a dialog listing suspected quote errors such as quote at start/end of line or after/before space when curl is the wrong direction, or open quote preceded by word character, or paragraph end without closing double quotes (unless next para begins with double quotes), or unconverted straight quotes.

Dialog has several buttons that make corrections, and also supports Cmd/Ctrl+left/right click to make best guess fix, which is to convert to open curly if quote is straight, or to swap open & close if quote is curly. 

After Convert to Curly Quotes is run, the Check is run automatically. However, unlike GG1, the conversion to curly quotes does not need to be done with the convert feature for the check to work. So if the user makes manual quote changes at any point they can just re-run the check.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -57,6 +57,7 @@ from guiguts.misc_tools import (
     DEFAULT_SCANNOS_DIR,
     DEFAULT_REGEX_SCANNOS,
     DEFAULT_STEALTH_SCANNOS,
+    convert_to_curly_quotes,
 )
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
@@ -682,6 +683,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_tools.add_button("R~ewrap Selection", self.file.rewrap_selection)
         menu_tools.add_button("C~lean Up Rewrap Markers", self.file.rewrap_cleanup)
         menu_tools.add_separator()
+        menu_tools.add_button("Convert to Curly ~Quotes", convert_to_curly_quotes)
         unmatched_menu = Menu(menu_tools, "Un~matched")
         unmatched_menu.add_button("Bloc~k Markup...", unmatched_block_markup)
         unmatched_menu.add_button("~DP Markup...", unmatched_dp_markup)
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -58,6 +58,7 @@ from guiguts.misc_tools import (
     DEFAULT_REGEX_SCANNOS,
     DEFAULT_STEALTH_SCANNOS,
     convert_to_curly_quotes,
+    check_curly_quotes,
 )
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey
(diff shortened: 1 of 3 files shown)


srjfoo:
A couple of questions:
* Is there a reason not to ignore straight quotes within proofers' notes?
* Is there some way for the user to flag continuation quotes so that they don't keep coming up when re-running the check?

---
The following was identified as "double quote not closed", probably because the close quote was within no-wrap? (the one following the single not converted was the "double close quote unexpected" to match the "double quote not closed") The open single wasn't converted, either. Was that because of the no-wrap, or something else. I've seen several cases of singles not being converted. Or is it because of the mid-line apostrophe?

image

---

(Apologies for the quality of the dialog -- not my usual fare, but I figured fiction was the most likely to have conversation that needed conversion of straight quotes to curly.) Ahem. Anyway ... another example of the quote within the proofers' note having an effect on the surrounding text. In this case, P2 correctly identified a missing double quote in a note. But as far as the results are concerned, the suspect message was "Double open quote followed by space". There isn't a message about a double close quote not expected, because the quote within the note was converted and balanced the one at the end of the paragraph.

image

---

So, I guess the question is -- is the quote conversion something that should not take place until after the no-wrap markup and proofers' notes are removed? If that's the case, then I think my notes above can be ignored, and I'll re-test after removing the notes and the no-wrap. I probably should have read the manual about converting the quotes, before testing. 😁 .

windymilla:
> * Is there a reason not to ignore straight quotes within proofers' notes?

I think trying to second guess how to deal with things within proofers notes is fraught with danger. It's very easy to construct plausible situations that either reveal or hide errors depending on what the proofer has written. My argument would be that if a proofer's note causes an unwanted warning - that's a good hint to the PPer that they need to deal with the note. If something in a proofer's note hides what would otherwise be a warning, then the PPer will discover it anyway when they deal with proofers notes. In general, I think it's better to deal with most/all proofers notes before running tests & checks - if the proofers notes are valid, e.g. pointing out a typo, then fixing them will reduce the number of warnings you get from a spell check. If the proofer is wrong, then you're going to just delete their note anyway. So, in summary, I don't think we want checks to try to detect/allow for proofers notes. AFAIK, GG1 doesn't, and I don't recall any PPers suggesting that be changed.  

> * Is there some way for the user to flag continuation quotes so that they don't keep coming up when re-running the check?

No. The check does not report the following because although the first para doesn't have close quotes, then very next line has open quotes due to continued speech:
```
“First paragraph

“continuation.”
```
Or is it some other case you are asking about?

> The following was identified as "double quote not closed", probably because the close quote was within no-wrap?

No, it's not really to do with no-wrap - it's because it's a separate paragraph as far as GG is concerned (due to the blank line). I think it's a rare enough case that it's OK to have a false positive there.

> So, I guess the question is -- is the quote conversion something that should not take place until after the no-wrap markup and proofers' notes are removed?

Personally, I think so - as mentioned above, I think it's better to deal with them as soon as possible. You then have the text as you think it should be, rather than with things that you know are errors and random proofer comments about fixing them. At that point, IMHO, you can do all the checking, formatting, converting, wrapping, etc.

srjfoo:
The `Quote<=>Space` button doesn't seem to work for me. That is, it doesn't convert either a straight double-quote or single or a curly to a space. Also, not sure what the `Delete Space` button is supposed to do? I tried doing a cross-check with GG1, and discovered that the way the smart quote checks work is so vastly different that I didn't think it was a good comparison.

Straight single quotes that follow an open double quote (either right next to the open double or later in the string) seem to never be converted. Interestingly, when I use the button to convert straight to curly, it does an excellent job of guessing, except where it should be an apostrophe. Is this why those aren't converted?

srjfoo (approved):
LGTM!

PR #528, opened 2024-11-16
Highlight cursor line

### LINKED ISSUE #522: Highlight cursor line
The line number (in the left-hand gutter) is highlighted for the current line. I suggest highlighting the line of text itself, as is somewhat commonplace these days in code editors. It would need to be a subtle highlight, and pushed to the bottom of the tag stack so it doesn't interfere with any other tags.

(And potentially be configurable as not every PP'er may want it to be turned on?)

A few examples from Visual Studio Code attached. I'm thinking of something like a pale blue or very light gray background in light mode, and a dark gray (lighter than the background, though) in dark mode.

Screenshot 2024-11-11 at 12 38 30 AM
Screenshot 2024-11-11 at 12 39 53 AM
Screenshot 2024-11-11 at 12 39 33 AM

@windymilla I plan to look at this one.

tangledhelix:
Subtly highlights the line the cursor is positioned on.

This tag is lowered to be below other tags, to avoid interfering with other highlighting. It may be that it should be the last thing run in `after_idle` for that reason.

(Marking as draft until #525 is merged)

Fixes #522

### DIFF
--- a/src/guiguts/highlight.py
+++ b/src/guiguts/highlight.py
@@ -22,6 +22,7 @@ class HighlightTag(StrEnum):
     STRAIGHT_SINGLE_QUOTE = auto()
     CURLY_SINGLE_QUOTE = auto()
     ALIGNCOL = auto()
+    CURSOR_LINE = auto()
 
 
 class HighlightColors:
@@ -85,6 +86,11 @@ class HighlightColors:
     ALIGNCOL = {
         "Light": {"bg": "greenyellow", "fg": "black"},
         "Dark": {"bg": "green", "fg": "white"},
+    }
+
+    CURSOR_LINE = {
+        "Light": {"bg": "#efefef", "fg": "black"},
+        "Dark": {"bg": "#303030", "fg": "white"},
     }
 
 
@@ -482,3 +488,16 @@ def highlight_aligncol_in_viewport(viewport: Text):
 def remove_highlights_aligncol() -> None:
     """Remove highlights for alignment column"""
     maintext().tag_delete(HighlightTag.ALIGNCOL)
+
+
+def highlight_cursor_line() -> None:
+    """Add a highlight to entire line cursor is focused on."""
+    maintext().tag_delete(HighlightTag.CURSOR_LINE)
+
+    # Don't re-highlight if there's currently a selection
+    if not maintext().selected_ranges():
+        _highlight_configure_tag(HighlightTag.CURSOR_LINE, HighlightColors.CURSOR_LINE)
+
+        row = maintext().get_insert_index().row
+        maintext().tag_add(HighlightTag.CURSOR_LINE, f"{row}.0", f"{row+1}.0")
+        maintext().tag_lower(HighlightTag.CURSOR_LINE)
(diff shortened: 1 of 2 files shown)


tangledhelix:
This is ready for testing, it just shouldn't be merged ahead of #525 because it's based on that branch.

tangledhelix:
NTS: when a selection is active, the cursor line is still active behind it and can usually be seen. Sometimes it's the remainder of the last line of the selection; if the selection ends in a full line, the cursor position is column 0 on the next row, leading to the highlight being on the entire row just below the selection. Should probably detect when a selection is active and skip over doing the highlighting.

srjfoo (approved):

windymilla (commented):
Are we happy with the behavior when the text screen is split? The line number is highlighted in blue (for me) in the text window that has focus, and grey in the other text window. The cursor line is highlighted in the window that has focus, but nothing in the other window.
IMO, it's consistent and acceptable, but I'd value your opinions @tangledhelix & @srjfoo

windymilla:
BTW - while testing split screen, I discovered a bug when you try to turn off the line numbers. I don't believe that's anything to do with this PR, so I put in a new issue for that (#530). Thought I'd mention it in case either of you noticed the same.

srjfoo:
> BTW - while testing split screen, I discovered a bug when you try to turn off the line numbers. I don't believe that's anything to do with this PR, so I put in a new issue for that (#530). Thought I'd mention it in case either of you noticed the same.

Good catch. It never even occurred to me to check that -- line numbers are so much a part of my text editing expectations that I have a hard time imagining not wanting them!

srjfoo:
> > > BTW - while testing split screen, I discovered a bug when you try to turn off the line numbers. I don't believe that's anything to do with this PR, so I put in a new issue for that (#530). Thought I'd mention it in case either of you noticed the same.
> > 
> > 
> > Good catch. It never even occurred to me to check that -- line numbers are so much a part of my text editing expectations that I have a hard time imagining not wanting them!
> 
> We could remove the ability to hide them - difficult to see anyone objecting to that, and it would simplify the UI, code & docs.

That sounds like the sort of question that might ought to be asked of those in the gg2 dev channel who are doing some alpha testing for us. I certainly don't object, but I've learned that assuming nobody will can come back to bite us in the derrière. Unless we just want to say "this is the way it is, get used to it"?

windymilla:
#525 is merged now, so I'm happy to merge this whenever you're satisfied @tangledhelix

tangledhelix:
> #525 is merged now, so I'm happy to merge this whenever you're satisfied @tangledhelix

I'm still working on it a bit (and it needs rebasing on the original branch again). The cursorline highlights even when a selection is active, which I'm not a fan of, will try to make that cleaner.

windymilla:
> > #525 is merged now, so I'm happy to merge this whenever you're satisfied @tangledhelix
> 
> I'm still working on it a bit (and it needs rebasing on the original branch again). The cursorline highlights even when a selection is active, which I'm not a fan of, will try to make that cleaner.

I quite liked what I saw (blue selection on top of grey "current" line) but am happy for you to make a change if you've seen something that doesn't look as neat as it might.

tangledhelix:
Fixed; will no longer show cursorline when a selection is in effect.

This is ready for a re-review, but leaving as draft for now until #533 is merged (branch rebase...)

srjfoo (approved):
❤️

windymilla (approved):
Does what it says on the tin!

tangledhelix:
Rebased, marked as ready.

PR #533, opened 2024-11-18
Dynamically detect dark/light mode for "Default" theme

### LINKED ISSUE #532: Refactor _highlight_configure_tag to use is_dark_theme
As discussed in #525, consider updating `_highlight_configure_tag` to use `is_dark_theme` to auto-detect light/dark when calculating what colors to apply.

Potentially this would simplify a few things, including perhaps not needing to define a "Default" value in the color maps at all, and using only "Light" and "Dark".

tangledhelix:
Move highlight color decision down the stack to `_highlight_configure_tag()`. Note that this only affects highlight.py; the base window colors and things like the color of selected text are out of scope here.

- Light theme: uses "Light" colors
- Dark theme: uses "Dark" colors
- Default theme: detects light/dark OS mode using `is_dark_mode()` and chooses Dark or Light colors.

If we adopt this we won't need to define Default colors for each tag type in highlight.py anymore; just Light and Dark.

Testing notes:

Using Default theme, open a document and set an alignment column (that's easiest because the coloring changes between themes). Then change the OS to the other mode and the window should change (at least it does in macOS). The highlight will not immediately change because it is triggered by on_change. Click into the window to move the cursor, or do anything else, and it should update the alignment column color.

Fixes #532

### DIFF
--- a/src/guiguts/highlight.py
+++ b/src/guiguts/highlight.py
@@ -40,61 +40,51 @@ class HighlightColors:
     QUOTEMARK = {
         "Light": {"bg": "#a08dfc", "fg": "black"},
         "Dark": {"bg": "#a08dfc", "fg": "white"},
-        "Default": {"bg": "#a08dfc", "fg": "black"},
     }
 
     SPOTLIGHT = {
         "Light": {"bg": "orange", "fg": "black"},
         "Dark": {"bg": "orange", "fg": "white"},
-        "Default": {"bg": "orange", "fg": "black"},
     }
 
     PAREN = {
         "Light": {"bg": "violet", "fg": "white"},
         "Dark": {"bg": "violet", "fg": "white"},
-        "Default": {"bg": "violet", "fg": "white"},
     }
 
     CURLY_BRACKET = {
         "Light": {"bg": "blue", "fg": "white"},
         "Dark": {"bg": "blue", "fg": "white"},
-        "Default": {"bg": "blue", "fg": "white"},
     }
 
     SQUARE_BRACKET = {
         "Light": {"bg": "purple", "fg": "white"},
         "Dark": {"bg": "purple", "fg": "white"},
-        "Default": {"bg": "purple", "fg": "white"},
     }
 
     STRAIGHT_DOUBLE_QUOTE = {
         "Light": {"bg": "green", "fg": "white"},
         "Dark": {"bg": "green", "fg": "white"},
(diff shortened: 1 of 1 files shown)


tangledhelix:
It may be possible to detect the OS level change somehow, but I haven't tried as of yet.

One idea might be to store the currently observed state (that is, the result of `is_dark_theme`) as a value on root() or somewhere; on each on_change, see if it has changed from the last round, and if so, it could trigger some redraw callback that just calls a few highlighters.

I'm not sure whether I think it's worth pursuing it. It's more elegant, but we already tell users to restart if they change the theme.

One pathological case is auto-mode changes, e.g. when macOS can be configured to automatically switch between light and dark depending on the time of day. Even then, the user just needs to interact with GG to trigger the redraw. And I suspect the majority of users don't auto-switch anyway.

srjfoo (approved):

windymilla (approved):
I'm happy with that for now - as you said (possibly elsewhere) we may be able to use the same idea for other non-highlight colors in future, and if so, we can refactor as necessary.

src/guiguts/highlight.py
```diff
@@ -174,6 +164,13 @@ def _highlight_configure_tag(
         tag_colors: Dictionary of fg/bg colors for each theme.
     """
     theme = preferences.get(PrefKey.THEME_NAME)
+
+    if theme == "Default":
+        if maintext().is_dark_theme():
```
windymilla, line 169: Do you think it would be better/simpler to depend on `is_dark_theme` whatever the theme name, rather than only when the theme is Default, i.e. completely ignore the theme name here? The only case that would be altered is if the "Dark" theme were actually light in color, or the "Light" theme actually dark. In those (unlikely) cases, perhaps we'd prefer the colors to be chosen on the basis of the actual theme darkness, rather than the bizarre name.

tangledhelix: Good idea. The latest commit has that (and the code is simpler to boot).

windymilla:
Looks good on Windows in all 3 themes

srjfoo (approved):
Just a note that if switching between light and dark mode at an OS level on Macs, if the highlight needs to change, it won't actually change dynamically until the main text window is in the foreground (from my testing). I suspect that las to do with the fact that, to change the OS-level theme, GG2 is not the active app. I don't think that's a case that's worth worrying about; I would expect PPers to set a theme and stick with it.

All of the other changes I tried did update dynamically. 👍

PR #540, opened 2024-11-23
Make mypy stricter

windymilla:
Add `--disallow-untyped-defs` option to mypy run.

Without this, it doesn't catch
```def highlight_aligncol_in_viewport(viewport: Text):```
which doesn't have a return type declared.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -43,8 +43,8 @@ jobs:
         poetry run python -m sphinx -M coverage docs docs/build -W
     - name: Static type check with mypy
       run: |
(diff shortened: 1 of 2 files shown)


tangledhelix (approved):
Looks good to me - I had believed this check was already in place! I didn't test it myself (Github Actions will do that for us).

windymilla:
Thanks @tangledhelix - yes, I was surprised when I ran my "check everything" script (where I do have that flag set) and it reported an error in master. I didn't add it originally to Github Actions because I was unsure how much of a pain it would be, and so just left the settings at the defaults.

PR #542, opened 2024-11-24
Globally initialize highlight tags at startup

### LINKED ISSUE #537: Initialize all highlight tags at startup instead of as-needed
See #41 for discussion.

Rather than configure highlight tags repeatedly `_highlight_configure_tag()` and having to do so repeatedly because we are using `tag_delete()`, instead initialize them all once at startup, and never delete a tag, only use e.g. `tag_remove(tagname, "1.0", "end")` to remove the tag from the text, rather than removing the tag from existence entirely.

Also implement a ranking of tags with `.raise()`, `.lower()` (which tags display over top of other tags and in what order) as discussed in #41. Proposed (lowest to highest):

- cursorline (which is suppressed when a selection is active)
- alignment column
- surrounding quotes & brackets
- selection(s), if active
- manual highlights (single/double quotes, string_or_regex)
- spotlight (e.g. current match for a search)

tangledhelix:
Rather than reconfigure highlight colors every time they're used, configure once at startup.

Rather than delete tags entirely every time we want to remove them from the UI, instead only remove the application of the tag (to text regions) and leave the tag itself defined (and configured with colors).

Define an order for highlight precedence, all in one place rather than scattered around the code.

Refactored a couple of instances of `"end"` to `tk.END`.

Fixes #537

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -525,6 +525,9 @@ class MainText(tk.Text):
         # Force focus to maintext widget
         self.after_idle(lambda: grab_focus(self.root, self, True))
 
+        # Initialize highlighting tags
+        self.after_idle(self.highlight_initialize_tags)
+
     def do_nothing(self) -> None:
         """The only winning move is not to play."""
         return
@@ -2489,12 +2492,11 @@ class MainText(tk.Text):
 
     def remove_highlights(self) -> None:
         """Remove active highlights."""
-        self.tag_delete(HighlightTag.QUOTEMARK)
+        self.tag_remove(HighlightTag.QUOTEMARK, "1.0", tk.END)
 
     def highlight_quotemarks(self, pat: str) -> None:
         """Highlight quote marks in current selection which match a pattern."""
         self.remove_highlights()
-        self._highlight_configure_tag(HighlightTag.QUOTEMARK, HighlightColors.QUOTEMARK)
         self.highlight_selection(pat, HighlightTag.QUOTEMARK, regexp=True)
 
     def highlight_single_quotes(self) -> None:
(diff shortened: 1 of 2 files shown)


src/guiguts/maintext.py
```diff
@@ -435,6 +435,13 @@ def switch_text_peer(event: tk.Event) -> None:
         # Force focus to maintext widget
         self.after_idle(lambda: grab_focus(self.root, self, True))
 
+        # Initialize highlighting tags
+        from guiguts.highlight import (  # pylint: disable=import-outside-toplevel
```
windymilla, line 439: I would prefer we follow pylint's suggestion and only import at the top of the file. Is this actually revealing that we have circular imports between maintext and highlight? So far I have managed to avoid those, as I am fearful of a tangled web developing. Is there an alternative, or are we stuck with this?

windymilla (commented):
Seems to work fine - just a small question about the imports

src/guiguts/maintext.py
tangledhelix: That's exactly it. So we can either import on demand in `Maintext.__init__` or an alternative is to move the `HighlightTag` and `HighlightColors` classes, and the `highlight_initialize_tags` method, into `maintext.py` or somewhere else that is already initialized, or this is needed to avoid a circular import exception.

This was also seen in #519 and I would assume also affects the imports in maintext.py for #525 and #528. (So basically all of the highlighting bits).

For **this** PR that would probably be a small number of things, but once you consider all the other highlighter work, I think we'd end up moving much or all of highlight.py to be part of maintext.py (because most things in highlight.py are calling `maintext()` for something or other). If you strongly want to avoid this type of import, I'm okay with doing that. One could argue that highlighting is just one of the attributes / activities of MainText.

Editorial note: in the prior sentence, when I was typing "attributes", Github or my browser or something suggested "atrocities" as an autocompletion 😆

tangledhelix: To be more direct, I will (in a different branch than this PR, for now) work on a refactor to move highlighting into maintext.py. It probably won't be that hard but we'll want to re-test *all* of the highlighters after.

tangledhelix:
Rebased.

src/guiguts/maintext.py
tangledhelix: Handled in #544

srjfoo (approved):

PR #544, opened 2024-11-26
Refactor highlight.py out of existence

tangledhelix:
Related to #542

This is, primarily, moving most of highlight.py into maintext.py as part of the `MainText` class. Various other bits patched up to continue working.

Also moves the state variables for alignment column from the root window to the maintext object.

I tested this as I went, but I would encourage testing all of the highlighting features again just to be sure everything is still working:
- Search | Highlight single quotes in selection
- Search | Highlight double quotes in selection
- Search | Remove highlights
- Search | Highlight alignment column
- Search | Highlight surrounding quotes & brackets
- Cursor line highlight
- "spotlight" highlight used by e.g. Word Frequency or Word Distance Check
- Anything else you can think of involving highlights or their coloration.

There are no more circular import exceptions as far as I'm aware.


windymilla (approved):
Thanks - I was unable to break any of the highlighting.

srjfoo:
I don't know if this is something you want to fix or not, but I assume it's unintentional:

image

PPers won't even notice it unless there's a character in line 1, column 1 (which there was for one of my test files). Unless the cursor is at column 0, of any line, L1, C1 is highlighted, along with the column following the cursor position.

Which brings me to the next point, which, I think, is related to how the cursor position is identified.

If the cursor is positioned preceding column 1, the status bar identifies that as c0. The cursor column position is (and I believe has been, traditionally?) identified in GG by the character preceding it. That is, if the cursor position is at L:14 C:0, the cursor is positioned before the first character on line 14, and if C=1, then it is following the first character on the line. 

In this particular case, if C=1, it is actually column 2 that is highlighted. **_EXCEPT_** that if C=0, column 1 is not highlighted.

It's entirely possible that there's no good reason fot column 1 to be highlighted, but, as with L1,C1 being highlighted when alignment column >1, if there's text there, it's an inconsistency.

-----

At least on default/dark, the highlight color for straight double quotes is either the same as the color for the column alignment highlighting, or very close. (I'd take the latter a bit darker, maybe?) Again, it may not be worth worrying about, but I thought I'd mention it.

srjfoo (approved):
> > I don't know if this is something you want to fix or not, but I assume it's unintentional:
> 
> Aha! I have seen this before, although I was not able to reproduce it and therefore never fixed it. I didn't understand what conditions were needed to repro, but you've articulated them here. I'll open another ticket to fix that (it's not introduced by this refactor).

Other than the above and that I have some color adjustment suggestions for dark/default-dark that I need to work through again, everything looked good. I was assuming that color adjustments might be in another PR?

tangledhelix:
> Other than the above and that I have some color adjustment suggestions for dark/default-dark that I need to work through again, everything looked good. I was assuming that color adjustments might be in another PR?

Yes, I'll make a PR for that.

PR #547, opened 2024-11-27
Improve handling of illegal language codes

### LINKED ISSUE #546: Illegal language settings permitted & then causing error
As described [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1353243#p1353243) it appears the user can set the language to "English", which is actually an illegal language code. Language codes should contain lowercase letters and maybe an underscore.

windymilla:
1. Don't allow illegal values to be saved
2. Output an error message, rather than using an assertion.

Fixes #546

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -894,7 +894,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         )
 
         the_statusbar.add(
-            "languages label", tooltip="Click: Set language(s)", text="Lang: "
+            "languages label",
+            tooltip="Click: Set language(s), e.g. 'en' or 'de fr'",
+            text="Lang: ",
         )
         the_statusbar.add_binding(
             "languages label", "ButtonRelease-1", self.file.set_languages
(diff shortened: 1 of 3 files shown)


srjfoo (approved):
Does what it says on the label.

srjfoo:
Actually ... I had an idea just after I sent the approval ...

I tried a 3-letter language code (`abk`). Then `abk en`, then, getting sillier, `abcd`, and finally, `english`. All were accepted. The last one I used was `english`, and gg did not give an error message either then or upon restarting the application. Is that what you want?

windymilla:
That's correct. It just checks that the language is lowercase possibly with an underscore. It's not worth checking it against all possible language codes

srjfoo:
> That's correct. It just checks that the language is lowercase possibly with an underscore. It's not worth checking it against all possible language codes

I was thinking more along the lines of checking against installed dictionaries, and if there isn't one, asking if it should be created -- but that also might not make any sense.

windymilla:
I don't think we should limit it to installed dictionaries

srjfoo:
Is there a value of having an assigned language where there isn't a dictionary?

windymilla:
> Is there a value of having an assigned language where there isn't a dictionary?

It could get used in the language attribute of the HTML file, or the user may be going to create a custom language dictionary in the future?

PR #549, opened 2024-11-28
Catch edge-case exception in macOS clipboard

### LINKED ISSUE #548: Exception thrown in affirm_clipboard_contents
As reported by @srjfoo:

```
23:03:18: ERROR - root.py:63 - Tkinter Exception
  File "/opt/homebrew/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/tkinter/__init__.py", line 1967, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/tkinter/__init__.py", line 861, in callit
    func(*args)
  File "/Users/sharonjoiner/dpdev/guiguts-py-3.11/src/guiguts/maintext.py", line 1160, in affirm_clipboard_contents
    proc.communicate(input=self.clipboard_get().encode())
                           ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/tkinter/__init__.py", line 926, in clipboard_get
    return self.tk.call(('clipboard', 'get') + self._options(kw))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_tkinter.TclError
```

In further testing, I discovered that these conditions must be met to trigger the exception:

1. Must be on macOS (because `affirm_clipboard_contents` raises `NotImplementedError` on other platforms)
2. The system clipboard must currently contain non-text (e.g., an image) - Emoji did not trigger the error since they are characters, not images.
3. There must not be a selection active in Guiguts.
4. The user must trigger either the "copy" or the "cut" function (via context menu, or menu bar, or key bind).

Under these conditions, Guiguts throws `_tkinter.TclError`.

If a selection is active, then the cut or copy operation copies that text before running `affirm_clipboard_contents` and therefore the problem isn't triggered.

tangledhelix:
If macOS clipboard contains non-text (e.g., an image) and user triggers smart_cut or smart_copy functions, TclError is thrown when TCL code "clipboard get" is run. Catch error and return from method if this happens.

Fixes #548

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1156,8 +1156,11 @@ class MainText(tk.Text):
             raise NotImplementedError("This function only works on macOS")
 
         # Use pbcopy macOS command to "touch" the clipboard contents
-        with subprocess.Popen(["/usr/bin/pbcopy"], stdin=subprocess.PIPE) as proc:
-            proc.communicate(input=self.clipboard_get().encode())
+        try:
+            with subprocess.Popen(["/usr/bin/pbcopy"], stdin=subprocess.PIPE) as proc:
+                proc.communicate(input=self.clipboard_get().encode())
+        except tk.TclError:
+            pass
 
     def smart_copy(self) -> str:
         """Do column copy if multiple ranges selected, else default copy."""


tangledhelix:
See #548 for how to reproduce the problem; can test on master branch and on this branch to compare whether you get an exception under the same conditions.

srjfoo (approved):

windymilla (approved):

PR #550, opened 2024-11-28
Never return a starting row less than 1

### LINKED ISSUE #545: Fix errant 1.0 highlight in "highlight alignment column" mode
As reported by @srjfoo in review of #544:

----------

I don't know if this is something you want to fix or not, but I assume it's unintentional:

image

PPers won't even notice it unless there's a character in line 1, column 1 (which there was for one of my test files). Unless the cursor is at column 0, of any line, L1, C1 is highlighted, along with the column following the cursor position.

Which brings me to the next point, which, I think, is related to how the cursor position is identified.

If the cursor is positioned preceding column 1, the status bar identifies that as c0. The cursor column position is (and I believe has been, traditionally?) identified in GG by the character preceding it. That is, if the cursor position is at L:14 C:0, the cursor is positioned before the first character on line 14, and if C=1, then it is following the first character on the line.

In this particular case, if C=1, it is actually column 2 that is highlighted. **_EXCEPT_** that if C=0, column 1 is not highlighted.

It's entirely possible that there's no good reason fot column 1 to be highlighted, but, as with L1,C1 being highlighted when alignment column >1, if there's text there, it's an inconsistency.

At least on default/dark, the highlight color for straight double quotes is either the same as the color for the column alignment highlighting, or very close. (I'd take the latter a bit darker, maybe?) Again, it may not be worth worrying about, but I thought I'd mention it.

tangledhelix:
The first row is always row 1; there is no row 0, so never return row 0 when examining the current viewport's range of rows.

Fixes #545

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2544,8 +2544,8 @@ class MainText(tk.Text):
         # {top,bot}_frac contain a fractional number representing a percentage into
         # the document; do some math to calculate what the top or bottom row in the
         # viewport should be, then use min/max to make sure that value isn't less
-        # than 0 or more than the total row count.
-        top_line = max(int((top_frac * end_index.row) - offscreen_lines), 0)
+        # than 1 or more than the total row count.
+        top_line = max(int((top_frac * end_index.row) - offscreen_lines), 1)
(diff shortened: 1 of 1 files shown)


windymilla (approved):
LGTM!

srjfoo (approved):

PR #552, opened 2024-11-28
Highlight align column before cursor, not after

### LINKED ISSUE #551: Change highlight align column behavior (before cursor, not after)
GG1's behavior was to highlight the column just **before** the cursor, not just after (as is the current behavior in GG1).

Aside from being consistent with the behavior PPers are used to, the comments in GG1 indicate this has another reason:

```perl
# Highlight column immediately preceding cursor for consistency with ruler
```

Note we have a pending issue to implement a ruler (#61) although it's not in place today.

tangledhelix:
To be consistent with Guiguts 1, highlight the column to the left of the cursor, not to the right.

Fixes #551

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2777,13 +2777,16 @@ class MainText(tk.Text):
         while row <= end_row:
             # find length of row; don't highlight if row is too short to contain col
             rowlen = int(self.index(f"{row}.0 lineend").split(".")[1])
-            if 0 < col < rowlen:
+            if 0 <= col < rowlen:
                 self.tag_add(HighlightTag.ALIGNCOL, f"{row}.{col}")
             row += 1
 
     def highlight_aligncol(self) -> None:
         """Add a highlight to all characters in the alignment column."""
-        if self.aligncol_active.get():
+        # Check that alignment column is 0 or higher; there are no negative
+        # columns in a textview. Since the column is decremented when align
+        # highlight is turned on, it's possible that this value is set to -1.
+        if self.aligncol_active.get() and self.aligncol >= 0:
             self.tag_remove(HighlightTag.ALIGNCOL, "1.0", tk.END)
 
             self.highlight_aligncol_in_viewport(self)
@@ -2797,9 +2800,12 @@ class MainText(tk.Text):
     def highlight_aligncol_callback(self, value: bool) -> None:
         """Callback when highlight_aligncol active state is changed."""
         if value:
-            self.aligncol = self.get_insert_index().col
+            # Highlight column immediately preceding cursor for consistency with ruler.
+            # (There is no ruler as of today, but we'll implement one in GG2 at some point...)
+            self.aligncol = self.get_insert_index().col - 1
             self.highlight_aligncol()
         else:
+            self.aligncol = -1
             self.remove_highlights_aligncol()
 
     def highlight_cursor_line(self) -> None:


tangledhelix:
Testing notes:
- Position the cursor and turn on the alignment column highlight
- Highlight should be on the column to the *left* of the cursor, not the right like before
- This should work even when positioned on column 1 (will highlight column 0)
- If positioned on column 0, highlighting is skipped over, although the mode remains turned on. This is because highlighting to the left of column 0 wouldn't display anything anyway.

windymilla (approved):
It works as described. I did consider whether we should keep this one as it was, and change the yet-to-be-implemented ruler to also mark the column after the cursor. However, looking at the GG1 ruler, which looks like this:
`....,....1....,....2...` etc
The "1" represents column 10, and the "2" represents 20 - that wouldn't work if we changed which column got marked on the ruler - either that or we'd only have three periods before the first comma, or something else weird. Also, the row/col status bar button shows column "0" when the cursor is at the start of the line, and we definitely don't want to change that behavior.
So, unless anyone can think of a better solution, I think we'll have to stick with what I implemented for the GG1 ruler.
This PR is therefore what we need for column highlighting.

windymilla:
> * If positioned on column 0, highlighting is skipped over, although the mode remains turned on. This is because highlighting to the left of column 0 wouldn't display anything anyway.

Just a thought on this point. What would you think about still highlighting column 0 in this case (as well as when the cursor is in column 1)? I know strictly that would be inconsistent, but if a user turned on column highlighting with the cursor in column 0, it's almost certainly what they wanted (however unlikely that may seem). Just a thought...

windymilla:
> One more reason to leave it as-is;...<snip>

Good point!

srjfoo:
> > * If positioned on column 0, highlighting is skipped over, although the mode remains turned on. This is because highlighting to the left of column 0 wouldn't display anything anyway.
> 
> Just a thought on this point. What would you think about still highlighting column 0 in this case (as well as when the cursor is in column 1)? I know strictly that would be inconsistent, but if a user turned on column highlighting with the cursor in column 0, it's almost certainly what they wanted (however unlikely that may seem). Just a thought...

If another vote is needed, I'd rather just see it documented that if a PPer wants to align with column 0, they should use the left edge of the maintext window. There's not really a need for a highlighted alignment column for c0. I would consider highlighting at c1 for both c0 and c1 to be more confusing than helpful.

...

> * If positioned on column 0, highlighting is skipped over, although the mode remains turned on. This is because highlighting to the left of column 0 wouldn't display anything anyway.

I know it would be more work, but why not just alert the PPer that column highlighting cannot be displayed for c0, and leave it turned off. That way when they do set a column highlighting, at least they don't have to unset it before they can.

BTW, one inconsistency I noticed is that "Highlight Surrounding Quotes & Brackets" persists, but "Highlight Alignment Column" does not. I'm okay with it that way, but just wanted to make sure that was intentional.

-----

Definitely, we should keep the column reporting in the status bar and the ruler (once implemented) and the alignment column consistent.

srjfoo (approved):

tangledhelix:
> > * If positioned on column 0, highlighting is skipped over, although the mode remains turned on. This is because highlighting to the left of column 0 wouldn't display anything anyway.
> 
> I know it would be more work, but why not just alert the PPer that column highlighting cannot be displayed for c0, and leave it turned off. That way when they do set a column highlighting, at least they don't have to unset it before they can.

Opened #557 for that.

> BTW, one inconsistency I noticed is that "Highlight Surrounding Quotes & Brackets" persists, but "Highlight Alignment Column" does not. I'm okay with it that way, but just wanted to make sure that was intentional.

It is intentional, to mimic the GG1 behavior.

PR #556, opened 2024-11-30
Dynamically detect theme change

### LINKED ISSUE #555: Detect theme change and reconfigure highlight colors
As discussed in #543, the refactor to set up tag colors at startup seems to have broken the ability to adapt to the theme changing while GG is running.

Proposal:
- Store the result of `is_dark_theme` somewhere, e.g. a class variable on maintext
- During `on_change`, check `is_dark_theme` against that stored value
- If it has changed, then set the class variable to the new value and reconfigure the tag colors

tangledhelix:
Detect changes in theme and adjust tag colors.

Testing notes: just change the theme in any way, and it should work.

- Changing in GG preferences: should take effect immediately
- Changing OS dark mode: will have to interact with GG to trigger update

Fixes #555

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -525,8 +525,11 @@ class MainText(tk.Text):
         # Force focus to maintext widget
         self.after_idle(lambda: grab_focus(self.root, self, True))
 
+        # Whether we were on dark theme the last time we looked (bool)
+        self.dark_theme = self.is_dark_theme()
+
         # Initialize highlighting tags
-        self.after_idle(self.highlight_initialize_tags)
+        self.after_idle(lambda: self.highlight_configure_tags(first_run=True))
 
     def do_nothing(self) -> None:
         """The only winning move is not to play."""
@@ -635,6 +638,8 @@ class MainText(tk.Text):
             self.root.after_idle(self._do_linenumbers_redraw)
             self.root.after_idle(self._call_config_callbacks)
             self.root.after_idle(self.save_sash_coords)
+            # run `highlight_configure_tags` _before_ other highlighters
+            self.root.after_idle(self.highlight_configure_tags)
             self.root.after_idle(self.highlight_quotbrac)
             self.root.after_idle(self.highlight_aligncol)
             self.root.after_idle(self.highlight_cursor_line)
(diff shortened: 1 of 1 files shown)


src/guiguts/maintext.py
```diff
@@ -2829,10 +2834,28 @@ def highlight_cursor_line(self) -> None:
             row = self.get_insert_index().row
             self.tag_add(HighlightTag.CURSOR_LINE, f"{row}.0", f"{row+1}.0")
 
-    def highlight_initialize_tags(self) -> None:
-        """Initialize highlighting tags and determine their prioritization.
-        Intended to run once at startup.
+    def highlight_configure_tags(self, **kwargs: Any) -> None:
```
windymilla, line 2837: Is this equivalent to
```
def highlight_configure_tags(self, first_run: bool = False) -> None:
```
then later instead of 
```
if "first_run" in kwargs and kwargs["first_run"]:
```
using
```
if first_run:
```
If so, then I think the latter would be clearer, but I may have missed some subtle difference.

windymilla (commented):
I like this way of configuring the colors/order, and only changing what's needed. Just a small question about the definition of `highlight_configure_tags`. 
It seemed to work fine, starting up with the correct color (e.g. for the current line highlight) and changing when the theme changed.

src/guiguts/maintext.py
tangledhelix: Functionally those are equivalent, yes. I actually started with that form. I changed it because the initial call had a "magic value" argument that I thought read less clearly. That is, using a positional argument, your call in `__init__` would be:

```python
self.highlight_configure_tags(True)
```

Which works, but it's not clear what True means without checking the function definition.

tangledhelix: I can see your point about being less readable on the other end (in the function definition), but I was hoping the docstring would clarify it enough.

windymilla: On nomenclature, I thought a keyword argument was just one that had a default value in the function def, so I think in my suggestion `first_run` is still a keyword argument. The `kwargs` thing is when you have some "unknown" keyword args.

windymilla: Just looking up some Python stuff to try to understand the nomenclature better

tangledhelix: *smacks self in forehead*

Well, I learned something new today! I didn't realize you could pass the name of a positional argument! I wonder if that changed in the Python 2 to 3 shift... I was 99% sure that didn't work...

Anyway, it does work, and I just pushed a new version with that change.

windymilla: NTS: This answer was the most helpful: https://stackoverflow.com/a/57819001

windymilla (approved):

srjfoo (approved):

PR #559, opened 2024-12-01
Pop warning dialog if attempting alignment column at col=0

### LINKED ISSUE #557: Don't activate alignment column at column 0
As noted by @srjfoo in #552:

>> @tangledhelix said:
>> If positioned on column 0, highlighting is skipped over, although the mode remains turned on. This is because highlighting to the left of column 0 wouldn't display anything anyway.
>
> I know it would be more work, but why not just alert the PPer that column highlighting cannot be displayed for c0, and leave it turned off. That way when they do set a column highlighting, at least they don't have to unset it before they can.

If the cursor is on column 0, display a notice that alignment column can't be placed here, and do nothing else.

tangledhelix:
If at column 0, and trying to turn on alignment column, pop a warning window, and don't turn alignment mode on.

At any other column, behaves as before.

Fixes #557

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2812,9 +2812,17 @@ class MainText(tk.Text):
     def highlight_aligncol_callback(self, value: bool) -> None:
         """Callback when highlight_aligncol active state is changed."""
         if value:
+            col = self.get_insert_index().col
+            if col == 0:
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

windymilla (approved):
LGTM

PR #562, opened 2024-12-02
Hide insert cursor when there's a selection

windymilla:


srjfoo (commented):
I like the selection block without the cursor, but de-selecting with arrow keys only works correctly for left and right (beginning and end of selection) arrow keys. I checked, and `master` matches this behavior, as well, so it's not something specific to this PR:

up arrow goes up from the location of the endpoint of the selection, whether it's visible or not. That is, if the selection was top left to bottom right, and you click the up arrow, the cursor winds up on the line above the end-point of the selection, in the same column as the endpoint of the selection, not one line above the beginning of the selection block. Similar behavior if the selection was bottom to top, when using the down-arrow to de-select. Caveat -- if you use the button in the toolbar to restore the selection, you can change which is apparently start and end-point by using left and right arrow to de-select the block.

I'd appreciate @tangledhelix's opinion on whether we should proceed with this change. I think hiding the cursor without fixing how de-selecting works with up and down arrows would be a mistake, and much as I like no visible cursor in a selection, I think that it might be a leeeetle bit confusing for users under the circumstances.

windymilla:
I'll take a look at doing the same with Up/Down as we do with Left/Right.
There's another related issue: if you double click a word to select it by clicking near the middle of the word, not one end or the other, then you use Shift+Left/Right or Ctrl+Shift+Left/Right (& maybe Up/Down), instead of it extending/shrinking the selection from one end of the selected word, it does it from the place where you double clicked.

srjfoo:
> Latest commit should cause Up/Down to behave as expected.

For me, the up and down arrows now behave the same as left (up) and right (down) arrows. I.e. they put the cursor at the beginning (up) or end (down) of the former selection, instead of the line above (up) or below (down) the former selection, in the same relative column. Everything else looks good, and this behavior doesn't bother me, I just wanted to make sure that it's intentional.

tangledhelix:
LGTM, and this behavior is what I had envisioned (but not stated anywhere), to just make up act like left, and down act like right. I see it works the same in columnar selection as well (which is good).

windymilla:
> ...I see it works the same in columnar selection as well (which is good).

Oh good - I forgot about column selection, but glad to hear it works OK.

tangledhelix:
I just thought to try - and it appears that Shift- prefixing those 4 keys will let you create a selection just like holding shift with arrow keys will. Nothing to do with this - just something that I noticed....

windymilla:
> Just after I posted that I found a problem 😄
> 
> The "emacs movements" in Mac (and Linux?) (e.g. Ctrl-B, Ctrl-F) aren't caught and behave the old way still. May want to catch those and make the behavior match the arrows, at least on the platforms where they're supposed to work.
> 
> * Ctrl-B = left
> * Ctrl-F = right
> * Ctrl-N = down
> * Ctrl-P = up

I would prefer not to have those at all, because at least on Windows, those are valuable shortcut keys for doing something useful. However, on Macs only, search is Cmd-F, not Ctrl-F, so you can keep the emacs movements on Macs.

tangledhelix:
For _consistency_ possible also catch Ctrl-A and Ctrl-E as:

- Ctrl-A, move to start of the line at beginning of selection
- Ctrl-E, move to end of the line at end of selection

As opposed to now where they can be either side depending which line the hidden cursor is actually sitting on...

tangledhelix:
> > Just after I posted that I found a problem 😄
> > The "emacs movements" in Mac (and Linux?) (e.g. Ctrl-B, Ctrl-F) aren't caught and behave the old way still. May want to catch those and make the behavior match the arrows, at least on the platforms where they're supposed to work.
> > 
> > * Ctrl-B = left
> > * Ctrl-F = right
> > * Ctrl-N = down
> > * Ctrl-P = up
> 
> I would prefer not to have those at all, because at least on Windows, those are valuable shortcut keys for doing something useful. However, on Macs only, search is Cmd-F, not Ctrl-F, so you can keep the emacs movements on Macs.

That's what I mean, just catch them on the platform(s) they currently are used for movement on, so they behave the same as arrows. Not add them to Windows.

windymilla:
> I just thought to try - and it appears that Shift- prefixing those 4 keys will let you create a selection just like holding shift with arrow keys will. Nothing to do with this - just something that I noticed....

I've just noticed this comment, so will adjust my commit slightly to try to let that behavior continue to work.

srjfoo:
> > I just thought to try - and it appears that Shift- prefixing those 4 keys will let you create a selection just like holding shift with arrow keys will. Nothing to do with this - just something that I noticed....
> 
> I've just noticed this comment, so will adjust my commit slightly to try to let that behavior continue to work.

I'm not entirely sure what's going on here, but in my testing:

1: Regular click-and-drag multi-line selection.
* If I drag top to bottom, both shift+right arrow and ctrl+shift+f increase the selection from the endpoint of the beginning selection, and shift+left arrow and crl+shift+b decrease the selection from the endpoint of the beginning selection.
* If I drag bottom to top, similarly, shift+left arrow and ctrl+shift+b will increase the selection from the start point of the beginning selection and shift+right arrow and ctrl+shift+f will decrease the selection from the start point of the beginning selelction.

What this says to me is that the selection is not being treated as a unit. 

2: Column multi-line selection. All four options (shift plus left or right arrow, or ctrl+b or ctrl+f) convert the column selection to a regular selection and from there follow the steps above, with the addition that the keycombo that converts the selection from columnar to normal also selects/deselects a single character, depending on the direction of the movement. 

3: Same as 2 but using up/down arrows and ctrl+n or ctrl+p, converts from column to normal selection and increases or decreases the selection up or down a line from the position of the invisible cursor.

4. Ctrl+a/e deselect and move immediately to beginning of first line of selection/end of last line of selection, but there is not an arrow equivalent. One might think that cmd+left or right arrow would, but they actually match what the left and right arrow do alone; another cmd+arrow is necessary to move to the beginning or end of the relevant line.

Hopeing this all makes sense.

windymilla:
> 1: Regular click-and-drag multi-line selection.
> 
> * If I drag top to bottom, both shift+right arrow and ctrl+shift+f increase the selection from the endpoint of the beginning selection, and shift+left arrow and crl+shift+b decrease the selection from the endpoint of the beginning selection.
> * If I drag bottom to top, similarly, shift+left arrow and ctrl+shift+b will increase the selection from the start point of the beginning selection and shift+right arrow and ctrl+shift+f will decrease the selection from the start point of the beginning selelction.
> 
> What this says to me is that the selection is not being treated as a unit.

This is expected/deliberate, and I think it makes sense. If you drag top to bottom or bottom to top, then adjust the selection with shift-arrows, it extends/shrinks from where you released the mouse after dragging. TBH there isn't a single "right answer", but I think this is consistent and predictable for the user.

> 2: Column multi-line selection. All four options (shift plus left or right arrow, or ctrl+b or ctrl+f) convert the column selection to a regular selection and from there follow the steps above, with the addition that the keycombo that converts the selection from columnar to normal also selects/deselects a single character, depending on the direction of the movement.

I don't think there's a practical alternative here. It's very much an edge case, and although perhaps unexpected, it is consistent with other extending behavior for column selections, e.g. Shift-click outside the selection. I don't think it's worth the dev time to rewrite that behavior

> 3: Same as 2 but using up/down arrows and ctrl+n or ctrl+p, converts from column to normal selection and increases or decreases the selection up or down a line from the position of the invisible cursor.

As above.

> 4. Ctrl+a/e deselect and move immediately to beginning of first line of selection/end of last line of selection, but there is not an arrow equivalent. One might think that cmd+left or right arrow would, but they actually match what the left and right arrow do alone; another cmd+arrow is necessary to move to the beginning or end of the relevant line.

I've pushed another commit that should fix this.

> Hope this all makes sense.

Although I've said I think most of these are OK, thanks for the extensive testing and questioning the behavior.

tangledhelix:
I'm on the fence about whether the shift-ctrl-{emacs keys} makes sense or is intuitive, but even on the master branch they work (although they follow the cursor, which isn't handled magically there) so I think it's fine. By default they are doing something anyway.

LGTM as-is.

srjfoo (approved):
LGTM, too!

Issue #384, opened 2024-08-13
Tools->Convert Fractions->Unicode Fractions inconsistently misses some


okrick:
The 'Tools -> Convert Fractions -> Unicode Fractions Only' and 'Tools -> Convert Fractions -> Unicode Fractions or Super/Subscript' options inconsistently fail to convert fractions on the first line of a block markup. For example, the fraction '1/4' might remain unchanged despite the conversion settings. This issue does not appear to occur with the 'Tools -> Convert Fractions -> All to Super/Subscript' option."  For example:

```
/*
1/4 POUND CHINESE CURED PORK (Cut into fine shreds.) (See page 4 for recipe for curing pork.)
1/2 CUP BAMBOO SHOOT (Cut into fine, long shreds.)
4 CHINESE WATER CHESNUTS[**P1: CHESTNUTS?] (Pared and cut into fine shreds.)
1/2 CUP CELERY (Cut into fine, long shreds.)
1 STEM OF GREEN ONION TOP (Cut small.)
6 EGGS.
1 TEASPOONFUL CHINESE SALTY SAUCE.
*/
```
sometimes results in 
```
/*
1/4 POUND CHINESE CURED PORK (Cut into fine shreds.) (See page 4 for recipe for curing pork.)
½ CUP BAMBOO SHOOT (Cut into fine, long shreds.)
4 CHINESE WATER CHESNUTS (Pared and cut into fine shreds.)
½ CUP CELERY (Cut into fine, long shreds.)
1 STEM OF GREEN ONION TOP (Cut small.)
6 EGGS.
1 TEASPOONFUL CHINESE SALTY SAUCE.
 */
```

The only thing I found consistent is that the missed fraction is always the list's first item.  But I can't replicate it every time despite repeated tries.  It seems random.

[projectID6579ecdbc1e27.zip](https://github.com/user-attachments/files/16605759/projectID6579ecdbc1e27.zip)

windymilla:
@okrick, I'm sorry - despite repeated attempts, I can't replicate this.

srjfoo:
@okrick, neither can I. I'm not doubting that you've seen an issue with it, but the only way I could get the 1/4 on the first line to not convert was to fail to select the `1` in the fraction. I tried multiple different selections, and multiple times (though not more than 3 to 5 times each):
* Select all
* Selecting the block including the no-wrap markup
* Selecting the block without the markup
* Selecting the line
* Selecting just the character.

[...]

PR #568, opened 2024-12-06
Fix fraction convert error

### LINKED ISSUE #563: Convert Fractions misses fractions at start of line
If a number with fraction starts in column 0, then it is missed. Happens with both "2-1/2" and "1/2" styles, and all 3 conversion choices. Inserting a blank in col. 0 fixes the issue.

Version 2.0.0a10
Windows 11 64-bit

windymilla:
Fraction failed to convert if previous line ended in a letter. This was due to surprising (to me) regex behavior where a dollar matches end-of-string OR a newline at the end of the string.

Fixes #563 

Almost certainly this was the inconsistent conversion failure noted in #384 by @okrick in lists of ingredients for recipes, most of which had lines ending with a period, so didn't trigger the bug - sorry for not managing to reproduce it then @okrick!


rtonsing (approved):
I was able to confirm it. Thanks!

srjfoo (approved):
LGTM!

PR #570, opened 2024-12-07
Fix bug when typing over a column selection

windymilla:
Previously it converted the column selection to a
regular selection, then deleted it and inserted the typed character.

Now, it just deselects the text and inserts the
character as usual. This is not perfect, but is less surprising than previous behavior.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -545,6 +545,18 @@ class MainText(tk.Text):
         self.bind_event(
             "<Configure>", self._on_change, add=True, force_break=False, bind_peer=True
         )
+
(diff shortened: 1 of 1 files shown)


tangledhelix (approved):
Seems to work as you describe, LGTM

srjfoo (approved):
Agreed -- it works as advertised. 

I do still recommend, though, that we document that typing over a column-select does not behave the same as common text editors.

windymilla:
Comment added to [manual](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/Edit_Menu#Selecting_Columns)

PR #574, opened 2024-12-16
VScode debugger to force default settings

tangledhelix:
Adds a VScode debugging mode that forces all default settings by calling Guiguts with --nohome in effect.


windymilla (approved):
Excellent - very helpful, thanks!

PR #577, opened 2024-12-17
Initialize peer colors at startup

### LINKED ISSUE #576: Text view split sometimes uses wrong theme
Under these conditions:

- Theme is configured as Dark
- No file is currently loaded

If you turn on the split text view, then the peer will be brought up in Light theme.

If a file is loaded, then the theme is handled correctly.

Tested on macOS, not on others. No other combination of theme settings seems to be a problem, only this one.

tangledhelix:
Initialize the peer editor's colors at startup even if peer isn't currently visible. Otherwise in certain conditions, its theme won't match the other editor when the peer is made visible.

Fixes #576

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -513,6 +513,8 @@ class MainText(tk.Text):
         self.peer.bind(
             "<<ThemeChanged>>", lambda _event: theme_set_tk_widget_colors(self.peer)
         )
+        theme_set_tk_widget_colors(self.peer)
+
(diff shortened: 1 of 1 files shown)


tangledhelix:
Testing notes:
- If the split view is enabled, turn it off
- Set Guiguts theme to Dark
- Quit Guiguts, restart
- Don't load a file
- Turn on split view

To reproduce the buggy behavior, follow the above steps on the master branch. And the same process will test that it's working correctly on this branch.

windymilla (approved):
Good catch, thanks!

srjfoo (approved):
Works for me!

PR #580, opened 2024-12-18
Implement HTML autogeneration

windymilla:
This does not do everything that a PPer needs to consider when creating the HTML file, but it attempts to do the basic autogeneration of the HTML in the same way as GG1. There will usually be further work to be done, especially of the frontmatter, tables, illustrations, etc., which really require human input. Under most circumstances the output should be valid HTML, but if there are two chapters with the same number/title, (e.g. if a project contains two books in one document) it is possible to get duplicate IDs, which again will need human intervention.

Tasks done by autogeneration:
1. Trim trailing spaces
2. Adjust pagemark positions so they don't come mid-word
3. Convert entities such as ampersands
4. Convert the main body, spotting headings, block markup, poetry indents, etc.
5. Convert inline markup (e.g. using the dialog, you can have `<i>` converted to `<em class="italic">`
6. Convert smallcaps/allcaps
7. Convert footnotes & sidenotes
8. Add page anchors/page numbers
9. Add chapter divs around headings to support ebookmaker conversion
10. Wrap very long HTML lines

Unlike GG1, I think you can use Undo fairly safely after HTML generation.


rtonsing:
Autogenerated TOC needs '#' before IDs.

rtonsing:
This is current behavior in GG1, and just my opinion, but the /* */ markup insertion of 
`<span style="margin-left: x.xem;">`
is useless because they are usually tables that I will reformat anyway, so I delete them.

windymilla:
> Autogenerated TOC needs '#' before IDs.

Thanks - I'll take a look at that

windymilla:
> This is current behavior in GG1, and just my opinion, but the /* */ markup insertion of `<span style="margin-left: x.xem;">` is useless because they are usually tables that I will reformat anyway, so I delete them.

I agree, at least as far as my own PPing is concerned. However, I would prefer for now to mimic the GG1 behavior, and only consider removing it in future if there is substantial support for such a move. It's possible that in the same way the PPer can change /& to /P to get poetry markup instead of default /* behavior, that we might consider if a /T (for Table) markup might be useful. I'm not sure how much, if any help it could be, but it could be worth investigating.

For tables, it sounds like you'd probably prefer it if you changed the /*...*/ to /X...X/, which just puts a <pre> at the start/end of the table, which is easily deleted before you do the manual work of coding the table.

windymilla:
> Is there a way to designate the h1 text? Using h1 markup doesn't work.

Good catch - I've neglected to code that. I think GG1 assumes the first bit of text that isn't an illo, etc. is the title. I'm extracting that in GG2 to display in the dialog "title" field and the HTML header, but forgot to add the h1 markup in. Thanks.

windymilla:
Pushed commit to fix points raised in @rtonsing's review

rtonsing (approved):

srjfoo (approved):
I've exercised it a bit with a couple of test files, and what I see looks good. However, if there are still some subtle issues, I think they're only going to be found by active PPers actually using it on real projects.

PR #581, opened 2024-12-18
Simplify use of checker dialogs

windymilla:
Due to the way it developed, the Checkerdialog
base class required tools that used it to set certain padding and columnspan to align. Fix this in the
base dialog instead.

Apart from minor adjustments to alignment, there
should be no behavioral change.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -183,10 +183,8 @@ class CheckerDialog(ToplevelDialog):
             takefocus=False,
         ).grid(row=0, column=4, sticky="NSE", padx=2)
 
-        self.rerun_button = ttk.Button(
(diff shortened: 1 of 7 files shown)


srjfoo (approved):
image

The padding seems a bit unbalanced top to bottom. Spelling and Footnotes appear to be well-balanced -- in fact, everything else that I looked at had either a text-input field or a full-scale button rather than a radio button at the bottom (I doubt that I hit them all, but is that what's going on here?

Other than that, looks good to me, and you may consider that to be nit-picking.

windymilla:
Not too nit-picky - I think when we get nearer to, or perhaps even after, the first beta release, it would be good for someone to go through all the dialogs and try to make the padding consistent. All this PR did was to stop every developer having to add some padding of their own to try to match the Re-run button.

PR #587, opened 2024-12-20
Fix exception wrapping blank line in index

### LINKED ISSUE #585: Rewrap All error
Seems to just be this file, but I don't know where the issue is. Happens with both current code and alpha a11.
It should probably be caught and reported more helpfully.

1. Open file
2. Select Tools->Rewrap All

Windows 11 64-bit

image
[railways.zip](https://github.com/user-attachments/files/18201970/railways.zip)

windymilla:
For normal wrapping, the blank line is taken as a paragraph so is not itself wrapped. However, because indexes have been formatted into single whole lines, they are wrapped one line at a time. This accidentally included wrapping the blank lines. The wrapping routine hadn't been written to expect wrapping an empty string.

1. In lower wrapping routine, deal with an empty paragraph without raising an exception.
2. In upper index wrapping routine, only wrap lines that have text

Finally, while testing, I found an independent bug (of mine) which probably wouldn't occur in RL. I made a small index by deleting most of the lines from the example supplied, which left some page marks at the start of the closing `I/` line. It then failed to recognize this as the end of the index, and said "no match for opening `/I`". Fixed.

Fixes #585

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2293,7 +2293,7 @@ class MainText(tk.Text):
 
                     # Find matching close markup within section being wrapped
                     if close_index := self.search(
-                        rf"^{re.escape(block_type)}/\s*$",
+                        rf"^{PAGEMARK_PIN}*{re.escape(block_type)}/\s*$",
                         line_start,
                         stopindex=WRAP_END_MARK,
                         nocase=True,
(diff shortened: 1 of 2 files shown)


rtonsing (approved):
It works now. The /I explains why I didn't see an issue with the others I tried.

tangledhelix (approved):

PR #589, opened 2024-12-20
Fix column select failures

windymilla:
1. Column select now works in peer
2. Column select now works if text is scrolled right so that left-hand edge is off-screen
3. `rowcol` function now uses the focus widget, i.e. main text or peer. I thought previously it was OK to just use the main text, but if the index that is being converted to a rowcol is an `@x,y` string, it matters which widget you use to convert it. Symptom of
this failing was that if you scrolled the main text horizontally, when you started a column select in
the peer, the region actually selected didn't match the region dragged over - it was offset by the amount that main was scrolled horizontally. The function that has been changed is used *a lot*, but I'm hoping it won't have any effect on its other usage.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1531,7 +1531,7 @@ class MainText(tk.Text):
         maxlen = -1
         y_maxlen = -1
         for row in range(minrow, maxrow + 1):
-            geometry = self.bbox(f"{row}.{cur_rowcol.col}")
+            geometry = self.focus_widget().bbox(f"{row}.{cur_rowcol.col}")
             if geometry is None:
                 continue
             line_len = len(self.get(f"{row}.0", f"{row}.0 lineend"))
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -415,19 +415,6 @@ class MainText(tk.Text):
         self.aligncol = -1
         self.aligncol_active = tk.BooleanVar()
 
-        # Column selection uses Alt key on Windows/Linux, Option key on macOS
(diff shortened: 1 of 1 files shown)


tangledhelix (changes requested):
Column select seems to work in the main text pane as expected now (column 0 scroll out of view, or otherwise).

Unfortunately I see strange behavior in the peer. **Some** selecting is happening but it's selecting a block to the left, even if I'm dragging NW-SE or SW-NE ...

Screenshot 2024-12-21 at 12 06 14 AM

In this picture, I had started to option-drag from the `a` beginning `adipiscing`, and dragged down and to the right. But I ended up getting this selection.

srjfoo:
Agreeing with @tangledhelix -- it does do a column select in peer, but the selection goes to the left, even when dragging to the right. For me, the selection goes to the left edge of the screen (i.e. if the left edge is at c7, the column selection's left edge is c7, even if I started dragging from c28).

Column select now works in the main pane , but doesn't drag below the bottom of the pane, either in split or un-split mode, unless I combine the cursor positioning with a two finger drag (macOS -- probably equivalent to using a mouse-wheel in addition to the click-drag?).

If I do a column-select to just below the bottom or just above the top edge of the screen and let up on the mouse, the selection changes to a regular selection, and seems to be in an infinite selection loop. Just for fun, I started one at about line 700 and let it go. Sure enough, it didn't stop until it got to line 1.

tangledhelix:
LGTM (aside from scrolling past edge of screen; I will look at that but in a different PR).

srjfoo (approved):

Issue #586, opened 2024-12-20
Convert to Curly Quotes 


okrick:
This file contained four instances of double quote marks at the beginning of lines, likely intended as ditto marks. When I used the `Tools > Convert to Curly Quotes` function, these double quotes were converted to left double quotes, while all subsequent double quotes within the same paragraph were incorrectly reversed.

I'm unsure of the best approach for handling this in the future. My suggestion is to convert any isolated double quotation marks to `Double Low Nine Quotation Marks („)`. This allows for the remaining double quotes to be automatically assigned correctly.

I understand that this approach might not be universally preferred. However, I believe that instances of isolated double quotes that should have been corrected during the proofing stage should be infrequent. Regardless, the `„` marks can be easily located and replaced with the preferred character by the proofreader.

Perhaps the `„` character could be specified in GG2's preferences, allowing each proofreader to customize their workflow.

[belgian.zip](https://github.com/user-attachments/files/18204548/belgian.zip)

windymilla:
Thanks for the report Rick, and the example file with which I was able to reproduce what you see. 

I have a counter-suggestion as explained in the points below (not necessarily in order of importance) 

1. GG1 behaves the same as GG2 currently does, but I think we can do better than that
2. I am reluctant to silently assume that a double quote at the start of a line with a space after it is an intended ditto mark. I think that would risk missing the relatively common occurrence of speech starting at the beginning of a line, but with a scanno-ed space appearing after it.
3. There is already code to detect ditto marks (like GG1), and the rule is that the double quotes must have two spaces either side. In that case (as you can see by editing your file prior to curly quote conversion and adding 2 spaces before plus an extra space after those ditto marks) the double quotes are converted to close curly quotes, and there is no problem with the remaining quotes being assigned correctly because it knows the first one was a ditto mark.
4. I believe most PPers use close curly quotes for ditto marks, and if they don't, they can easily search for `  ”  ` (double spaced close quotes) and replace them with their preferred character.
5. HOWEVER, where the ditto marks appear at the start of the line, I think it would be very reasonable to say that they do not have to be preceded by two spaces, **as long as they are followed by 2 spaces** (to avoid my concern in point 2 above). A PPer of your file would therefore need to add an additional space after those ditto marks to flag that they really were ditto marks and not just open quotes that the OCR had added a space to.

[...]

PR #590, opened 2024-12-21
Detect ditto marks better in curly quotes

windymilla:
Previously, only quotes that had 2 spaces either side were converted to ditto marks. Double quotes are now
interpreted as ditto marks if they are at start of line (with optional space before) and have 2 spaces after, or if they are at end of line (with optional space after) and have 2 spaces before.

This compromise is to try to detect ditto marks without letting badly spaced quotes pass by undetected, hence the requirement for a double space before/after/

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -1745,11 +1745,18 @@ def convert_to_curly_quotes() -> None:
             continue
 
         # Apart from special cases, alternate open/close double quotes through each paragraph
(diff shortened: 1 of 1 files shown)


windymilla:
Fixes #586

windymilla:
Testing notes:

Linked issue has this file: https://github.com/user-attachments/files/18204548/belgian.zip

srjfoo (approved):

PR #591, opened 2024-12-21
Add "Copy Results" button to checkers & WF

windymilla:
Button copies the list of results to the clipboard. This can be done with Ctrl-A & Ctrl-C, but that's a bit obscure, and at least one person didn't realize.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -140,15 +140,15 @@ class CheckerDialog(ToplevelDialog):
         super().__init__(title, **kwargs)
         self.top_frame.rowconfigure(0, weight=0)
         self.header_frame = ttk.Frame(self.top_frame, padding=2)
-        self.header_frame.grid(column=0, row=0, sticky="NSEW")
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
Does what it says on the box, so I'm going ahead and approving it.

However, I honestly feel like we'd be better off documenting the select-all/copy/paste facility than adding a button that's not needed. It's easier to add it later as a future feature once we're more feature complete than it is to take it away at a later time because you need the space for something else.  Besides, now it looks more cluttered.

Because Guiguts 1 does it this way does not strike me as a good reason to continue doing it this way.

tangledhelix:
> Because Guiguts 1 does it this way does not strike me as a good reason to continue doing it this way.

Does Guiguts 1 do it this way? I see no such function in WF in GG1. I tried the copy button in Spell Query but GG1 crashed the first time, and in a later test didn't crash, but doesn't seem to have done anything either. (Certainly the spell query results weren't in my pastebuffer afterward).

I was going to ask about a right-click menu, but some of these tools use right-click as a way to clear items out of the list so that probably isn't possible.

PR #597, opened 2024-12-24
Recognize ppgen `.bn` commands

windymilla:
This is to allow a PPer to load a ppgen file that has `.bn` commands in, and have page image files associated with the correct place in the file to aid continued work on the file.

It is not a solution to the issue of ppgen output files then being loaded by a PPVer using GG2.

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -758,7 +758,7 @@ class File:
         Returns:
             True if `.bn` commands were present.
         """
-        regex = r"^ *\.bn *(.+?)\.png"
+        regex = r"^\.bn +(.+?)\.png"
         flag_matches = maintext().find_matches(
             regex,
             IndexRange(maintext().start(), maintext().end()),
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -247,7 +247,9 @@ class File:
         self.languages = preferences.get(PrefKey.DEFAULT_LANGUAGES)
         bin_matches_file = self.load_bin(filename)
         self.mark_page_boundaries()
-        flags_found = self.update_page_marks_from_flags()
+        flags_found = (
+            self.update_page_marks_from_flags() or self.update_page_marks_from_ppgen()
+        )
         if not bin_matches_file:
             # Inform user that bin doesn't match
             # If flags are present, user has used page marker flags to store the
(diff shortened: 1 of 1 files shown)


src/guiguts/file.py
```diff
@@ -747,6 +749,46 @@ def update_page_marks_from_flags(self) -> bool:
...
+    def update_page_marks_from_ppgen(self) -> bool:
+        """If `.bn` ppgen commands in file, replace all existing page marks.
+
+        Command is of the form: `.bn a001.png // optional comment`
+        We only want `a001`
+
+        Returns:
+            True if `.bn` commands were present.
+        """
+        regex = r"^ *\.bn *(.+?)\.png"
```
tangledhelix, line 761: ppgen dot-commands **must** have the dot as the first character of the line. So the leading ` *` right after the caret should be removed.

After `bn` is ` *` for "zero or more spaces". Here, zero spaces would be invalid, so this should become ` +?`.

New regex: `^\.bn +?(.+?)\.png`

src/guiguts/file.py
```diff
@@ -747,6 +749,46 @@ def update_page_marks_from_flags(self) -> bool:
...
+            # Will always match because same regex as above
+            assert extract is not None
+            img = extract[1]
+            # Set mark and store page info
+            maintext().set_mark_position(page_mark_from_img(img), match.rowcol)
+            self.page_details[img] = PageDetail(match.rowcol.index(), style, number)
+            # Subsequent pages are same style as first, page numbers increment
+            style = STYLE_DITTO
+            number = NUMBER_INCREMENT
+        maintext().set_modified(True)
```
tangledhelix, line 789: Is this here so that the file is marked dirty in order to save a new .json? The text file itself hasn't been modified, correct?

tangledhelix (changes requested):
Looks good mostly (at least, in my head - I didn't try to test). If you make the regex update then I'll try with some of my existing ppgen project files.

src/guiguts/file.py
windymilla: Ah, I didn't know about the first point, but I did know about the second - thanks for catching it!

windymilla: Fixed

src/guiguts/file.py
tangledhelix: It seems appropriate, I just asked for my own understanding (and to verify the text file didn't change, because I don't think this operation should touch the text itself).

tangledhelix:
Tested with my [pascal project](https://github.com/tangledhelix/DP_pascal/blob/main/pascal-src.txt), works like a charm! Ship it 🚀

tangledhelix (approved):

PR #598, opened 2024-12-24
Add Regex Cheat Sheet to Help menu

windymilla:
Opens manual page in browser

(As requested in [DP forum](https://www.pgdp.net/phpBB3/viewtopic.php?p=1355160#p1355160))


srjfoo (approved):

tangledhelix (approved):
Works perfectly on Mac.

On my Ubuntu VM nothing happens, but that's probably just that my browser isn't set up correctly or something. I don't know the correct way to set that up and will assume the `webbrowser` module does the right thing on Linux and the error is on my part.

PR #603, opened 2024-12-24
Add First/Last button to Search dialog

windymilla:
The First/Last button finds the first occurrence of the string in the file if searching forwards, and the last occurrence if searching backwards.

Note that searching backwards can be done by turning on "Reverse" checkbox. The currently selected value of "Reverse" is itself reversed if the First/Last button is Shift clicked, just like the Search button.

So, if "Reverse" is off, click First/Last for the first match, shift-click it for last match. If "Reverse" is on, click First/Last for last match and shift-click it for first match.

Since this behavior is identical to the Search button behavior, it should be intuitive to the user. If not, they don't need to use the Shift key and can just use the Reverse checkbox.

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -136,6 +136,20 @@ class SearchDialog(ToplevelDialog):
             "<Shift-Return>", lambda *args: self.search_clicked(opposite_dir=True)
         )
 
+        # First/Last button - find first/last occurrence in file
+        first_button = ttk.Button(
+            self.top_frame,
+            text="First/Last",
+            takefocus=False,
+            command=lambda *args: self.search_clicked(first_last=True),
+        )
+        first_button.grid(row=2, column=2, padx=PADX, pady=PADY, sticky="NSEW")
+        mouse_bind(
+            first_button,
+            "Shift+1",
+            lambda *args: self.search_clicked(first_last=True, opposite_dir=True),
+        )
+
         # Count & Find All
         self.count_btn = ttk.Button(
             self.top_frame,
@@ -320,11 +334,14 @@ class SearchDialog(ToplevelDialog):
         self.search_box.icursor(tk.END)
         self.search_box.focus()
 
-    def search_clicked(self, opposite_dir: bool = False) -> str:
+    def search_clicked(
+        self, opposite_dir: bool = False, first_last: bool = False
+    ) -> str:
         """Search for the string in the search box.
 
         Args:
             opposite_dir: True to search in opposite direction to reverse flag setting
+            first_last: True to begin search at start/end of file
         Returns:
             "break" to avoid calling other callbacks
         """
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Seems to work just fine.

Personal opinion, only: I honestly don't see a need for it. It's a one-button shortcut for home/end (cmd+up/down arrow) plus find again. I'm far from a minimalist, but I'd rather limit the clutter until we need something. So, it's something of a design philosophy, I guess. You've got users wanting it because they had it before, and it's always harder to take something away than it is to not implement it in the first place.

tangledhelix (changes requested):
After reading the forums I'll have to disagree with Sharon's review. It seems to be something the users want.

However, as implemented I find it confusing. Reading your explanation I understand how it works, but my 2¢ is it would be better simplified:

- Label the button "First"
- On click, it goes to the first match
- No shift-click binding

I could see that if Reverse is checked, going to Last instead could make sense. Is it possible to make the label on the button reactive to the state of the Reverse checkbox, such that if Reverse is checked, the label is Last, but otherwise it's First?

I'd be just as happy not doing that, and having a no-frills "First" button.

srjfoo:
> After reading the forums I'll have to disagree with Sharon's review. It seems to be something the users want.

Part of my review was my personal opinion, and the main reason that I'm not in favor of it is that I just don't see a real need for it -- it adds clutter, and it's not really necessary.

I've accepted that it will probably be implemented, and I hate to be the one voting against some things. My main concern is the "give a mouse a cookie" syndrome (aka "slippery slope"?). The more things like this that are implemented, the more will be requested, and the interface will be nickeled and dimed to death.

That being said, @windymilla has certainly implemented a lot of things that I've asked for, as have you.

windymilla:
Agree with @tangledhelix that it's unnecessarily complicated (though actually simpler to implement) with the Shift-to-reverse option. I'll put up another commit that removes that feature.

windymilla:
> Is it possible to make the label on the button reactive to the state of the Reverse checkbox, such that if Reverse is checked, the label is Last, but otherwise it's First?

My latest commit does this, and removes the Shift-to-reverse complication.

srjfoo:
> > Is it possible to make the label on the button reactive to the state of the Reverse checkbox, such that if Reverse is checked, the label is Last, but otherwise it's First?
> 
> My latest commit does this, and removes the Shift-to-reverse complication.

I'm not seeing this -- it still says `First/Last`. I tried disabling Wrap Around and multi-replace, but those had no effect. Did I misunderstand?

srjfoo (approved):

tangledhelix (approved):

windymilla:
> I'll admit that I'm really missing the shift-for-reverse,

If a user says, "hey, I can shift-click on Search, why can't I do that on First/Last?" we can add it in a heartbeat

PR #612, opened 2024-12-28
Avoid useless redraws of column ruler

tangledhelix:
- Don't redraw ruler when pref is disabled
- Don't redraw peer's ruler when window not split

Also a couple of whitespace changes to mollify pylint.


windymilla (commented):
LGTM

Issue #238, opened 2024-05-13
Click-drag to select stops at edge of text window on macOS.


srjfoo:
Tested in macOS Sonoma.

Click anywhere in the text window, hold down the mouse button, and scroll (either up or down). The selection is stopped at the edge of the page, and the page will not advance unless the scroll wheel is used in addition (can be mimicked using the trackpad: click-hold-with-thumb and one-finger drag to simulate the click-drag of the initial selection, then two-finger-drage once the edge of the page is reached. (Also, when doing this, you have to twitch the mouse for the selection to show up.

windymilla:
Works OK on Windows

windymilla:
Can you click where you want the selection to start, then scroll to the place you want it to end, then Shift-click there?

windymilla:
Also, do you see this behavior? https://discussions.apple.com/thread/255386688?sortBy=best

srjfoo:
> Can you click where you want the selection to start, then scroll to the place you want it to end, then Shift-click there?

Yes, in gg2. I had to stop and think for a moment, because that doesn't work in gg1 (which I was playing with, for comparison, yesterday).

srjfoo:
> > Also, do you see this behavior? https://discussions.apple.com/thread/255386688?sortBy=best
> 
> I don't know if that exactly describes what I was seeing, but the major part, with the selection stopping at the edge when clicking and dragging, is what I see.
> 
> **_But_** I'm not seeing the behavior described by that post in either Notes or Preview.

Confirmed just now that I'm seeing the same click-drag_to_select-stop_at_edge_of_window behavior under Ventura on an Intel Mac.

PR #614, opened 2024-12-29
Fix redraw handling (regression bug)

### LINKED ISSUE #613: Redraws not properly handled on mouse click to move cursor
When placing the cursor with a mouse click, the line & column rulers are not redrawn (and other redraws also don't happen, e.g. quote/bracket highlights).

This bug was introduced in a54e3465b6cc971be7a5fd67b180d26bab410221 from #238 when Button-1 (left-click) was overridden to fix scrolling past the edge of the window while selecting text.

tangledhelix:
In #238, binding of button-1 (left-click) created a regression that didn't fire redraws on click properly.

Fixes #613

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -3215,24 +3215,28 @@ class MainText(tk.Text):
     ) -> None:
         """When Button-1 is pressed, turn mouse-drag monitoring on"""
         self._autoscroll_active = True
+        self._on_change()
 
     def _autoscroll_stop(
         self, event: tk.Event  # pylint: disable=unused-argument
     ) -> None:
         """When Button-1 is released, turn mouse-drag monitoring off"""
         self._autoscroll_active = False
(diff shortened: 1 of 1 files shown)


tangledhelix:
Testing notes:

Click in the file to move the cursor (and do nothing else). The line & column ruler should update. If quotes/brackets mode is turned on, those highlights should also update.

Without this patch applied, clicking doesn't redraw properly. You have to move with the arrow keys or scroll or take some other action than clicking in order to trigger a redraw.

tangledhelix:
It's probably also worth re-testing that select-dragging past the window edge still works like it should, too.

srjfoo (approved):
Select/drag beyond the edge of the screen (NSEW) still works for normal select. Column-select beyond the edge of the screen still doesn't scroll (which was expected -- I just wanted to confirm that I tested it).

windymilla (approved):

PR #619, opened 2024-12-29
Split checker dialogs so they all get their own settings

windymilla:
The common settings (sorting method and suspects-only) and geometry (position & size) used to be shared amongst most checker dialogs, e.g. Search and Footnote Fixup. There were a few exceptions, like Find All. 
They should all now be individual to the dialog, so you can position each one where you like, make it an appropriate width for the type of messages it shows, turn on/off suspects-only, and change the sort method without affecting the other dialogs.

Testing notes: Illo Fixup and Sidenote Fixup were previously very closely linked and shared their data, so needed extra work to extricate them, so please especially check they are working independently of one another.


tangledhelix (approved):
All the dialogs I tried remembered size/position, seems to work. A few parts of the code I didn't quite understand but in general it looks like what I expected.

srjfoo (approved):

PR #620, opened 2024-12-30
Column-select scrolling at edge of window

### LINKED ISSUE #596: Column selection scrolling bug
In column select mode, scrolling doesn't work properly.

This may be limited to Mac (need to test Linux again).

tangledhelix:
Adds support for scrolling past the window edge in column select mode.

It's currently a bit imperfect / visually choppy; possibly that could be improved. But it seems to function properly.

I took the opportunity to simplify the non-column select mode scrolling to use less code. It should work the same as before, except I made the vertical scrolling a bit faster.

As before, the non-column select mode scrolling is only for Mac (because it isn't needed for Win/Linux).

The column select scrolling code is currently active on **all platforms**.
- Mac: because it's needed
- Windows: unknown (to me) - would appreciate feedback as to whether it was needed in the first place
- Linux: scrolling technically functions without this code, but it's **painfully** slow, so I opted to leave it on for Linux to override the normal behavior.

Testing notes:
- Re-test normal selection, going past the window edge, to make sure it still works as expected
- Do the same in column select mode; it should behave similar to normal selections as far as scrolling (literally the same code mostly).

And I'm particularly interested in @windymilla results on Windows with this branch compared to master. I did no testing on Windows.

Fixes #596

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -3266,27 +3266,20 @@ class MainText(tk.Text):
         if 0 <= widget_x <= width and 0 <= widget_y <= height:
             return
 
-        delay = 50
+        scroll_delay = 100
 
-        # Scroll in the appropriate direction (x or y).  If scrolling
-        # vertically, set the delay before the next loop a little longer.
-        # (Because that "feels right".)
+        # Scroll in the appropriate direction (x or y).
         if widget_y < 0:
             event.widget.yview_scroll(-1, "units")
-            delay = 150
         elif widget_y > height:
             event.widget.yview_scroll(1, "units")
-            delay = 150
         elif widget_x < 0:
             event.widget.xview_scroll(-1, "units")
-            delay = 100
         elif widget_x > width:
             event.widget.xview_scroll(1, "units")
-            delay = 100
 
-        self.after(delay, lambda: callback(event))
         self._on_change()
-        self.after(delay, lambda: self._autoscroll_callback(event))
+        self.after(scroll_delay, lambda: self._autoscroll_callback(event))
 
 
 def img_from_page_mark(mark: str) -> str:
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1584,6 +1584,9 @@ class MainText(tk.Text):
 
         self.do_column_select(IndexRange(self.rowcol(TK_ANCHOR_MARK), cur_rowcol))
 
+        # Handle scrolling past edge of widget
+        self._autoscroll_callback(event)
+
     def column_select_release(self, event: tk.Event) -> None:
         """Callback when column selection is stopped via mouse button release.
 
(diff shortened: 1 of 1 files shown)


windymilla (commented):
I think this may be the best we can manage (easily) on Windows:
1. It still flashes normal selection/column selection/normal selection/column selection while scrolling (same as master)
2. It scrolls more quickly than master (which is unusable) - it's slightly disorienting because of point 1, and all the flashing makes you think it has scrolled really quickly, but in fact the speed is OK
3. If you release the mouse button outside the window, it switches to normal selection and just scrolls slowly for ever (until you click back in the main window).

So the only improvement over master is point 2, but even that is worth having.

Point 3 is a nuisance, but I'm quite willing to believe it's not practical for us to detect the mouse release when it is outside the application (thought it would be possible to detect the mouse leaving the window, if that would help)

I wondered if you bound the mouse release to self.root instead of the text widget (or as well as?) whether it would at least detect a release that was still inside the whole application, even if outside the text window, e.g. if your mouse was over the status bar when released. However when I made the obvious 1 line addition, it didn't work.

There are other possibilities, but nothing I could think of was smooth and seamless. Unless you have any other ideas

It may be that this is another thing that should go in the "Known Issues" section.

srjfoo:
> It's currently a bit imperfect \/ visually choppy; possibly that could be improved. But it seems to function properly.

For what it's worth, it's not just the column-select that is visually choppy. For me, both normal select, and column select drag beyond the window edge with no problem anymore (thinking mostly vertically now, because there just isn't a long way to drag horiziontally), **_but_** the highlighting won't show unless:

* you move the cursor back inside the window, or
* you twitch the mouse sideways a column while still below the window.

---

> 3\. If you release the mouse button outside the window, it switches to normal selection and just scrolls slowly for ever (until you click back in the main window).

For what it's worth, it's the same on macOS. I was hoping that would be fixed when the scrolling was fixed, but I guess not.

At least it's a Known Issue in both macOS and Windows -- don't know about Linux.

srjfoo (approved):

windymilla:
Added comment to known issues - I think what we have is good enough that it's not worth putting more effort into it at this point.

windymilla (approved):

tangledhelix:
> 3. If you release the mouse button outside the window, it switches to normal selection and just scrolls slowly for ever (until you click back in the main window).
> 
> Point 3 is a nuisance, but I'm quite willing to believe it's not practical for us to detect the mouse release when it is outside the application (thought it would be possible to detect the mouse leaving the window, if that would help)
> 
> I wondered if you bound the mouse release to self.root instead of the text widget (or as well as?) whether it would at least detect a release that was still inside the whole application, even if outside the text window, e.g. if your mouse was over the status bar when released. However when I made the obvious 1 line addition, it didn't work.

That's the thing - if you add some debug prints, the release event does fire even if you release outside the window, and the scroll callback stops running. Yet the scrolling happens. I can't tell what's driving it unless it's some sort of a default behavior... I'll keep tinkering with it.

windymilla:
> some sort of a default behavior

I think that's exactly it - it thinks it's doing normal selection dragging (which is built-in)

PR #621, opened 2024-12-30
Fix "move to para" bug with multi-para FNs

### LINKED ISSUE #611: Move FNs to Paragraphs breaks page markup
I joined the 2 footnotes on separate pages then reindexed them and saved the result below.

When I clicked on `Move FNs to Paragraphs` two of the page break markers were broken.

```
controversies which are settled, if at all, not by being decided one
way or the other, but by becoming obsolete.]-----File: 029.png---------------------------------------------------------

By way of illustrating these general statements I
```
```
I shall try to show the nature of egotism and selfishness in Chapter
VI.]-----File: 103.png---------------------------------------------------------
third class called, perhaps, ego-altruistic, is empty;
```

[projectID64deb99d98161.zip](https://github.com/user-attachments/files/18264464/projectID64deb99d98161.zip)

windymilla:
Move footnotes to paragraph end failed if the footnote itself had several paragraphs. Fixed by Quentin.

Fixes #611

### DIFF
--- a/src/guiguts/footnotes.py
+++ b/src/guiguts/footnotes.py
@@ -369,10 +369,13 @@ class FootnoteChecker:
         the corresponding footnote will be moved. There may be many marks at the
         same location.
 
-        The insertion point for marks is immediately before the line that separates
+        The insertion point for marks is immediately BEFORE the line that separates
         the end of a paragraph from all that follows. That line is either a blank
-        line or a Page Marker. Care is taken to ensure that all marks are located
-        on the same page as the paragraph that they follow.
+        line that is not followed by a '[Footnote ...' line or is a Page Marker if
+        that immediately precedes such a blank line. In this way care is taken to
+        ensure insertion marks are located on the same page as the paragraph that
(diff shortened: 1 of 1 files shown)


srjfoo:
Unless I'm missing something very basic, this does not fix the issue.

Test file used: [Aggravating ladies](https://www.pgdp.net/c/project.php?id=projectID64c1772169c99). It's a fairly short project, with 30 footnotes and 3 continuation footnotes, one of which (footnote 16) has 4 paragraphs. I'm still getting 
image
for that one footnote when moving the footnotes.

Initially, I was mistakenly testing after using page separators, and at that exact point (i.e. footnote 16), when I added page marker flags after moving the footnotes, just to see where that one fell, the page marker flage was between the paragraph where the anchor was and the footnote (i.e. the footnote was at the top of the next "page"). The blank line that should have been between the footnote and the following paragraph was missing.

windymilla:
@srjfoo - ready for you to test again. Thanks

srjfoo (approved):

PR #625, opened 2024-12-31
Fix illo/sn move bug if file edited after tool run

### LINKED ISSUE #623: Illo/SN fixups can move the wrong line
If the file is edited such that the line number of the illo/sn to be moved no longer matches what is in the Illo/sn dialog, the  wrong line gets moved.

1. Load file
2. Select an illo to be moved
3. Delete 3 lines from higher up in the file
4. Move the illo (without re-running illo fixup)
5. Instead of the illo being moved, a line which is 3 lines away is moved instead.

NTS: Caused by using cached line numbers, rather than using the checker dialog marks

windymilla:
Code was using out of date line numbers - altered to
refresh the internal line numbers using the checker
dialog marks.

NTS: Merge after #619 

Fixes #623


windymilla:
Testing notes: see linked issue #623

srjfoo (approved):

PR #626, opened 2024-12-31
Fix select-auto-scroll problem on out-of-bounds release

tangledhelix:
When select-dragging beyond the widget boundary, releasing the click would result in the selection converting back to linear, and scrolling infinitely until clicking back into the window.

This was caused by the event falling back to the default, due to bind_event() suppressing the original event handler. Change this to an additional binding to allow the normal release behavior to be processed.

This commit also adds a check to stop calling the scroll handler when the boolean is False, moves the boolean being set into the click and release handlers (instead of deeper down where it also affected the shift-click mechanism), and uses the event widget reference instead of relying on focus_widget as it was doing before.

This is a followup to #596 and #620.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -573,13 +573,12 @@ class MainText(tk.Text):
             f"<{modifier}-B1-Motion>", self.column_select_motion, bind_peer=True
         )
         self.bind_event(
-            f"<{modifier}-ButtonRelease-1>", self.column_select_release, bind_peer=True
-        )
-        self.bind_event(
-            "<KeyRelease-Alt_L>",
-            lambda _event: self.column_select_stop(),
+            f"<{modifier}-ButtonRelease-1>",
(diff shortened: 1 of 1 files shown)


tangledhelix:
Testing notes:
* Re-test select and column select when you drag past the edge of a window (should scroll)
* Re-test releasing the mouse click outside the text area (it should stop scrolling and have a proper selection)
* Shift-click to extend selection should still work
* Shift-Opt-click (Mac) or Shift-Alt-click Win/Linux) should extend a selection in column-select mode
* When dragging a column select, the cursor should be a crosshair
* This should all work in split window mode too

tangledhelix:
Seems to work as expected on Ubuntu.

windymilla (approved):
That's fixed it for me! Thanks for persevering with this

srjfoo (approved):
Fixed for me, too!

PR #634, opened 2025-01-01
Use consistent definition of word in WF

### LINKED ISSUE #628: Word Frequency fails to position cursor
Capture

Ran the Alpha/Num check.  Except for `000.jpg`, selecting any item on the list automatically positions the cursor at the beginning of the item in the document.

[travels.zip](https://github.com/user-attachments/files/18283866/travels.zip)

windymilla:
Because regexes include underscore in their idea of word characters, an occurrence preceded by underscore of a word found by WF would not be found by "whole word" search. So, the word would be listed in WF, but when the user clicked it, the word would not be found.

Fixed by changing search's "whole word" to wrap the search string with a more complex regex, rather than `\b...\b`

Also changed WF's idea of a word to include embedded underscores (in the same way as embedded periods are included), e.g. "i_001.jpg", but to strip leading or trailing underscores, which signify italics, e.g. "_dog_".

Fixes #628

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2149,8 +2149,13 @@ class MainText(tk.Text):
                     slurp_text = slurp_text + "\n"
         else:
             search_string = re.escape(search_string)
+        # Don't use \b for word boundary because it includes underscore
         if wholeword:
-            search_string = r"\b" + search_string + r"\b"
+            search_string = (
+                r"(?=[[:alnum:]])(?<![[:alnum:]])"
+                + search_string
+                + r"(?<=[[:alnum:]])(?![[:alnum:]])"
+            )
         # Preferable to use flags rather than prepending "(?i)", for example,
         # because if we need to report bad regex to user, it's better if it's
         # the regex they typed.
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -101,11 +101,11 @@ class WFWordLists:
                 line = line.lower()
             line = re.sub(r"<\/?[a-z]*>", " ", line)  # throw away DP tags
             # get rid of nonalphanumeric (retaining combining characters)
-            line = re.sub(r"[^'’\.,\p{Alnum}\p{Mark}*-]", " ", line)
+            line = re.sub(r"[^'’\.,\p{Alnum}\p{Mark}*_-]", " ", line)
 
             def strip_punc(word: str) -> str:
                 """Strip relevant leading/trailing punctuation from word."""
-                return re.sub(r"^[\.,'’-]+|[\.,'’-]+$", "", word)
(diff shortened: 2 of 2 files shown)


okrick:
I'm pleased you found a solution for GG2. However, I disagree with the assessment that GG1 was also inconsistent. 000.jpg is present in the GG1 WF Alpha/num list, and double-clicking it successfully locates the subsequent occurrence of 000.jpg.

windymilla:
> I'm pleased you found a solution for GG2. However, I disagree with the assessment that GG1 was also inconsistent. 000.jpg is present in the GG1 WF Alpha/num list, and double-clicking it successfully locates the subsequent occurrence of 000.jpg.

Yes, GG1 was consistent between the WF word list and clicking on it, and that is what is being fixed in GG2. 

However, I was pointing out that the inconsistency between whole word search and `\b` search was the same as GG1, and is not being fixed in GG2 (at time of writing). To see this inconsistency in GG1, bring up the Search dialog and type `001.jpg` into the search field. Click Whole Word, and search - it will find those occurrences. Now, click Regex instead of Whole Word and change the search term to `\b001jpg\b`, and you will find the search fails.

srjfoo (approved):
> Note that one (I think inconsequential) consequence of this PR is that doing a Whole Word search for `dog` will find a match in `_dog` or `_dog_` or `dog`, but attempting the "same" thing manually by doing a Regex search for `\bdog\b` will not find the occurrences with underscores, and will only find `dog`. This is because `\b` considers underscore to be a word character, so there is no word break between `_` and `dog`. This is not fixable, except by inspecting the regex in the code, detecting use of `\b` (but avoiding `\\b`, but not avoiding `\\\b`, etc!) and replacing it with something more complicated. I don't think this would be a good use of time.

Should this potentially be documented in known issues? Or just in the WF section of the manual?

windymilla:
> Should this potentially be documented in known issues? Or just in the WF section of the manual?

Thanks. Added to Known Issues. It's not really a WF thing since WF behavior is now self-consistent - it's a Search dialog thing.

okrick:
Upon further review, this behavior might simply be a documentation issue. It could be confusing, but likely not widespread. Unless, of course, numerous PPers frequently use WF to edit UTF-8 files where underscores are intended to represent italics.

windymilla:
> Upon further review, this behavior might simply be a documentation issue. It could be confusing, but likely not widespread. Unless, of course, numerous PPers frequently use WF to edit UTF-8 files where underscores are intended to represent italics.

I've added a bit to the manual to describe the unavoidable regex inconsistency (`\b` compared with Whole Word) which shouldn't generally be a problem for people. I think what we now have as a consequence of your report is better than we had previously, particularly for the case you describe of underscores for italics, which non-ppgen PPers would quite likely be doing.

PR #635, opened 2025-01-01
Implement F1 key to bring up manual page

windymilla:
In any dialog (or main window) press F1 function key to bring up a relevant manual page.

Note that if your function keys are set up to perform other actions by default, you may need to press `<fn>+F1`, or something similar.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -805,7 +805,9 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
     def init_help_menu(self) -> None:
         """Create the Help menu."""
         menu_help = Menu(menubar(), "~Help")
-        menu_help.add_button("Guiguts ~Manual", self.show_help_manual, "F1")
+        menu_help.add_button(
+            "Guiguts ~Manual", self.show_help_manual, "F1", force_main_only=True
+        )
         menu_help.add_button("About ~Guiguts", self.help_about)
         menu_help.add_button("~Regex Quick Reference", self.show_help_regex)
         menu_help.add_button(
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -805,7 +805,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
     def init_help_menu(self) -> None:
         """Create the Help menu."""
         menu_help = Menu(menubar(), "~Help")
-        menu_help.add_button("Guiguts ~Manual", self.show_help_manual)
+        menu_help.add_button("Guiguts ~Manual", self.show_help_manual, "F1")
         menu_help.add_button("About ~Guiguts", self.help_about)
         menu_help.add_button("~Regex Quick Reference", self.show_help_regex)
         menu_help.add_button(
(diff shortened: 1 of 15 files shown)


src/guiguts/widgets.py
```diff
@@ -65,7 +66,11 @@ def __init__(
 
         self._do_config()
         self.save_config = False
-        self.bind("<Configure>", self._handle_config)
+        self.bind("<Configure>", self._handle_config)  #
```
tangledhelix, line 69: Why the trailing `#` here? Seems an unnecessary change...

src/guiguts/application.py
```diff
@@ -805,7 +805,7 @@ def init_view_menu(self) -> None:
     def init_help_menu(self) -> None:
         """Create the Help menu."""
         menu_help = Menu(menubar(), "~Help")
-        menu_help.add_button("Guiguts ~Manual", self.show_help_manual)
+        menu_help.add_button("Guiguts ~Manual", self.show_help_manual, "F1")
```
tangledhelix, line 808: I think this being in the menu means it is fired even when the main window isn't focused. See my main comment about multiple pages being opened.

Perhaps instead of binding here in the menu, remove this menu-based binding and put an F1 bind on the main window.

tangledhelix (changes requested):
If I have any dialogs open, I have two pages opened in the browser when I press `F1`: one for the dialog in question, and another at the main page of the GG manual. I'm guessing it's because the Help menu has a binding to F1, and each dialog also has a binding.

If I open many dialogs, I still only get two pages opened: the focused dialog's help page, and the main app help page. So it seems that at least for dialogs, it honors what is focused.

srjfoo:
> If I have any dialogs open, I have two pages opened in the browser when I press `F1`: one for the dialog in question, and another at the main page of the GG manual. I'm guessing it's because the Help menu has a binding to F1, and each dialog also has a binding.
> 
> If I open many dialogs, I still only get two pages opened: the focused dialog's help page, and the main app help page. So it seems that at least for dialogs, it honors what is focused.

I actually really, really like having the manual page/section for the focus window opened. I'd rather see the focus window get the only page, though. If I have two tools open, and I bring up the manual entry for both, I get two copies of the top of the main page. If the main text window has focus, it makes sense for it to bring up to the beginning of the document.

windymilla:
Sorry guys. It was a typical last minute edit failure. I did all the dialogs and was just about to commit, then thought, "I should add F1 to the main window too". So I did and checked it worked and committed without going back to find it had broken the dialog behaviour. I'll fix it.

src/guiguts/widgets.py
windymilla: Oops - thanks

src/guiguts/application.py
windymilla: Decided I'd prefer the "F1" accelerator to continue to appear in the menu (if only to remind users which key it is to get help), so I've added an arg to `add_button` to force binding only to main window (which it already did for inbuilt virtual events anyway)

windymilla:
Rebased. Should all be working now. If focus is in a dialog, it should open the man page for that dialog. If focus is in the main window, it should open the top level man page. I made is so that the shortcut string "F1" continues to appear in the Help Menu, Manual entry, to remind the user which key it is.

srjfoo (approved):

tangledhelix (approved):

PR #637, opened 2025-01-02
Lower priority of SPOTLIGHT highlight tag

### LINKED ISSUE #631: Improving Cursor Visibility in Hyphenation Check (and other checks using a siimilar window)
When performing a hyphenation check (and other checks using the same style of window), selecting an item in the list highlights the corresponding text in the document. However, the cursor within the highlighted area is difficult to distinguish due to the overlapping colors. This makes it challenging to precisely select and edit specific characters, such as hyphens, within the highlighted text.

To enhance usability, it would be beneficial to implement a different color for the cursor when it is positioned within the highlighted text. This improved color contrast would make the cursor more visible, enabling users to accurately select and edit individual characters within the highlighted region.

tangledhelix:
Testing notes:

Load a file and run a WF check, e.g. hyphens (as noted in the original ticket). Clicking on a match will take you to that text in the file and highlight with SPOTLIGHT (e.g. orange background in light theme). Select one or more characters in the text. On the master branch, the cursor disappears and you can't see what's selected. On this branch, you can see what you've selected. (Should be the same if you select with keyboard or mouse.)

Fixes #631

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -3218,10 +3218,10 @@ class MainText(tk.Text):
         # ** THE ORDER MATTERS HERE **
         #
         for tag, colors in (
-            (HighlightTag.SPOTLIGHT, HighlightColors.SPOTLIGHT),
             (HighlightTag.QUOTEMARK, HighlightColors.QUOTEMARK),
             # "sel" is for active selections - don't override the default color
             ("sel", None),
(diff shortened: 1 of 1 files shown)


tangledhelix:
Additional note to avoid confusion: the cursor disappearing is normal; that is not changed by this PR. Only the visibility of what text is selected has changed.

srjfoo (approved):

windymilla (approved):
Perfect!

PR #638, opened 2025-01-03
Fix a couple column selection bugs

windymilla:
1. Exception on non-Mac platforms when using column select for first time due to non-initialized variable
2. Beginning drag before pressing modifier key didn't initialize everything that was needed, so when you dragged outside the window it reverted to slow
continuous scroll (mac-only)

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1521,7 +1521,7 @@ class MainText(tk.Text):
         return "break"  # Skip default behavior
 
     def column_select_click_action(self, start: IndexRowCol) -> None:
-        """Do that needs to be done when the user clicks to start column selection.
+        """Do what needs to be done when the user clicks to start column selection.
         Since user can click first then press modifier key, or hold modifier while
         clicking, there are two routes into this point.
 
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -790,7 +790,9 @@ class MainText(tk.Text):
             self.bind_event(
                 "<B1-Motion>", self._autoscroll_callback, add=True, bind_peer=True
             )
-            self._autoscroll_active = False
+
+        # Also used for column selection, not just above bug fix, so needed on all platforms
+        self._autoscroll_active = False
 
         # Need to wait until maintext has been registered to set the font preference
         preferences.set(PrefKey.TEXT_FONT_FAMILY, family)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: since these were probably only non-Mac bugs, I think all you can do is check that my changes haven't broken normal or column selection, inside & outside the window, and with split window or not.

tangledhelix:
> Testing notes: since these were probably only non-Mac bugs, I think all you can do is check that my changes haven't broken normal or column selection, inside & outside the window, and with split window or not.

I'll also test it on Linux when I get time.

srjfoo (approved):
FWIW, in `master` on macOS, if I start the selection, then press the `option` key, it does switch to column select, but doesn't change the cursor.

I always press `option` first, so hadn't noticed -- not all platforms switch to column select if the option key is pressed after the selection is started. In fact, BBEdit does not. However, in BBEdit, after you start a column select, you can let up on the option key and it continues in column-select mode until the end of the selection.

In fact, I looked at Sublime Text, and TextEdit, and both of them behave the same way.

Then I looked at VSCode ... and was kind of weirded out. Maybe it's my configuration, but what we consider a normal select (i.e. without `option`/`alt`), does a column select, and if I hold down `option`, I get what we call a normal selection. Is this normal behavior for VSCode?

At any rate -- the cursor change to the cross-hair is fixed on Mac, also.

windymilla:
> FWIW, in `master` on macOS, if I start the selection, then press the `option` key, it does switch to column select, but doesn't change the cursor.

That's as expected, because on Macs, normal selection uses Dan's code, but on Windows it doesn't so it has to notice the the `Alt` key has been pressed and begin to behave differently - one of the bugs was that it wasn't doing that.

> I always press `option` first, so hadn't noticed -- not all platforms switch to column select if the option key is pressed after the selection is started. In fact, BBEdit does not. However, in BBEdit, after you start a column select, you can let up on the option key and it continues in column-select mode until the end of the selection.
> 
> In fact, I looked at Sublime Text, and TextEdit, and both of them behave the same way.

Interesting

> Then I looked at VSCode ... and was kind of weirded out. Maybe it's my configuration, but what we consider a normal select (i.e. without `option`/`alt`), does a column select, and if I hold down `option`, I get what we call a normal selection. Is this normal behavior for VSCode?

No, not for me. No modifier keys gives me normal selection. I have to press Shift & Alt to get column selection - I can press those before or after starting to click & drag.

> At any rate -- the cursor change to the cross-hair is fixed on Mac, also.

Good :)

tangledhelix (approved):
I wasn't able to reproduce the exception on Linux even on the master branch. But in any case, it seems to work properly, no exceptions, on Mac and Linux for me.

PR #639, opened 2025-01-04
Improve column selection

windymilla:
This fixes the flickering between column and regular selection when column selecting outside the text window.

Although the flickering only happened on Windows, it was reported to be "choppy" on macOS. It's possible this might improve that, or at least not make things worse.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -810,6 +810,40 @@ class MainText(tk.Text):
         # Initialize highlighting tags
         self.after_idle(lambda: self.highlight_configure_tags(first_run=True))
 
+        # Set up proxy to intercept Tcl commands
+        name = str(self)
+        self._orig_self = name + "_orig"
+        # If debugging is being complicated by everything passing through self.__proxy
+        self.tk.call("rename", name, self._orig_self)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:

Potentially this could have broken normal or column selection using the mouse. It should not have affected keyboard selection using shift+arrows, nor toggling between normal/column selection.

At worst, this seems to fix the Windows issue (and doesn't make macOS worse); at best, it improves the choppiness on macOS.

windymilla:
Documenting failure here, so I don't do it again.

Using the proxy idea is really too much of a sledgehammer to crack a nut, with all the bits of flying nutshell you would expect. Firstly, there's the problem that you while using the proxy, all the Tcl command stream goes via the proxy from mainloop, so if your Python code has `except tk.TclError` it won't catch it, because the error isn't happening via a tkinter call, but through the proxy command you've set up.
Secondly, if you try to switch the use of proxy on and off, to limit it to the specific circumstance you want, you can easily hit problems where things happen while you've got your widget renamed, and so, for example, it can't find "maintext" as a 
(proxy) command or a widget.

It's possible I've made some mistake and the above can be avoided, but I've been unable to do that.

PR #641, opened 2025-01-04
Compress pptxt report of adjacent spaces/dashes

### LINKED ISSUE #640: Minor enhancement to PPtxt output: suppress repeated adjacent dash and character checks reporting on same line
Dashes, pluses, vertical lines, and equals are used and repeated a lot in tables, it would clean up the output a lot to limit the instances reported on the same line. The attached file has lines repeated 17 times for adjacent spaces, for example.

The same could apply for the adjacent spaces check.

Maybe a future enhancement. I use the tool frequently to check PG submissions, so I notice the amount I have to page through more than most.

BTW, I much prefer 2.0 PPtxt to Guiguts 1.x. I never did understand the purpose in reporting parentheses and curly quotes as special characters.

[railways.txt](https://github.com/user-attachments/files/18307548/railways.txt)

windymilla:
Instead of reporting every occurrence of multiple spaces or multiple dashes separately, report once per line, with a count of how many occurrences, e.g.
`23.4: (x3) a ---- a ---- a ---- a`

Fixes #640

### DIFF
--- a/src/guiguts/tools/pptxt.py
+++ b/src/guiguts/tools/pptxt.py
@@ -1259,7 +1259,7 @@ def adjacent_spaces_check() -> None:
         if re.search(r"\s\s+?", line) and not re.search(r"^\s+?", line):
             no_adjacent_spaces_found = False
             # Generate a new dialog line for each match on the line.
-            report_all_occurrences_on_line(r"\s\s+", line, line_number)
+            report_multiple_occurrences_on_line(r"\s\s+", line, line_number)
 
         line_number += 1
 
@@ -1322,6 +1322,33 @@ def double_dash_replace(matchobj: re.Match) -> str:
     """
 
     return matchobj.group(1) + "—" + matchobj.group(2)
+
+
+def report_multiple_occurrences_on_line(
+    pattern: str, line: str, line_number: int
+) -> None:
+    """Report multiple occurrences in one message for this line.
+
+    Args:
+        pattern: A regex of the pattern to be matched and highlighted.
+        line: The string on which to match the pattern.
+        line_number: The line number of the line in the file.
+    """
+
+    matches = list(re.finditer(pattern, line))
+    if not matches:
+        return
+    # Get start/end of first/last error on line.
+    error_start = matches[0].start()
+    error_end = matches[-1].end()
+    # Add record to the dialog with prefix if repeated matches.
+    checker_dialog.add_entry(
+        line,
+        IndexRange(f"{line_number}.{error_start}", f"{line_number}.{error_end}"),
+        error_start,
+        error_end,
+        error_prefix=f"(x{len(matches)}) " if len(matches) > 1 else "",
+    )
 
 
 def report_all_occurrences_on_line(pattern: str, line: str, line_number: int) -> int:
@@ -1847,7 +1874,7 @@ def dash_review() -> None:
         for record in a_hh:
             line_number = record[0]
             line = record[1]
-            report_all_occurrences_on_line(r"(\p{Pd}\p{Pd}+)", line, line_number)
+            report_multiple_occurrences_on_line(r"(\p{Pd}\p{Pd}+)", line, line_number)
 
     # Report hyphen-minus
 
--- a/src/guiguts/tools/pptxt.py
+++ b/src/guiguts/tools/pptxt.py
@@ -628,6 +628,8 @@ def weird_characters() -> None:
             # Under the header add to the dialog every line containing an instance of the weirdo.
             # If there are multiple instances of a weirdo on a line then the line will appear in
             # the dialog multiple times, each time highlighting a different instance of it.
+            # If character is in `report_once_list`, just report once per line (used for text tables)
+            report_once_list = "+|"
 
             prev_line_number = -1
             regx = "(" + "\\" + weirdo + ")"
@@ -647,7 +649,10 @@ def weird_characters() -> None:
                 # ... but note that a new dialog line is generated for each time the word appears
                 # on the line so there may be more than 5 dialog lines output.
                 line = book[line_number - 1]
-                report_all_occurrences_on_line(regx, line, line_number)
+                if weirdo in report_once_list:
+                    report_multiple_occurrences_on_line(regx, line, line_number)
+                else:
+                    report_all_occurrences_on_line(regx, line, line_number)
                 count += 1
                 prev_line_number = line_number
 


windymilla:
Testing notes:
[This file from the linked issue](https://github.com/user-attachments/files/18307548/railways.txt) has both spaces and dashes.
Run Tools->PPtext and scroll down results to see compressed output. In `master` each occurrence is reported on a separate line.

rtonsing (approved):
Excellent! I like the count shown in red.

There are still repetitions in "Character checks" for '+' and '|' if you want to handle those.

windymilla:
> There are still repetitions in "Character checks" for '+' and '|' if you want to handle those.

OK - those are now done. 

Prior to this, PPtxt was very much based on the online version, which was based on quite an old tool. IMHO, we should consider a few more changes, for example, several of the characters in "Character checks" are by no means suspicious these days, now that we're not restricted to ASCII. Fractions would be one example, but I'm sure there are others that often appear in that check - if too many characters are reported, the check becomes useless because the user can't be bothered to check all of them.

rtonsing:
Verified.

I agree about fractions and some other characters, super/subscripts for example. but I'm sure we will need multiple opinions. I'm not sure about the em-dash check either.

srjfoo:
Okay, I've got a question for y'all. The adjacent space check picks up on lines 360, 361 and 370. What about lines 361-366 and 371-375? They have a bunch of adjacent spaces, too. The only difference I see is that the lines that were identified have something in the first column, and the lines that weren't don't have anything in the first column. Is that because of the "(poetry, block-quotes, etc., are ignored)" comment.

I'm assuming that's correct behavior, but since I'm not used to it, it doesn't quite make sense to me, so I thought I'd ask -- it's certainly the same as what's in `master`.

The other thing that jumped out at me was the HTML tags getting highlighted for the `/` in the closing tag. Oh. Looking closer, there must be logic to abandon the check after a certain number of hits, after concluding that they're HTML tags? And since the `<i>` and `<sc>` elements are closer to the top of the file, the check is abandoned before it gets to the `<b>`s?

Anyway, since y'all didn't mention the table rows that have multiple spaces, but nothing in the first column, I'm assuming that you're okay with that. I'll go ahead and approve, because it certainly seems to work well otherwise, and matches the way the old version selects rows to show.

srjfoo (approved):

windymilla:
> Okay, I've got a question for y'all. The adjacent space check picks up on lines 360, 361 and 370. What about lines 361-366 and 371-375? They have a bunch of adjacent spaces, too. The only difference I see is that the lines that were identified have something in the first column, and the lines that weren't don't have anything in the first column. Is that because of the "(poetry, block-quotes, etc., are ignored)" comment.
> 
> I'm assuming that's correct behavior, but since I'm not used to it, it doesn't quite make sense to me, so I thought I'd ask -- it's certainly the same as what's in `master`.

Whether it's "correct" or not is a matter of opinion, I guess, but it is "intended", and AFAIK the same as other versions of pptext. 

> The other thing that jumped out at me was the HTML tags getting highlighted for the `/` in the closing tag. Oh. Looking closer, there must be logic to abandon the check after a certain number of hits, after concluding that they're HTML tags? And since the `<i>` and `<sc>` elements are closer to the top of the file, the check is abandoned before it gets to the `<b>`s?

Quite a few of the checks in pptext have limits on the number of similar warnings they output. In fact, I think pptext is/was supposed to be run on a text file, so there should be no `<i>`, `<sc>`, etc., by that stage. It would also mean that the whole table would be indented (rather than left-aligned but enclosed in `/X` markup) so it wouldn't report multiple spaces, etc.

At some point, I think we should consider how much we want to allow pptext to move away from the historical behavior of the old GG1 version and the online version. 

There are several changes that could potentially make it more useful:

- the way it lists the first 5 errors of one type, and then just says "there are more of these"
- several more characters could be removed from the ones it flags as unusual
- floating double quotes with 2 spaces either side should be ignored (since they are probably ditto marks)
- remove good em-dashes from the em-dash report
- mixed case words which are actually Capitalized-Hyphenated shouldn't be reported
- don't report ` .123` as a spaced period since it has digits after it
- add more exceptions to double punctuation report, e.g. `pp.,` and maybe single capital letter before period, e.g. `N. Y.,`

rtonsing:
> Okay, I've got a question for y'all. The adjacent space check picks up on lines 360, 361 and 370. What about lines 361-366 and 371-375? They have a bunch of adjacent spaces, too. The only difference I see is that the lines that were identified have something in the first column, and the lines that weren't don't have anything in the first column. Is that because of the "(poetry, block-quotes, etc., are ignored)" comment.

I consider that a problem, although different from this original issue. I don't see any reason not to also report adjacent spaces on lines with leading spaces. Guiguts 1.x also does not report them, but online pptext _does_:

`362:              57 |  52  |        | 470 | 110  |  33  |    18    |    99    |  2972  |   704   |  340  |  2⅝ | 6⅜ | 3¼ | 8¾ |   14    | 12
`

rtonsing:
-     the way it lists the first 5 errors of one type, and then just says "there are more of these"
But online pptext has a 'verbose' option to list all of them
-     several more characters could be removed from the ones it flags as unusual
Agree
-     floating double quotes with 2 spaces either side should be ignored (since they are probably ditto marks)
Agree
-     remove good em-dashes from the em-dash report
Agree
-     mixed case words which are actually Capitalized-Hyphenated shouldn't be reported
Agree
-     don't report  .123 as a spaced period since it has digits after it
Agree
-     add more exceptions to double punctuation report, e.g. pp., and maybe single capital letter before period, e.g. N. Y.,
Agree

I just ignore the issues reported for HTML and block markup, but I **do** want to fix other issues before generating the HTML. An option to ignore standard PGDP markup would be nice, I agree they should be reported for the final text version. But of course, providing options for PPtxt would be a whole new thing.

windymilla:
> I consider that a problem, although different from this original issue. I don't see any reason not to also report adjacent spaces on lines with leading spaces. Guiguts 1.x also does not report them, but online pptext _does_:
> 
> `362: 57 | 52 | | 470 | 110 | 33 | 18 | 99 | 2972 | 704 | 340 | 2⅝ | 6⅜ | 3¼ | 8¾ | 14 | 12 `

Thanks for this and your other comments @rtonsing - I'll link to this discussion from a new issue I'm going to put in as a future feature to review pptext's error output.

rtonsing:
Thanks, I was wondering if I should add a new issue.

It is about time I looked into making updates to online pptext, myself.

windymilla:
> It is about time I looked into making updates to online pptext, myself.

In an ideal world, they would both behave the same, but I don't think that's practical given their very different environments

PR #645, opened 2025-01-05
Added recent modules to sphinx index

windymilla:
Also a couple of minor doc "errors" fixed.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -2,8 +2,9 @@
 Guiguts Documentation
 #####################
 
+*******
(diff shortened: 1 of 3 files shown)


windymilla:
Not quite sure what you can do to test this, and even if it's not right, no-one will ever notice, so perhaps I should just merge it.

srjfoo (approved):
I've approved it, but I agree. I can't think of anything I know how to do to test it. It _*seems*_ reasonable, so I think just go ahead and merge?

PR #650, opened 2025-01-10
Playing around with highlight colors.

srjfoo:
The main one I really want to see changed is the spotlight color, from `orange` to `darkorange`, at least for the dark themes -- white text just doesn't show as well on `orange` as it does on the `darkorange`.

The rest is playing around only for the colors for the dark themes to try to get better contrast between text color and background color. Some of what I did was to change the double-quotes to use related but different colors, and the straight quotes, ditto, and make them (for dark theme -- I didn't touch the light theme) a bit more noticeable.

It occurrred to me later that it might be better to have straight quotes' colors be related, and curlies' colors be related, but that can be a talking point.

My feelings will absolutely not be hurt if y'all throw up your hands in horror. 😁 (on everything but the spotlight color change :D ) I'm starting this out as a draft. No big hurry.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -292,17 +292,17 @@ class HighlightColors:
     # Must be a definition for each available theme
     QUOTEMARK = {
         "Light": {"bg": "#a08dfc", "fg": "black"},
-        "Dark": {"bg": "#a08dfc", "fg": "white"},
+        "Dark": {"bg": "darkmagenta", "fg": "white"},
     }
 
     SPOTLIGHT = {
         "Light": {"bg": "orange", "fg": "black"},
-        "Dark": {"bg": "orange", "fg": "white"},
+        "Dark": {"bg": "darkorange", "fg": "white"},
     }
 
     PAREN = {
         "Light": {"bg": "violet", "fg": "white"},
-        "Dark": {"bg": "violet", "fg": "white"},
+        "Dark": {"bg": "mediumpurple", "fg": "white"},
     }
 
     CURLY_BRACKET = {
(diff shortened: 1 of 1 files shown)


srjfoo:
And obviously, I did this wrong, because I forgot to set it to draft -- and also don't have everything in place to properly get things tested, so -- advice, please!

windymilla:
If you want to set it to Draft, there should be a link just under the Reviewers list on the right. If you don't have that, then maybe I need to alter your access:
image

windymilla:
> It occurrred to me later that it might be better to have straight quotes' colors be related, and curlies' colors be related, but that can be a talking point.

Sometimes, I guess a user might want to be alerted to the fact that some quotes were straight instead of curly (though curly checking tool reports this)< and sometimes they might want to be alerted to which are single and which are double. Personally, I think it's better to have double quotes related, and single quotes related, which is what we currently have I think.

The changes all look fine to me, if you're happy with the colors you've chosen.

srjfoo:
> > It occurrred to me later that it might be better to have straight quotes' colors be related, and curlies' colors be related, but that can be a talking point.
> 
> Sometimes, I guess a user might want to be alerted to the fact that some quotes were straight instead of curly (though curly checking tool reports this)< and sometimes they might want to be alerted to which are single and which are double. Personally, I think it's better to have double quotes related, and single quotes related, which is what we currently have I think.
> 
> The changes all look fine to me, if you're happy with the colors you've chosen.

I'm happy with the colors I've chosen -- at least for now 😁 

I do see the "change to draft" link now -- I didn't notice it last night, but I may not have been looking in the right place. I can't assign reviewers, and I don't know if that makes the difference, or if I was just blind!

tangledhelix (approved):
The colors look fine to me...

windymilla (approved):

PR #652, opened 2025-01-10
Improve behavior of "R&S"

### LINKED ISSUE #649: Improve Search & Replace "R & S" action
1. Enter a search text that appears multiple times
2. Enter a replacement text
3. Click Search
4. Click Replace
5. Click R & S
6. "No text found to replace" is shown, even though there are more matches

It would be better to search to the next instance.

windymilla:
If unable to do a replace and search, most likely because the user already did a replace, then just do the "search".

Fixes #649

Testing notes: see issue

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -548,8 +548,12 @@ class SearchDialog(ToplevelDialog):
             start_index = maintext().index(MARK_FOUND_START)
             end_index = maintext().index(MARK_FOUND_END)
         except tk.TclError:
-            sound_bell()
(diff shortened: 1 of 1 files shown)


rtonsing (approved):
It works!

srjfoo (approved):
I totally misunderstood the solution proposed in the issue -- I can live with this. I still think that it's not unreasonable to warn the user that they can't do a replace without a search first, but as long as the replace is ignored, and just the S of R&S is done, there's no likelihood of a screwy replace happening.

PR #653, opened 2025-01-11
Add default extension during File-->Save As

### LINKED ISSUE #651: Add a default filename extension for saving
By default, newly created and saved files in GG2 lack a filename extension. To enhance user experience, I propose a default extension of '.txt' for new files. While '.html' is also a valid option, '.txt' is likely more suitable for the majority of new files. 

This suggestion could be easily overridden by the user, providing flexibility. 

Given that this aligns with GG1's behavior, I recommend considering this as a feature enhancement for future versions of GG2.

windymilla:
First detect if the file is an HTML file, i.e. has the first line `<!DOCTYPE html>`
Also check if the filename currently has an extension. If no extension, or extension is `.txt`, then suggest `.html` for HTML files.
If no extension suggest `.txt` for non-HTML files. Use the suggested extension when pre-filling the
Save As dialog. User can override and set their own extension if they want to.

Fixes #651

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -294,10 +294,29 @@ class File:
         Returns:
             Chosen filename or None if save is cancelled
         """
+        # If no current extension, or ".txt" extension, set extension to ".html"
+        # if it's an HTML file. Similarly for text files, otherwise leave alone.
+        cur_ext = os.path.splitext(self.filename)[1]
+        html_file = maintext().get("1.0", "1.end") == "<!DOCTYPE html>"
+        if html_file and cur_ext in ("", ".txt", ".html"):
+            extension = ".html"
+        elif not html_file and cur_ext in ("", ".txt"):
+            extension = ".txt"
+        else:
+            extension = cur_ext
+        match extension:
+            case ".html":
+                file_type = "HTML files"
+            case ".txt":
+                file_type = "Text files"
+            case _:
+                file_type = f"{extension} files"
+        suggested_fn = os.path.splitext(self.filename)[0] + extension
+
         if fn := filedialog.asksaveasfilename(
-            initialfile=os.path.basename(self.filename),
-            initialdir=os.path.dirname(self.filename),
-            filetypes=[("All files", "*")],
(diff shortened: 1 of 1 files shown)


srjfoo:
image


This startled me a bit, but I had to go verify in master that nothing is auto-filled by the OS, and let my brain think a bit. Then I realized that what some other apps usually do is open an empty file with the default title of "untitled.ext" where ext is whatever the default extension is for that application (.txt, .psd). Others use some other kind of default (FF uses "blanktab.html", for instance).

May I suggest something like that? When opening an empty document in GG, set the title bar to "untitled" or something else of your choice, and use that as the default filename. If a user accidentally just presses return without typing over the default name you've given it, they'll create a dot file, which will be hidden on macOS and Linux.

windymilla:
> May I suggest something like that? When opening an empty document in GG, set the title bar to "untitled" or something else of your choice, and use that as the default filename. If a user accidentally just presses return without typing over the default name you've given it, they'll create a dot file, which will be hidden on macOS and Linux.

I didn't do exactly this (for technical reasons relating to how to detect if the file has or hasn't ever been saved, and which directory "untitled.txt" should be saved in if you just click Save rather than Save As). I have made it pre-fill the Save As dialog with "untitled.txt" (or "untitled.html") rather than just ".txt" or ".html"

srjfoo (approved):
> > May I suggest something like that? When opening an empty document in GG, set the title bar to "untitled" or something else of your choice, and use that as the default filename. If a user accidentally just presses return without typing over the default name you've given it, they'll create a dot file, which will be hidden on macOS and Linux.
> 
> I didn't do exactly this (for technical reasons relating to how to detect if the file has or hasn't ever been saved, and which directory "untitled.txt" should be saved in if you just click Save rather than Save As). I have made it pre-fill the Save As dialog with "untitled.txt" (or "untitled.html") rather than just ".txt" or ".html"

That's fine -- my main concern was the `.txt` filenme that came up initially. As far as I'm concerned `untitled` is perfect for our purposes.

PR #661, opened 2025-01-12
Fix Find All highlight priority

### LINKED ISSUE #660: Find All within selection has a highlight priority issue
You can't see the orange spotlight showing the found match because the selection has higher priority. 

Suggested solution is to clear selection when jumping to a spotlighted match.
Adding
`        maintext().clear_selection()`
at about line 933 in checkers.py (just after it positions the insert cursor) should do it.

windymilla:
When Find All was done inside a selection, the spotlight match was hidden by the selection.
Fix by clearing selection when a spotlight is displayed, usually by clicking a message line in a checker dialog, also when the dialog is created.

Fixes #660

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -920,6 +920,7 @@ class CheckerDialog(ToplevelDialog):
             maintext().set_insert_index(
                 IndexRowCol(start), focus=(focus and not is_mac())
             )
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Since we have the ability to restore the previous selection, the PPer can do that if they really need to search within the same selection.

PR #664, opened 2025-01-13
Implement bookloupe

### LINKED ISSUE #463: Bookloupe
Bookloupe (and gutcheck before it) has been a mainstay of PP/PPV checking for a looooooong time.
1. PPers may find it reassuring to continue to have bookloupe available as an explicit tool, rather than combining it with pptxt.
2. At least in my experience, bookloupe was often run near the start of the PPing process, and pptxt near the end, so separate tools would make sense from that point of view too.
3. Consider if there are some bookloupe checks that are out-of-date (like "non-ascii character")

Previously this PR said

> Since bookloupe and pptxt have very similar aims, it makes sense to just have the one tool, rather than duplicate features. However, they may be some important checks in bookloupe that are not in pptxt, and could be transferred.
> 
> At the very least, it would be good to document here or somewhere else what features of bookloupe are covered by pptxt.

windymilla:
This tool is not identical in behavior to the standalone tool used in GG1, but it does do essentially the same tests, and reports them in a similar way.

There will be reports from this that the GG1 tool does not report, and vice versa.

The aim is to provide a useful tool for PPers.

Future work will probably include more on curly quotes, and other features of "recent" DP books.

Fixes #463

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -102,6 +102,11 @@ Word Frequency
 Tools
 *****
 
+Bookloupe
+=========
+    
+.. automodule:: tools.bookloupe
+
 Jeebies
 =======
     
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -60,9 +60,10 @@ from guiguts.preferences import preferences, PrefKey
 from guiguts.root import root
 from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
-from guiguts.tools.pptxt import pptxt
+from guiguts.tools.bookloupe import bookloupe_check
 from guiguts.tools.jeebies import jeebies_check, JeebiesParanoiaLevel
 from guiguts.tools.levenshtein import levenshtein_check, LevenshteinEditDistance
+from guiguts.tools.pptxt import pptxt
 from guiguts.utilities import is_mac, folder_dir_str
 from guiguts.widgets import themed_style, theme_name_internal_from_user
 from guiguts.word_frequency import word_frequency, WFDisplayType, WFSortType
@@ -696,6 +697,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         menu_tools = Menu(menubar(), "~Tools")
         menu_tools.add_button("Basic Fi~xup...", basic_fixup_check)
         menu_tools.add_button("~Word Frequency...", word_frequency)
+        menu_tools.add_button("~Bookloupe...", bookloupe_check)
         menu_tools.add_button(
             "~Spelling...",
             lambda: spell_check(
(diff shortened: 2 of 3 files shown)


windymilla:
NTS: Look at "unspaced quote" problem noted by @srjfoo

srjfoo:
`Query standalone 1` appears to be broken in GG1, and is not quite completely corrected in GG2. If you have a mixed number, like 1-1/2, in GG2, the whole number part is identified as a standalone 1 (I would argue that since it's part of a mixed number, it's not standalone).

GG1, unlike GG2, spotlights the whole line where the fraction is (not a complaint, just an observation). Except the 1. It does this for both mixed numbers and normal fractions with an individual 1 in the numerator (again, since it's part of a fraction, is it really a standalone 1?).

image

image

image

Or is this a case of being ultra paranoid in the checks?

srjfoo:
`No punctuation at para end?` is off somehow. In one case (GG2), it's there, but has a closing italics tag following it. The other two are also identified by GG1, and I think I see what's happening.  If there are two sequential lines without a line break, it's treated as a paragraph. GG2 does a good job of identifying headings and omits them from the list, but GG1 does not.

`Endquote missing punctuation?` -- I just happened to notice that the column given in GG2, is that of GG1+1 -- is that a result of decisions we've made specific to GG2?

`Asterisks?` -- GG1 seems to lump multiple consecutive asterisks in one result. GG2 does each individually, so it finds a properly formatted proofer's note twice.

This comment and the previous one are mostly highlighting things that jumped out at me -- I was comparing GG2 results on one project with GG1 on the same file. This does not constitute exhaustive testing, because, while I did some initial testing switching between split pane and non-split, and switching between panes, I did not do that for all tests. I looked at the results I got for GG2 and hid all but the corresponding GG1 results (since I couldn't sort the GG1 list alphabetically).

The main one I don't like is the fractions, but if there's a reason for doing it that way, and it's expected behavior, so be it.

Let me know if there's anything specific that you want me to focus on, but I'm satisfied that if you don't want to fix the fractions, it's time to toss it to the wolves. 😁

windymilla:
> Let me know if there's anything specific that you want me to focus on, but I'm satisfied that if you don't want to fix the fractions, it's time to toss it to the wolves. 😁

Thanks - I'll have a look at the things you've just mentioned, then you can have a final look at those before we release it.

windymilla:
Hopefully all fixed now:
1. Inline markup was breaking end-para punc check
2. Multiple asterisks were generating multiple messages
3. Fractions like `1-3/4` generated "standalone 1"

windymilla:
NTS:

1. Tooltip should mention (Shift-)right-click
2. Slashes within fractions shouldn't be reported.

srjfoo (approved):

PR #666, opened 2025-01-14
Insert HTML header above first page mark

### LINKED ISSUE #657: HTML generation adds first page number before DOCTYPE declaration instead of at beginning of body
Using HTML generation in 2.0.0a13, the HTML for first page number is positioned at the very start of the file (because that's where the marker is) rather than at the very start of the \<body\> section.  My generated file starts:

`<p><span class="pagenum" id="Page_673">[Pg 673]</span></p><!DOCTYPE html>`

I'm PP'ing a periodical, which is why the page number is high and the first page in the project has a number and isn't, for example, an unnumbered title page. GG1 adds this to the start of the \<body\>. I haven't tested using a project with an unnumbered first page.

windymilla:
Since first page mark may be at the very start of the master file prior to HTML auto-generation, and page markers have LEFT gravity, anything inserted at the start of the file would go after the first page mark. Then when the pagenum spans were added, the first pagenum span would be above the HTML header.

Fix by temporarily switching the gravity to RIGHT while the header is inserted.

Fixes #657

Also, make it work for multiple page marks at the top.

### DIFF
--- a/src/guiguts/html_convert.py
+++ b/src/guiguts/html_convert.py
@@ -841,6 +841,22 @@ def insert_header_footer() -> None:
     User's header can either be a complete header, or just some CSS
     to be inserted at the end of the default CSS.
     """
+    # Temporarily set the gravity on page markers at the start
+    # of the file to RIGHT to ensure the header gets added above them.
+    # Need to create list of marks first before changing any gravities,
+    # because if pages 1 & 2 are both at the start, then changing the
+    # gravity of page 1 to RIGHT means that page 2 is no longer the
+    # "next" mark after page 1 (because page 1 is to the right of page 2).
+    # Then page 2 would be skipped and remain with LEFT gravity.
+    page_mark_list = []
+    page_mark = "1.0"
+    while page_mark := maintext().page_mark_next(page_mark):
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
By deleting some lines at the top of any CTF file, you can ensure that one or more page marks are right at the top of the file. Then do HTML Auto-generate. In `master`, spans for those pages will appear at the top of the HTML, above the HTML header. In this branch, they shouldn't.

Please test single or multiple page marks at the top of the file - multiple could happen if early pages are blank/illos/whatever, or just user being careless with page break positioning. It doesn't usually matter much because early pages are normally "No count" pages and so aren't shown in HTML anyway, but linked issue was a periodical with a numbered first page.

srjfoo (approved):
LGTM -- the main thing I noticed missing was the `working ...` message. At least one of the files I tried took a very long time (I think the one that threw the unexpected error :D).

windymilla:
> LGTM -- the main thing I noticed missing was the `working ...` message. At least one of the files I tried took a very long time (I think the one that threw the unexpected error :D).

I'll look at that separately - pretty sure it's nothing to do with this PR

PR #668, opened 2025-01-15
Add checker dialog undo update mechanism

### LINKED ISSUE #655: Undo/Redo should alert some areas when they happen
Past windy wrote in GG1:

```perl
# Override TextUndo::undo and redo so we can flag it to any routines that need to know
sub undo {
    my ($self) = @_;
    $self->SUPER::undo();
    ::errorcheckillosnupdateneeded();
}

sub redo {
    my ($self) = @_;
    $self->SUPER::redo();
    ::errorcheckillosnupdateneeded();
}
```

Future windy could do something similar.

windymilla:
If a checker dialog makes a change, e.g. moving an illustration down to a paragraph break, it may re-run the tool. However, if the user does Undo or Redo, there was no way to flag that the tool needed re-running again.

This commit adds such a mechanism. The mechanism is based within maintext, since that is where Undo/Redo occur, but convenience functions have been added to CheckerDialog because it is probably checkers that will want to take advantage of it.

The checker dialog, e.g. Illo & sidenote fixup, registers a function to be called on undo/redo. The base checker class handles registering that with maintext and removing it when the dialog is deleted.

Implemented for Illo/SN fixups, as an example of what could be done for any checker tool - it isn't needed for all checkers.

Note that if you have both the Illo and the SN dialog popped at the same time, when you do an undo, it will undo the most recent change and will update both the illo and SN checker dialogs, which might mean that the one you were not using becomes the focus one. The moral of that tale is, probably best not to work simultaneously on illos and SNs and then use Undo/Redo - unlikely that a PPer would do that anyway. Similarly, if you have the Illo dialog visible, then go and make some other manual edit to the text, then undo it, there's (currently) no way for the code to know that the change wasn't via the illo dialog, so it will update the illo dialog and select the illo message that was current at the point of undoing.

It's possible that further work could be done to make this cleverer, but it would be getting rather complex and prone to error IMO, so not doing it for now.

Another possible solution to the whole issue would be to disable the illo (and/or SN) dialog list if the user does an undo/redo until such time as they re-run the tool manually. This would possibly be safer, but not as convenient. Your thoughts are welcome. (See #669 for this type of solution)

Fixes #655

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -407,6 +407,33 @@ class CheckerDialog(ToplevelDialog):
                     maintext().mark_unset(mark)
             maintext().remove_spotlights()
 
+    def add_undo_redo_callback(self, func: Callable[[], None]) -> None:
+        """Add callback function to a list of functions to be called when
+        <<Undo>> or <<Redo>> virtual events happen.
+
+        Callback will be remove by `tidy_up` when checker dialog destroyed.
+
+        Args:
+            key: Key to identify the caller, e.g. dialog class name
+            func: Callback function to be added to list.
+        """
+        maintext().add_undo_redo_callback(self.__class__.__name__, func)
+
+    def tidy_up(self, event: tk.Event) -> None:
+        """Tidy up when the dialog is destroyed.
+
+        Calls the reset method which may be overridden.
+
+        Args:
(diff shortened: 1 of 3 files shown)


windymilla:
Merged #669 instead

PR #669, opened 2025-01-15
Alternative to #668 (less-smart undo)

windymilla:
See [#668](https://github.com/DistributedProofreaders/guiguts-py/issues/668) for main attempt to solve this issue.

In this commit, if you do an undo/redo, the list of
messages in the illo/SN checkers is removed until you
manually click Re-run.

This is just to see what it's like, and what you think is
best. It would probably be possible (though more
complicated) to gray out the text instead of removing it
and intercept user attempts to click on the messages and
either do nothing or put up a message saying "you need to
re-run the tool".

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -22,6 +22,7 @@ from guiguts.widgets import ToplevelDialog, TlDlg, mouse_bind, Busy
 MARK_ENTRY_TO_SELECT = "MarkEntryToSelect"
 HILITE_TAG_NAME = "chk_hilite"
 ERROR_PREFIX_TAG_NAME = "chk_error_prefix"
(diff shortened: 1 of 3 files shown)


windymilla:
Slightly less "less smart" version, i.e. slightly smarter.

When you undo, it now replaces the messages in the dialog with a single message telling you to click it to refresh. This behaves subtly differently to "Re-run", in that it re-selects the selected message. I first did this because of difficulties with using Re-run, but then decided I prefer it, to make it clear to the user they are doing something a bit different. Re-run is a complete reset, with no messages selected once it has re-run. 
If it's confusing, my current method could be applied to the Re-run button instead (or as well as clicking the message).

srjfoo (approved):

PR #686, opened 2025-01-20
Fix some  HTML autogen bugs

### LINKED ISSUE #684: Ugly exception running HTML Generator
Probably a problem in the file, but the error is unhelpful. It would be better to catch the exception and report the cause.

[fossilplants3.zip](https://github.com/user-attachments/files/18479117/fossilplants3.zip)

Image

Windows 11
current code

### LINKED ISSUE #685: Improve determination of HTML title in HTML Auto-generate
A file beginning like this
```
/C
                       CAMBRIDGE BIOLOGICAL SERIES

                              FOSSIL PLANTS
C/
```
causes the HTML Auto-generator dialog to suggest a title
` cambridge biological series`
i.e. all lowercase, and with a space at the start

NTS: Probably removing the space will fix the capitalization error.

windymilla:
1. If the final chapter heading is incorrectly formatted,
having 4 blank lines before, but not 2 after, the code
never found the end of the chapter heading, and so
later went wrong looking for `</h2>`, leading to
an exception. Fixed by trapping at end of file if code
thought it was still in a chapter heading. Fixes #684
2. When an error is trapped by raising an exception,
the code then continued to run, when it should
have stopped.
3. The code stopped 1 line from the end, so a final
footnote was reported as not having a blank line
following it, although it did.
4. Trim trailing/leading spaces from suspected title, which
also leads to capitalization working correctly. Fixes #685


windymilla:
Testing notes:
[File from linked issue](https://github.com/user-attachments/files/18479117/fossilplants3.zip) is a suitable test file - just load it and use HTML-->Autogenerate

As it stands, the file has only one blank line after the "FOOTNOTES" chapter heading. Although this is incorrect, it shouldn't cause an exception, as it does in `master`.
Still in `master`, even if you fix the above by adding a second blank line, you still get an error about a footnote not having a paragraph end. This was because the code wasn't checking the final line of the file.
Still in `master`, even if you fix that by adding a couple of blank lines at the end of the file, you'll notice that the suggested title in the Autogenerate dialog is all lowercase and has a space a the beginning.

In this branch, hopefully all of the above are fixed.

rtonsing (approved):
Much better, reports the chapter head missing 2 lines issue. When fixed, it completes OK.

The title is improved, too.

Note that after the error is reported, the file has been changed, so undo or restart without saving must be done. Some may consider this an issue.

windymilla:
> Note that after the error is reported, the file has been changed, so undo or restart without saving must be done. Some may consider this an issue.

It's not practical to discover all the errors before making a change, so it is inevitable that changes will be made. I can think of two possible solutions, one of which is automatically undoing the changes that have been made, and the other is to automatically re-load the backup file that is saved immediately before autogeneration. I've entered [an issue discussing the possibilities](https://github.com/DistributedProofreaders/guiguts-py/issues/687), and would be grateful if you have any comments @rtonsing

srjfoo:
> Testing notes: [File from linked issue](https://github.com/user-attachments/files/18479117/fossilplants3.zip) is a suitable test file - just load it and use HTML-->Autogenerate
> 
> As it stands, the file has only one blank line after the "FOOTNOTES" chapter heading. Although this is incorrect, it shouldn't cause an exception, as it does in `master`. Still in `master`, even if you fix the above by adding a second blank line, you still get an error about a footnote not having a paragraph end. This was because the code wasn't checking the final line of the file. Still in `master`, even if you fix that by adding a couple of blank lines at the end of the file, you'll notice that the suggested title in the Autogenerate dialog is all lowercase and has a space a the beginning.
> 
> In this branch, hopefully all of the above are fixed.

I didn't do all of your "still in master"s, but all that makes sense.

I didn't get a footnote error; "CAMBRIDGE BIOLOGICAL SERIES" was not converted to all lower case.

I did, however, get an alert box: "Chapter heading does not have two blank lines before end of file". Is that the expected response for point 1 above? I found it confusing.

And I don't see a `<title>` element or `<html>`, `<head>` or `<body>` elements or any CSS. What am I missing? I checked `master`, and it does have it, and checked this branch again, and it still doesn't.

srjfoo (approved):
Re: the following -- nm. I finally figured out when I was reviewing the next one that the other stuff was missing because HTML generation hadn't completed.

> And I don't see a `<title>` element or `<html>`, `<head>` or `<body>` elements or any CSS. What am I missing? I checked `master`, and it does have it, and checked this branch again, and it still doesn't.

PR #688, opened 2025-01-20
Improve behavior if HTML autogen errors

### LINKED ISSUE #687: Allow HTML autogeneration to fail more helpfully.
Currently if there is a reported error during autogeneration, the user will be left with a part-converted file. 
1. They might continue working with the file and not notice until much later that it wasn't completely converted. 
2. Or they will need to use Undo to revert back to the pre-autogeneration state (unsure if this is wise - will pagenum positions be preserved OK). 
3. Or they will need to load either their last-saved master file
4. Or they will need to load the `myfile-htmlbak.txt` file before fixing the error and retrying the autogeneration.

Scenario 1 is clearly undesirable, though perhaps unlikely, given that the user will have seen the error, but they may not interpret the error as meaning  "your file has not been converted fully".
Scenario 2 is potentially risky - I think at the moment, none of the changes that happen during HTML conversion will move page break marks, but I'm not sure of that, and definitely not sure that future changes will not do that.
Scenario 3 isn't ideal, because they may have made changes to the master file since their previous save, but before pressing Auto-generate. Those changes would be lost..
Scenario 4 is also not ideal, because when they load the htmlbak file, they would need to then do a Save As to save it again over the top of their master file - either that or they will end up with their file being called `myfile-htmlbak.txt` and a new backup file `myfile-htmlbak-htmlbak.txt`. This would be compounded if they needed to re-run autogeneration more than once.

Suggestions:
1. The code automatically re-loads the `myfile-htmlbak.txt` file, but names it `myfile.txt` within the program, thus avoiding the problems with scenarios 3 and 4. So, it is the same as re-loading `myfile.txt` but with the contents (& page breaks) of the pre-conversion backup.
2. Or, do the same as 1, but after giving the user the choice. For example, it could pop a dialog saying something like

> Since an error has occurred during HTML autogeneration, the file has not been completely converted, and will most likely be invalid. The pre-conversion backup file can be re-loaded automatically for you by clicking OK, and you can make the necessary changes before attempting autogeneration again. Alternatively, by clicking Cancel, you can look at the partly converted file in case that is helpful in determining what went wrong.

windymilla:
1. Output line number of snippet of text in error message to aid user in finding the problem
2. Raise exceptions throughout to give consistent error behavior
3. If an error occurs ask user if they want to load the pre-conversion backup file. If they do, load it but keep the current filename.

Fixes #687 

NTS: Merge after #686

### DIFF
--- a/src/guiguts/html_convert.py
+++ b/src/guiguts/html_convert.py
@@ -140,14 +140,18 @@ def html_autogenerate() -> None:
             icon=messagebox.QUESTION,
         )
         if reload:
-            old_filename = (
-                the_file().filename
-            )  # Save name because load_file will overwrite it
+            # Save name because load_file will overwrite it
+            old_filename = the_file().filename
             the_file().load_file(backup_fn)
-            the_file().filename = old_filename  # Restore original filename
-            the_file().store_recent_file(
-                old_filename
-            )  # Put it at the top of the recent files list
+            # Restore original filename
+            the_file().filename = old_filename
+            # Put it at the top of the recent files list
+            the_file().store_recent_file(old_filename)
+        # Go to error line number if available
+        line_num, is_line_num = re.subn(r"Line (\d+):.*", r"\1", str(exc))
+        if is_line_num:
+            maintext().set_insert_index(IndexRowCol(int(line_num), 0))
+
     Busy.unbusy()
 
 
--- a/src/guiguts/html_convert.py
+++ b/src/guiguts/html_convert.py
@@ -5,7 +5,7 @@ import importlib.resources
 import logging
 from pathlib import Path
 import tkinter as tk
-from tkinter import ttk
+from tkinter import ttk, messagebox
 from typing import Optional
 
 import regex as re
@@ -126,21 +126,40 @@ class HTMLGeneratorDialog(ToplevelDialog):
 def html_autogenerate() -> None:
     """Autogenerate HTML from text file."""
     Busy.busy()
-    fn = re.sub(r"\.[^\.]*$", "-htmlbak.txt", the_file().filename)
-    the_file().save_copy(fn)
-    css_indents.clear()
+    backup_fn = re.sub(r"\.[^\.]*$", "-htmlbak.txt", the_file().filename)
+    the_file().save_copy(backup_fn)
+    try:
+        do_html_autogenerate()
+    except SyntaxError as exc:
+        logger.error(exc)
+        # Check if user would like to reload backup file
+        reload = messagebox.askokcancel(
+            title="Re-load backup file?",
+            message="An error occurred during HTML autogeneration,\nand the file was not completely converted.",
+            detail='Click "OK" to re-load pre-conversion backup file,\nor "Cancel" to explore partially converted file.',
+            icon=messagebox.QUESTION,
+        )
+        if reload:
+            old_filename = (
+                the_file().filename
+            )  # Save name because load_file will overwrite it
+            the_file().load_file(backup_fn)
+            the_file().filename = old_filename  # Restore original filename
+            the_file().store_recent_file(
+                old_filename
+            )  # Put it at the top of the recent files list
+    Busy.unbusy()
 
+
+def do_html_autogenerate() -> None:
+    """Do the work of HTML autogenerate."""
+    css_indents.clear()
     maintext().undo_block_begin()
     remove_trailing_spaces()
     adjust_pagemark_postions()
     html_convert_entities()
     html_convert_title()
-    try:
-        html_convert_body()
-    except SyntaxError as exc:
-        logger.error(exc)
-        Busy.unbusy()
-        return
+    html_convert_body()
     html_convert_inline()
     html_convert_smallcaps()
     html_convert_footnotes()
@@ -150,7 +169,6 @@ def html_autogenerate() -> None:
     html_add_chapter_divs()
     html_wrap_long_lines()
     maintext().set_insert_index(maintext().start())
-    Busy.unbusy()
 
 
 def remove_trailing_spaces() -> None:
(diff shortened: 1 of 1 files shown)


windymilla:
Described [here in manual](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/HTML_Menu#HTML_Generator)

rtonsing (approved):
Tried it both ways, it works!

So why doesn't it automatically go to the line where the error occurred? :-)
Seriously, that would be nice too, but I understand if it isn't easy.

windymilla:
> Tried it both ways, it works!
> 
> So why doesn't it automatically go to the line where the error occurred? :-) Seriously, that would be nice too, but I understand if it isn't easy.

OK - it does go to the line number now. Should do it whether you do OK or Cancel.

rtonsing:
👍

srjfoo (approved):
I used fossilplants3.txt as my test file. I can't say that the line position of the cursor was especially helpful for the error message in this case, but I'm guessing it's there because that's the point at which it was determined that it _**was**_ an error. Definitely nicer to have a more specific error message.

windymilla:
> I used fossilplants3.txt as my test file. I can't say that the line position of the cursor was especially helpful for the error message in this case, but I'm guessing it's there because that's the point at which it was determined that it _**was**_ an error. Definitely nicer to have a more specific error message.

Could you be more specific about the cursor position problem, please @srjfoo. When I run it, it places the cursor on line 25447, which is the last blank line before the chapter heading that it's complaining about.

srjfoo:
> > I used fossilplants3.txt as my test file. I can't say that the line position of the cursor was especially helpful for the error message in this case, but I'm guessing it's there because that's the point at which it was determined that it _**was**_ an error. Definitely nicer to have a more specific error message.
> 
> Could you be more specific about the cursor position problem, please @srjfoo. When I run it, it places the cursor on line 25447, which is the last blank line before the chapter heading that it's complaining about.

Ohhhh, my bad! I think that my problem was that I wasn't thinking of the "Footnotes" heading as a chapter heading. I can totally understand the PPer missing that, to be honest. Perhaps, when placing footnotes at the end of the file, `FOOTNOlTES` heding should have the extra blank line added in between the heading and the first of the footnotes? (In another PR 😁)

windymilla:
> Perhaps, when placing footnotes at the end of the file, `FOOTNOlTES` heding should have the extra blank line added in between the heading and the first of the footnotes? (In another PR 😁)

When GG2 adds a LZ, it inserts two blank lines before the FOOTNOTES heading, which makes it a section heading. If the user adds more blank lines, or if there are extra blank lines at the LZ point, then it could end up with 4 blank lines, in which case the user would need to add another blank line after the heading. The pptext tool (I think) does a checklist of blank line spacings, which should alert the user to the problem.

srjfoo:
> > Perhaps, when placing footnotes at the end of the file, `FOOTNOlTES` heding should have the extra blank line added in between the heading and the first of the footnotes? (In another PR 😁)
> 
> When GG2 adds a LZ, it inserts two blank lines before the FOOTNOTES heading, which makes it a section heading. If the user adds more blank lines, or if there are extra blank lines at the LZ point, then it could end up with 4 blank lines, in which case the user would need to add another blank line after the heading. The pptext tool (I think) does a checklist of blank line spacings, which should alert the user to the problem.

When the LZ is at the end of the file, is it considered a chapter heading, or a section heading?

windymilla:
> When the LZ is at the end of the file, is it considered a chapter heading, or a section heading?

It's inserted as though it is a section heading, the same as LZs elsewhere. That could be changed of course (in another issue/PR 😬 )

srjfoo:
> > When the LZ is at the end of the file, is it considered a chapter heading, or a section heading?
> 
> It's inserted as though it is a section heading, the same as LZs elsewhere. That could be changed of course (in another issue/PR 😬 )
> 
> TBH, I'm not sure anyone would care whether the PPer submitted it as a chapter heading or a section heading.



> > When the LZ is at the end of the file, is it considered a chapter heading, or a section heading?
> 
> It's inserted as though it is a section heading, the same as LZs elsewhere. That could be changed of course (in another issue/PR 😬 )
> 
> TBH, I'm not sure anyone would care whether the PPer submitted it as a chapter heading or a section heading.

Okay, that helps, thanks. Basically, the PPer needs to decide whether it should be a section heading or a chapter heading if the footnotes are at the end of the file. So, either way, what was there was wrong: Chapter heading, add a blank line after; section heading, remove the two extra blank lines before. 

Given that, it's on the PPer.

windymilla:
> Given that, it's on the PPer.

Maybe a bit harsh, let's say 50-50 with GG2 taking the blame :) 

What would be possible is when the footnotes actually get moved to the LZs (wherever they are) if it finds the LZ title with 2 blank lines before, it could make sure there is just 1 blank line before the first footnote, but if there are 4 blank lines before the title, then it could ensure there are 2 blank lines before the footnote. What do you think?

srjfoo:
> > Given that, it's on the PPer.
> 
> Maybe a bit harsh, let's say 50-50 with GG2 taking the blame :)
> 
> What would be possible is when the footnotes actually get moved to the LZs (wherever they are) if it finds the LZ title with 2 blank lines before, it could make sure there is just 1 blank line before the first footnote, but if there are 4 blank lines before the title, then it could ensure there are 2 blank lines before the footnote. What do you think?

I'm good with that approach. It's still on the PPer, though, to decide whether it should be a section or a chapter, isn't it? And to make sure that the spacing is correct for whichever they want.

windymilla:
> I'm good with that approach. It's still on the PPer, though, to decide whether it should be a section or a chapter, isn't it? And to make sure that the spacing is correct for whichever they want.

Yes, final responsibility still lies with the PPer.

windymilla:
#691 added

rtonsing:
You can blame me :). I just didn't want the exception. Actually, the file needed more cleanup before converting.

PR #694, opened 2025-01-24
Prepare for alpha.14

windymilla:


CHANGELOG.md
```diff
@@ -1,6 +1,41 @@
...
+- When using Save As, GG2 now adds an appropriate extension, which is
+  `.txt` unless the file has an HTML header, in which case it is `.html`.
+  Also, if there is no filename, Save As suggests `untitled.txt`
+- Illo/Sidenote Fixup now automatically select the first message when run
+- HTML Autogeneration now reports more specific errors it discovers while
+  running, and allows user to re-load the previous backup to fix them
+
+### Bug fixes
+
+- Orange spotlights could be hidden by selected text
```
srjfoo, line 24: Is this about the select highlighting overriding the spotlighting in "Find all"? I've forgotten if there were other places this was happening, but if it's specific to "Find all", Ild mention that.

srjfoo (approved):
Looks good to me -- my only question was whether that one note should have more clarification.

CHANGELOG.md
tangledhelix: It would be anywhere the "spotlight" highlight is used (Find All, WF, Bookloupe, Jeebies, etc.)

srjfoo: > It would be anywhere the "spotlight" highlight is used (Find All, WF, Bookloupe, Jeebies, etc.)

Yes, but how many of tose are run on a selection?

tangledhelix: okrick's complaint, IIRC, is that when using these tools, you click in the tool to jump to a location, and perhaps you wish to make a change based on what you find. If you then try to select part or all of the spotlight-highlighted text, because "sel" was ranked lower than "spotlight," you could not see what text you had selected, making it very difficult to see what you're doing. So my recollection is it's more about not getting in the user's way when they are trying to select text that has been spotlight-highlighted.

tangledhelix: And so actually this text is wrong, isn't it? The problem was that the selection was being hidden by the spotlight?

tangledhelix: Reference: #631

tangledhelix: I think it should read instead "Selected text could be hidden by orange spotlights"

srjfoo: okrick's legitimate complaint before we changed priority of select and spotlight was that when spotlight was on top, and active, you couldn't see what you'd selected. When that was fixed, the select highlighting hid the spotlight in "find all in selection" results. As far as I know, but I don't remember for sure, the "find all in selection" was the only tool that selected things based on a selection. The specific fix for this was to remove the selection highlighting once the "find all in selection" was done, and not display the selection anymore, but only the spotlighting.

windymilla: Sharon's right - the comment is referring to [this PR](https://github.com/DistributedProofreaders/guiguts-py/pull/661). Although probably no-one else will have any idea what it's talking about, I'll clarify it slightly and then get it merged.

PR #695, opened 2025-01-25
Improve how code is executed on destroying dialog

### LINKED ISSUE #683: Improve handling of `<Destroy>` event in TopLevelDialogs
While reviewing #579, I noticed that the `<Destroy>` event was bound to the `top_frame` instead of the dialog. Further investigation reveals that this avoids the problem of the bound method being called once for every widget in the dialog (explanation below for future reference).

Suggestion:
1. In widgets.py, bind `tidy_up` to `self.top_frame` rather than `self`
2. In `tidy_up` there's no longer any need to check the widget class, since it will just be called once
3. Similarly in `checkers.py`, the `tidy_up` method doesn't need to check widget class (Also fix comment at top of method)
4. Consider renaming `tidy_up` to something with `destroy` in it
5. In `search.py` (once #579 is merged), instead of binding to `destroy`, override the `tidy_up` method (like checkers.py)

Explanation of difference between binding to dialog and binding to `top_frame`:
(Taken from [here](https://stackoverflow.com/questions/58328686/destroy-events-executes-the-binded-function-5-times))

1. Every widget has a set of 4 binding tags: itself, its class, its toplevel window, all
2. When you bind, you're binding to a tag, not to a widget
3. Therefore, if you pass `top_frame` (not a toplevel window) into the binding routine, it only matches the "itself" tag of the `top_frame` widget.
4. However, if you pass in `my_dialog` (a toplevel window), it matches not only the "itself" tag of the `my_dialog` widget, but also the `toplevel window` tag of all of its child widgets. So, you appear to have bound your method to the dialog and all its children.

windymilla:
1. Instead of binding to dialog, bind to `top_frame` so that binding is only called once, instead of for every dialog child.
2. Rename `tidy_up` to the more explicit `on_destroy`.
3. CheckerDialog just overrides `on_destroy`, calling the super method.
4. Search overrides the existing `reset` method instead of using an explicit binding.

Fixes #683

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -453,9 +453,6 @@ class CheckerDialog(ToplevelDialog):
         """Override method that tidies up when the dialog is destroyed.
 
         Needs to remove the undo_redo callback if there is one.
-
-        Args:
-            event: identifies the widget being destroyed.
         """
         super().on_destroy()
         maintext().remove_undo_redo_callback(self.__class__.__name__)
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -449,19 +449,15 @@ class CheckerDialog(ToplevelDialog):
         if entry_index is not None:
             self.select_entry_by_index(entry_index)
 
-    def tidy_up(self, event: tk.Event) -> None:
-        """Tidy up when the dialog is destroyed.
+    def on_destroy(self) -> None:
+        """Override method that tidies up when the dialog is destroyed.
 
-        Calls the reset method which may be overridden.
+        Needs to remove the undo_redo callback if there is one.
 
         Args:
             event: identifies the widget being destroyed.
         """
-        # Since this method is bound to the "<Destroy>" event on the dialog,
-        # it will also be called for all child widgets - ignore them.
-        if not issubclass(type(event.widget), ToplevelDialog):
-            return
-        super().tidy_up(event)
+        super().on_destroy()
         maintext().remove_undo_redo_callback(self.__class__.__name__)
 
     def new_section(self) -> None:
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes:
1. Should be no noticeable change for the user.
2. Particularly check that the new "highlight all matches" when you Search still works, and is cleared when you close the Search dialog.
3. Also that when a checker dialog (e.g. Jeebies) is dismissed, any visible spotlighting is removed.

tangledhelix (approved):

windymilla:
WF dialog should be OK as well now @srjfoo

srjfoo:
> WF dialog should be OK as well now @srjfoo

WF dialog no longer throws an exception, but
* the spotlighting is not removed, and
* although while the dialog is open, there's no visible cursor, after the dialog is dismissed, the cursor appear (which would be appropriate if there were no spotlighting).

windymilla:
> * the spotlighting is not removed, and
> * although while the dialog is open, there's no visible cursor, after the dialog is dismissed, the cursor appear (which would be appropriate if there were no spotlighting).

These are the same in `master` so I don't think they're caused by this PR. 

- I don't think spotlighting has ever been removed when WF dialog is dismissed, but it probably should be. Issue #714 added.
- (At least on Windows) When the focus is in the dialog, there is no visible cursor (true for all dialogs), but if you click to put focus in the main window, the cursor shows. This is why the cursor seems to appear when the dialog is dismissed - focus is returning to the main window. Also, it's not spotlighting that suppresses the cursor, it's selection highlighting.

windymilla:
@srjfoo - I decided the fix was so trivial for spotlight not being removed when WF is closed, that I've gone ahead and pushed another commit in this branch. So, the spotlight should now go when the dialog is closed.

srjfoo (approved):

PR #697, opened 2025-01-25
Ensure FOOTNOTES heading has correct blank lines

### LINKED ISSUE #691: Improve blank lines around FOOTNOTES headings
As described in #688, you can end up with 4 blank lines before a FOOTNOTES heading and only 1 blank line afterwards.
Suggestion: 
When moving footnotes to LZ, if the FOOTNOTES heading has 4 blank lines before, make sure that there are 2 blank lines before the first footnote

windymilla:
After moving footnotes to LZ, code now checks if the FOOTNOTES heading has 4 blank lines before it.
If it does, then make sure it has at least 2 blank lines after it.

Testing note: if you manually edit LZ blank lines in order to test this, you should re-run the footnotes tool to ensure it knows all the locations correctly.

Fixes #691

(Also added a new method to return the range of the whole file, and used it everywhere it previously created a range from
`maintext().start()` to `maintext().end()`. Just a bit of code improvement :) )

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -726,7 +726,7 @@ class File:
 
     def remove_page_flags(self) -> None:
         """Remove page flags."""
-        search_range = IndexRange(maintext().start(), maintext().end())
+        search_range = maintext().start_to_end()
         while match := maintext().find_match(
             PAGE_FLAG_REGEX,
             search_range,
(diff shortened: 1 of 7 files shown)


srjfoo (approved):
Testing note: if you manually edit LZ blank lines in order to test this, **_you should re-run the footnotes tool to ensure it knows all the locations correctly_**.

I resemble this remark, and cannot emphasize it enough! 🙄 Even after reading the instructions! You get very weird results if you forget!

PR #699, opened 2025-01-26
Fix bookloupe exception at end of table

### LINKED ISSUE #698: Exception thrown by Bookloupe
```
02:02:59: ERROR - root.py:63 - Tkinter Exception
  File "/opt/homebrew/Cellar/python@3.11/3.11.11/Frameworks/Python.framework/Versions/3.11/lib/python3.11/tkinter/__init__.py", line 1967, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/tools/bookloupe.py", line 855, in bookloupe_check
    _the_bookloupe_checker.check_file()
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/tools/bookloupe.py", line 146, in check_file
    self.run_bookloupe()
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/tools/bookloupe.py", line 177, in run_bookloupe
    self.check_para(para_first_step, step - 1, paragraph)
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/tools/bookloupe.py", line 350, in check_para
    if last_line[-1] not in para_punc:
       ~~~~~~~~~^^^^
IndexError: string index out of range
```

Text in the vicinity of the file where the error was thrown:

```
/*
==============+===========+======+===========+====+========+=========+========+=========+=======+==========+========+========
              |  George   | John |  Samuel   |John|  John  |Robert H.| George |  John   | John  |  James   | Edward |Benjamin
   STATES.    |Washington.|Adams.|Huntington.|Jay.|Hancock.|Harrison.|Clinton.|Rutledge.|Milton.|Armstrong.|Telfair.|Lincoln.
--------------+-----------+------+-----------+----+--------+---------+--------+---------+-------+----------+--------+--------
New Hampshire |      5    |   5  |     --    | -- |   --   |    --   |   --   |    --   |   --  |    --    |   --   |   --
Massachusetts |     10    |  10  |     --    | -- |   --   |    --   |   --   |    --   |   --  |    --    |   --   |   --
Connecticut   |      7    |   5  |      2    | -- |   --   |    --   |   --   |    --   |   --  |    --    |   --   |   --
New Jersey    |      6    |   1  |     --    |  5 |   --   |    --   |   --   |    --   |   --  |    --    |   --   |   --
Pennsylvania  |     10    |   8  |     --    | -- |    2   |    --   |   --   |    --   |   --  |    --    |   --   |   --
Delaware      |      3    |  --  |     --    |  3 |   --   |    --   |   --   |    --   |   --  |    --    |   --   |   --[**F2 removed bottom line as table continues next page]
*/
-----File: 019.png---------------------------------------------------------
/*
```

Based on some print-statement debugging, I believe the `*/` line is where the exception occurs (line 503 of attached file). Possibly because of `remove_inline_markup()` a couple lines earlier?

[presidents-utf8.txt](https://github.com/user-attachments/files/18549278/presidents-utf8.txt)

windymilla:
Last line of table consisted of symbols, like `=`. These all get stripped before checking if the last character is a suitable punctuation character. Exception was caused by checking last character of an empty string.

Fixes #698

### DIFF
--- a/src/guiguts/tools/bookloupe.py
+++ b/src/guiguts/tools/bookloupe.py
@@ -347,7 +347,7 @@ class BookloupeChecker:
         last_line = para_text.splitlines()[-1]
         last_line = self.remove_inline_markup(last_line)
         last_line = re.sub(rf"[^{para_punc}\p{{Letter}}\p{{Number}}]", "", last_line)
-        if last_line[-1] not in para_punc:
+        if last_line and last_line[-1] not in para_punc:
             self.dialog.add_entry(
                 "No punctuation at para end?",
                 IndexRange(maintext().rowcol(f"{end_index}-1c"), end_index),


tangledhelix (approved):

srjfoo:
The exception is gone, but now getting a "No punctuation at para end" where a paragraph is split between pages, but the next page is not a continuation of the text (i.e. has an illustration plate or blank page or both between text pages:

image

---

Is the mismatched single quote supposed to spotlight the whole block (double, also)? My apologies, but I honestly don't remember -- and this feels pretty overwhelming. Not so much if it's a short (2-3 line) para, but.... It looks like whole blocks are spotlighted for those two and also mismatched square brackets.

image

windymilla:
> The exception is gone, but now getting a "No punctuation at para end" where a paragraph is split between pages, but the next page is not a continuation of the text (i.e. has an illustration plate or blank page or both between text pages:

I'm afraid bookloupe (GG2 & GG1) is not sophisticated enough to allow for cases like that, so you will get a false positive. In RL, page seps would be removed and mid-para illos would be repositioned before running bookloupe I suspect. The thing is, you are running bookloupe to tell you about potential errors you might otherwise miss, so the PPer is better to deal with things they know they need to do first before running it, then let it find the things they've missed.

windymilla:
> Is the mismatched single quote supposed to spotlight the whole block (double, also)? My apologies, but I honestly don't remember -- and this feels pretty overwhelming. Not so much if it's a short (2-3 line) para, but.... It looks like whole blocks are spotlighted for those two and also mismatched square brackets.

Yes, it's "supposed" to, i.e. that is how it works. After it has gathered all the text for the paragraph it complains if there is an odd number of them (with some corrections/exceptions), so it just has the paragraph start & end for highlighting purposes, not the location of the "offending" quote. In fact, you don't know for sure which is the offending quote without some more complex stuff to do with trying to work out which were opening straight quotes and which were closing ones, bearing in mind that they may be poorly spaced. So, for now, I think we'll have to live with it. The only alternative I can think of would be to highlight the final character of the paragraph (to mean "Hey, I got to this point and was expecting to have found another quote by now") though of course the solution is not necessarily to add a quote at the end of the paragraph, which that might suggest to a PPer. But if you think it would be better, we could consider that.

srjfoo (approved):

PR #701, opened 2025-01-27
Only run search / highlight if search params changed

### LINKED ISSUE #700: Search performance is suboptimal after "highlight all matches" was added
Searching for a relatively common word like "of" in a larger file is not performant when you click `Search` to advance to the next match. Shortcut key like Cmd-G (Mac) is very fast, but clicking Search is slow.

(This is also true with less common words but it's still a noticeable difference in speed.)

Probably this is due to the all-matches highlighting which would happen in the click case but not in the hotkey case. Needs to be optimized.

tangledhelix:
If the search params have changed since last time the search button was clicked, then re-run the find-all to highlight matches. Otherwise don't.

Only the search pattern, match-case, regex, and whole-word are considered. While wrap-around and reverse matter for navigating search results, they do not matter in terms of the highlighter (which is the thing to avoid re-running).

Testing notes: everything about search should work the same as before and the way you expect. Except that navigating to next/prev match by clicking Search (or shift-clicking) or pressing enter (or shift-enter), should no longer be slow.

To stress-test this, grab a large and search for some common word (and, the, ...). On master, it should be slow to navigate between them with the search button or enter. In this branch, the initial search may not be fast, but navigating results should be.

If you then change any of the four parameters mentioned above, the "full search" will be done again.

The memorized params are part of the dialog instance, so if the dialog is closed they will be lost.

Fixes #700

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -298,6 +298,22 @@ class SearchDialog(ToplevelDialog):
             "<Destroy>", lambda _event: maintext().remove_search_highlights()
         )
 
+        class SearchParamsState:
+            """Ephemeral container to store information about search parameters.
+
+            Memorized parameter state can be compared to current parameter state
+            to determine whether the state changed since the last memorization.
+            """
+
+            def __init__(self) -> None:
+                self.pattern = ""
+                self.match_case = preferences.get(PrefKey.SEARCHDIALOG_MATCH_CASE)
+                self.regex = preferences.get(PrefKey.SEARCHDIALOG_REGEX)
+                self.whole_word = preferences.get(PrefKey.SEARCHDIALOG_WHOLE_WORD)
+
+        # container to track search parameter state
+        self.search_params_state = SearchParamsState()
+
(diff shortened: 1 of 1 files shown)


windymilla (approved):
This looks great & seems to perform as described.

I went down an interesting (to me!) rabbit hole regarding the SearchParamsState class, and using `@dataclass` as follows:

SearchParamsState init would grab the pattern from the search box instead of setting it to the empty string. It's empty when the dialog is created anyway.

Then search_params_memorize would become a 1 line method:
`self.search_params_state = SearchParamsState()`

And search_params_have_changed would use the way `==` works for `dataclass` classes and become
`return self.search_params_state == SearchParamsState()`

I haven't tried it out at all, and am happy with what you have, but I thought it was sufficiently interesting to share.

tangledhelix:
> I went down an interesting (to me!) rabbit hole regarding the SearchParamsState class, and using `@dataclass`

It does look interesting, although I don't think I want to rewrite this at present...

srjfoo (approved):

PR #716, opened 2025-01-30
Add unmatched HTML tag check

### LINKED ISSUE #454: Add unmatched HTML tag check
I think this is necessary, because the Nu validator will allow you not to close `<p>` for example, but we want tags such as that to be closed.

windymilla:
1. In the HTML menu, Unmatched HTML tags reports on open tags that are not closed, or close tags that were not opened.
2. It does not attempt to check whether the HTML is valid, for example, putting an `h1` inside a `span`. The rules on what is valid are too complex, and the rules on what would make sense syntactically even more so. Use the HTML validator to check if HTML is valid.
3. It also does not have a list of all the valid HTML tags, so will report
`<mytag class="myclass">some text</yourtag>`
because `mytag` isn't closed.
But it won't report 
`<mytag class="myclass">some text</mytag>`
even though `mytag` is not a valid HTML tag.

Fixes #454

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -40,6 +40,7 @@ from guiguts.misc_tools import (
     page_separator_fixup,
     PageSepAutoType,
     unmatched_dp_markup,
+    unmatched_html_markup,
     unmatched_brackets,
     unmatched_curly_quotes,
     unmatched_block_markup,
@@ -103,7 +104,7 @@ class Guiguts:
         root().tk.call("package", "require", "awdark")
         root().tk.call("package", "require", "awlight")
 
-        self.menu_file: Optional[Menu] = (
+        self.file_menu: Optional[Menu] = (
             None  # File menu is saved to allow deletion & re-creation
         )
         self.init_menus()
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: you'll probably find it easier to just create artificial examples, using either valid or invalid tag names. Note that the open tags can have other characters in them after the tag name, e.g. `<i lang="en">` but the close tag cannot, i.e. `</i>`

tangledhelix:
Did you consider using an HTML library for this? I don't want to make the assumption that you didn't, so I'm asking. My naive assumption has been that using one would be a better approach than doing this with regex, but I'm happy to accept your assessment if you considered that and decided regex was the correct approach.

windymilla:
> Did you consider using an HTML library for this? I don't want to make the assumption that you didn't, so I'm asking. My naive assumption has been that using one would be a better approach than doing this with regex, but I'm happy to accept your assessment if you considered that and decided regex was the correct approach.

Thanks Dan. I think an HTML library could certainly be useful for some of the checks we should add. For this specific check, there are a couple of points in favor of a home-grown solution:
1. The code for this test is very simple, and is just a slight adaptation of the unmatched inline DP markup check we already have, just with slightly more complex regexes to permit `<i class="taxonomy">` in addition to the basic `<i>`.
2. We don't want other HTML errors to stop us getting the missing close tag reports, which I suspect might happen (or at least be difficult to avoid) with an HTML library. This is because PPers often want to run checks at different points in their work, not necessarily following a "standard" workflow.
3. Lots of unclosed elements are permitted in HTML5. For example, you don't need to close `<p>` with `</p>`, and can just start a new paragraph with a fresh `<p>`. But we (DP?/PG?/Ebookmaker?) want things to be closed. So, although a library may provide a way of checking for "missing" close tags, it may not be as easy as checking error messages.

srjfoo:
`<link>`, `<meta>`, `<hr>`, `<br>`, `` are the self-closing tags that are going to be most common in our ebooks. To me, they just look like noise. Even if you ignore the first two, most projects are going to have some of the last three, aren't they?

Testing:

I picked a posted project at random and downloaded the HTML zip. After noting the above, I changed some of the open or close tags:
* closing `</p>` tag to `</b>`,
* opening `<p>` tag to `<div>`,
* closing `</p>` tag to `</div>`.

Only the first one was identified. The main difference that I see is that the first one, the tag type was changed from block level to inline, and the latter two were all block level, even though different tags (given that this feature is not supposed to be doing any kind of validation, though, that's probably a red herring).

In the test file, the three paragraphs were sequential. The first and third had a `<span class="pagenum" ...` within the para, but the middle one had no inline markup at all.

Maybe I'm misunderstanding what I'm supposed to be seeing? It did report `<mytag>` when I changed a `<p>` to that. Is it not identifying the `<p>`/`</div>` and `<div>`/`</p>` because there are matches in the file?

After playing around with it some more, I backed out the changes except for the `</p>` to `</b>`, and the paragraph following that, which is a `<div>` with a closing `<p>`. About 4 paragraphs down, there is a closing `</div>` at the end of the chapter.

I realize this feedback may be confusing, so let me know if you need clarification.

Summary
* Should self-closing tags be identified?
* I'm concerned that mis-matched pairs that I expected to be identified weren't.

srjfoo (approved):
A couple of comments:

Perhaps consider explicitly ignoring self-closing tags? You're not checking syntax or proper usage, and at least to me, they just add noise.

Should the documentation include the reminder that the location of the unmatched tag may not be where the actual error is? One of my tests, the "unmatched tag" that was actually identified was correct; the incorrect one was several paragraphs above it. And include the recommendation to re-run the test as errors are fixed?

windymilla:
@srjfoo - I don't think the documentation needs to say that. It's true of lots of checks and something that PPers should be used to. Essentially, it's saying there's an error and here is where I noticed it. The validator will be similar, and some footnote checks, e.g. if you miss the closing bracket and so one footnote runs into another, etc. I think that re-running as you fix is similar, in that it's not specific to this check, but to checks in general.

Good point on the self-closing tags. And, yay, past windy added an argument to the matching routine, called `ignore_reg` where you can pass in a regex that matches things you want to ignore, so I can pass in `ignore_reg="<(br|img).*?>",` (but with a few more tag types inside the _**parens**_ 😉)

srjfoo:
I tried adding `lang="fr"` to a closing `</p>` tag (also a class), and it was not identified, though the lang attribute added to a closing `</i>` tag did. If someone is careless enough to add something to a closing inline tag, mightn't they also add something to a block-level tag?

srjfoo (approved):

PR #722, opened 2025-01-31
Fix exception if WF is up when program quits

### LINKED ISSUE #720: Fix exception when program exits while WF is popped
@srjfoo - I decided the fix was so trivial for spotlight not being removed when WF is closed, that I've gone ahead and pushed another commit in this branch. So, the spotlight should now go when the dialog is closed.

_Originally posted by @windymilla in https://github.com/DistributedProofreaders/guiguts-py/issues/695#issuecomment-2624911334_
          
So trivial that I introduced a bug while adding the fix :grin:

windymilla:
Fixes #720

Testing notes:
Run Tools->WF, then quit the program without closing the WF dialog first.

### DIFF
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -442,7 +442,8 @@ class WordFrequencyDialog(ToplevelDialog):
         """Reset dialog."""
         super().reset()
         self.entries: list[WordFrequencyEntry] = []
-        maintext().remove_spotlights()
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
> Fix exception if WF is up when program quits

Dunno about Windows, but on my Mac, if I quit using `cmd`+`q`, there was no exception. However, if I closed using the red gummy, there was, and that case is fixed.

PR #724, opened 2025-02-01
Implement HTML validator

### LINKED ISSUE #450: Add HTML validator

windymilla:
Use the online nu validator via `requests` module

Fixes #450

Testing note:
Remember to run `poetry install` because it uses a new module.


rtonsing:
Tried it with several files, and saw no problems.
- Since only an HTML5 file will pass, should it specify "HTML5 Validator"?
- With an XHTML 1.0 file, it reports the error "Obsolete doctype" and lists other issues, which is much better than just barfing. Yet it is a valid file for its type, so it might be nice to make it more clear that the first error means it is not HTML5, and to use the online validator. Unless it is assumed and documented that GG only generates and supports HTML5.

tangledhelix (approved):
Tried on a number of files; seems to do what it says on the tin. The code looks good to me, though I didn't go through it with a fine-toothed comb... but you seem to do sane error checking and all...

I tried some HTML4, and it ran them, but complained predictably about the DOCTYPE and other things. That's to be expected.

windymilla:
> Tried it with several files, and saw no problems.
> 
> * Since only an HTML5 file will pass, should it specify "HTML5 Validator"?
> * With an XHTML 1.0 file, it reports the error "Obsolete doctype" and lists other issues, which is much better than just barfing. Yet it is a valid file for its type, so it might be nice to make it more clear that the first error means it is not HTML5, and to use the online validator. Unless it is assumed and documented that GG only generates and supports HTML5. That would be fine by me, don't know if others might disagree.

Thanks Robert.

- I'll change the menu entry to HTML5 Validator
- Since DP & PG are now strongly encouraging HTML5, there is no intention to make GG2 generate or support other types, such as XHTML 1.0. Having changed the menu entry as you suggested, I think that's probably sufficient warning for any PPer who tries to validate XHTML 1.0 files. I also think, given DP/PG's position, the "Obsolete doctype" is quite an accurate message.

Further note to developers, in case we go back on this decision: This [example python file](https://about.validator.nu/html5check.py) has info on arguments to pass for different file types, as well as using gzip to compress the sent file and response.

srjfoo (approved):
I intentionally changed an `</i>` to a `</b>` in one of my test files, and was delighted at how many errors it generated! 😁

Issue #705, opened 2025-01-28
Search highlights


okrick:
> When using the Search dialog to search for a string, all occurrences of the string are highlighted faintly, in addition to the first occurrence being selected as previously. These faint highlights are removed when the user begins a new search or closes the Search dialog

The faint highlights remaining after replacing a search term are distracting. Can the highlights of the replacement also disappear immediately?

![Image](https://github.com/user-attachments/assets/1a70a329-a4ca-4841-81fb-6f9ff1274248)

okrick:
The highlighting remains distracting when I edit the found item instead of replacing it. This was particularly noticeable when working with proofreader notes, such as when I accepted their suggestion for correcting a typo.

srjfoo:
The replacement search term should be the only thing that has the faint highlighting -- what was your replacement search term? The faint highlighting of the previous search should disappear when a new search is started, or when the search dialog is closed.

okrick:
The original search term was `/\*((\n|.)+?)\*/`. Despite replacing the selections, faint highlighting remains. I prefer to keep the Search & Replace dialog open while editing the found or replaced text, as closing it would require me to re-select the search and replace expressions from the drop-down menus, disrupting the workflow.

srjfoo:
Ah, I see ... the problem isn't with the search, it's that the replacement retains the highlighting?

tangledhelix:
@okrick I just opened PR #725 to address these issues. I can't add you as a reviewer there, so far as I can tell, but you're free to review it, of course.

The new implementation redraws these highlights during the idle-redraw part of the main event loop (similar to the alignment column or line numbers). Only the visible viewport(s) +/- about 80 lines each way are highlighted, to avoid performance problems in very large files while also trying to highlight multi-line blocks which may go off the edge of the viewport.

Since it's now dynamic, it should address all of your concerns. Responses below:

> The faint highlights remaining after replacing a search term are distracting. Can the highlights of the replacement also disappear immediately?

That should now be the case. So long as the replacement no longer matches the search pattern, the highlight will be removed. If the replacement happens to also match the search pattern, it will be re-highlighted, however.

> The highlighting remains distracting when I edit the found item instead of replacing it.

This should behave like the previous item - if your edit makes it no longer match the search, the highlight will go away.

> The original search term was `/\*((\n|.)+?)\*/`. Despite replacing the selections, faint highlighting remains.

[...]

PR #725, opened 2025-02-01
Reimplement "highlight all search matches"

tangledhelix:
This PR is mostly about improving / fixing the behavior of "highlight all matches" during a search. See #705 where I'll be going next to respond to some of the issues raised.

This PR includes:
- Moved `SearchDialog._find_all` to `MainText.find_all`. This was moved because trying to use it in the former location led to a circular import I couldn't resolve, and the method itself had no particular tie to `SearchDialog` or to search.py
- Improved `MainText.get_screen_window_coordinates` to deal with IndexRange, not a tuple of strings
- Reimplemented "highlight all matches" such that it will redraw the highlights on-demand during the viewport redraw phase, making it more dynamic, but also focusing only where the user can see, to avoid re-highlighting the entire file out of view. This brought a few fixes and improvements which I'll now post about in the #705 discussion.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -437,6 +437,10 @@ class MainText(tk.Text):
         self.aligncol = -1
         self.aligncol_active = tk.BooleanVar()
 
+        # whether search highlights should be active
+        self.search_highlight_active = tk.BooleanVar()
+        self.search_pattern = ""
+
         # Create Line Numbers widget
         self.linenumbers = TextLineNumbers(self.frame, self)
         self.linenumbers.grid(column=0, row=1, sticky="NSEW")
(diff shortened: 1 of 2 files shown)


tangledhelix:
@windymilla I'm happy to squash & rebase if you prefer, I haven't yet, in case you want to review each refactoring on its own. I refactored two things before doing the new highlighter work, and there's some linter cleanup after it too.

windymilla (approved):
Code and behavior look great. Thanks!

windymilla:
> @windymilla I'm happy to squash & rebase if you prefer, I haven't yet, in case you want to review each refactoring on its own. I refactored two things before doing the new highlighter work, and there's some linter cleanup after it too.

I'm happy either way thanks. If you don't squash & rebase before Sharon approves, I will squash when I merge. It's currently saying it can be merged with no conflicts.

srjfoo (approved):
LGTM!

PR #731, opened 2025-02-04
Don't create multiple tooltips

### LINKED ISSUE #730: Multiple tooltip windows?
I suspect a tooltip object is created each time bookloupe is re-run (and maybe other tools with the same calling structure). Then when the cursor enters the window, I suspect all the tooltips are displayed (probably covering each other up so the duplicates are unseen).

Possible solution - add an __init__ method to BookloupeCheckerDialog class, which calls the super's init, and adds the tooltip.

Check for other similar cases - i.e. where is the tooltip created.

windymilla:
Each time some of the tools were re-run, a new tooltip object was created. So, although not visible (because all the tooltips appeared on top of each other) after
several re-runs, each time the cursor entered the list of messages, several tooltips were displayed.

Fixes #730

Testing notes:
Should be nothing noticeable. Check tooltips work in some checker dialogs, Find All, Spelling, WF, etc.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -17,7 +17,7 @@ from guiguts.utilities import (
     is_mac,
     sing_plur,
 )
-from guiguts.widgets import ToplevelDialog, TlDlg, mouse_bind, Busy
+from guiguts.widgets import ToplevelDialog, TlDlg, mouse_bind, Busy, ToolTip
 
 MARK_ENTRY_TO_SELECT = "MarkEntryToSelect"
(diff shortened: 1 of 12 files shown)


srjfoo (approved):
Did this also need to be fixed for WF? I don't see it in the code changes (though I may just have missed it).

windymilla:
> Did this also need to be fixed for WF? I don't see it in the code changes (though I may just have missed it).

No - the difference between WF and other dialogs means that the bug doesn't happen for WF.

PR #744, opened 2025-02-08
Update README to include Poetry 2.x information

tangledhelix:
@srjfoo discovered that Poetry 2 breaks with former behavior for `poetry shell`, which is now deprecated.


srjfoo (approved):

windymilla (approved):
Thanks

Issue #483, opened 2024-10-05
Add shortcuts to scroll image viewer up/down


windymilla:
Shouldn't need focus to be in the viewer, but should be shortcuts that are unused in the main text

windymilla:
Would it also be good to have next/prev image buttons in the image viewer. They would just change the image, not move the text window at all. They would also need to turn off Auto Img.
One possible use case would be where, for whatever reason, there is no text between one page break and the next, so you can't click in that gap and then See Img.

rtonsing:
Not a big deal, but I would like it. I'm not 100% confident the buttons on the status bar will not skip non-text pages. In any case, I would like, at least sometimes, to go to next or previous page in the viewer without moving the text window.
This is helpful when one or more images have been moved to a paragraph break and no longer match the page number. I just want to see the image caption, not jump around in the text window.

srjfoo:
Going out on a limb here -- would it be better to pop a subordinate image viewer? I _**really**_ don't want to add more vertical space to the control block for the image pane. It's already significantly shorter than the text pane.

For instance, one other image control it might be nice to have is the ability to rotate an image.

windymilla:
> pop a subordinate image viewer

Do you mean a second viewer with the next/previous image in?

windymilla:
Another possible idea for whoever works on this: 
Add an "Enter" and "Leave" event binding. On Enter, show `<` and `>` buttons on the left & right edges halfway down, and on Leave, hide the buttons. 
Several apps do something like that:
![image](https://github.com/user-attachments/assets/462f1ec2-e5b6-4c53-81c7-5e0c8df19b8b)

tangledhelix:
I think with judicious use of Unicode (or image buttons), there's room to condense this entire control system while still having it make sense to humans. I made a PoC of a single-row control strip. There's room to improve this; it's just a mockup I threw together. If this is interesting, I'm happy to play with it more and turn it into a PR.

I didn't play much with weighting the various columns, but you can see the buttons have empty space, so they could be narrower than this to fit another couple of buttons (which might be ← and → or ❮ and ❯ or similar).

If there are concerns about clarity, we can always add tooltips on hover.

For me this reduced the control height from 71 pixels to 28 (60% reduction).

Before:

<img width="641" alt="Screenshot 2024-12-18 at 9 29 13 AM" src="https://github.com/user-attachments/assets/52f6fd8e-b850-4f43-8b96-33f266eef19c" />

After:

[...]

PR #745, opened 2025-02-10
Improvements to Image Viewer controls

tangledhelix:
Make image controls more compact, compressing into a single row.

Make zoom scaling a little finer (1.1 instead of 1.3).

Adds next / previous image buttons, allowing the user to navigate among images manually without affecting the currently displayed text. Using this navigation automatically turns off "Auto Image". The Image viewer will display any images in the image folder, even if they are not part of the base scanset (i.e., high-res scans, covers, etc., will also be displayed).

Addresses some, but not all, issues raised in #483.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -20,6 +20,7 @@ from guiguts.utilities import (
     process_accel,
     process_label,
     IndexRowCol,
+    sound_bell,
 )
 from guiguts.widgets import (
     ToplevelDialog,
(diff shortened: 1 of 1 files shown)


tangledhelix:
Testing notes:

- The previously-existing controls from the Image Viewer should continue to work as before; the buttons were just moved around.
- The new navigation controls `<` and `>` move to the previous or next image in the image directory, whether they're part of the scanset tagged in the page separators or not. This means you can go beyond the end of those scans and see the hi-res scans (assuming that they're in that same directory).
    - If you try to move previous to the first image, or past the last, you should hear a beep and nothing will happen.
    - Note that if auto-image is turned on, using these buttons will turn it off.
- Zooming in/out is a little less sensitive than before; the scaling factor is smaller, so the zoom happens in smaller steps. I'm completely open to feedback on this. The previous zoom step seemed a bit too large to me. I chose this value arbitrarily.

srjfoo (approved):

src/guiguts/mainwindow.py
```diff
@@ -534,6 +562,42 @@ def handle_configure(self, _e: tk.Event) -> None:
...
+            return
+
+        # Check current directory is valid
+        current_dir = os.path.dirname(self.filename)
+        if not os.path.isdir(current_dir):
+            logger.error(f"Image directory invalid: {current_dir}")
+            return
+
+        # Turn off auto-image; we're going to manual control.
+        if root().auto_image_state.get():
```
windymilla, line 582: I think it would be slightly preferable to replace these two lines with
`preferences.set(PrefKey.AUTO_IMAGE, False)`
i.e. to turn the Pref off, rather than directly manipulate the variable (although both should work). There are other places where auto-image gets turned off using the Pref.

windymilla (commented):
Looks and behaves great - just one suggestion made.

src/guiguts/mainwindow.py
tangledhelix: Ok, done

srjfoo (approved):

windymilla (approved):
LGTM - thanks!

PR #749, opened 2025-02-10
Improve HTML Auto-illustration

### LINKED ISSUE #736: HTML Auto-Illustrations suggestions
My _opinions_ on the generated HTML.
Example:
```
  <figure class="figcenter illowp56" id="i_06" style="max-width: 153.75em;">
    <img class="w100" src="images/i_06.jpg" alt="">
    <figcaption class="caption">
      <p>THE CONVERSION OF SAINT PAUL (DETAIL)</p>
    </figcaption>
  </figure>
```
In priority order:
1. Remove **\<p\>\</p\>** markup: it is not needed, and can add to or override CSS for the _\<figcaption\>_ or class defs.
I feel strongly about this one, I find the need for paragraph formatting in captions very rare, it just causes problems, such as left justifying instead of centering. It definitely contradicts and confuses the **class="caption"** or figcaption CSS, which styling will apply?
2. Remove **style="max-width: 153.75em;"**: if it is really needed, it should be included in the illowp56 CSS.
3. Remove **class="caption"**: I prefer to define CSS for the _\<figcaption\>_.
4. I think it is also adding an extra newline after it.

windymilla:
1. Remove `<p>` markup within caption
2. Remove `class=caption` from caption
3. Change HTML header file to make figcaption bold even without the `class=caption`
4. Remove superfluous trailing newline
5. Remove blank lines at start or end of caption

Fixes #736

### DIFF
--- a/src/guiguts/data/html/html_header.txt
+++ b/src/guiguts/data/html/html_header.txt
@@ -161,7 +161,7 @@ em.gesperrt
     font-style: normal;
 }
 
-.caption  {font-weight: bold;}
+figcaption  {font-weight: bold;}
 
 /* Images */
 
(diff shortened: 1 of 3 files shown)


rtonsing (approved):
All working as stated

PR #750, opened 2025-02-10
Don't mark up title with h1 and p

windymilla:
If the title (first bit of text in book) is marked up with `/*` or similar, then autogen surrounded it
with h1 and p markup.

Now, it just does h1.

### DIFF
--- a/src/guiguts/html_convert.py
+++ b/src/guiguts/html_convert.py
@@ -223,7 +223,8 @@ def html_convert_title() -> None:
             if re.match("[*$fxcr]/", selection, flags=re.IGNORECASE):
                 maintext().replace(f"{step}.0", f"{step}.end", "</h1>")
                 return
-            # Not complete - keep going
+            # Not complete - add <br> to end of previous line to preserve line breaks
+            maintext().insert(f"{step - 1}.end", "<br>")
             continue
         # Not in title yet, so skip blank lines, illos or block markup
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:

The following file shows the bug in `master`, but not in this branch. Removing the `/*` markup should also still work. 

```




/*
DRAGONS
AND
CHERRY-BLOSSOMS
*/




```

Because autogen only edits lines, rather than inserting or deleting them, the results will look like one of the following to preserve the number of lines previously occupied by the heading.
```
<h1>
DRAGONS
AND
CHERRY-BLOSSOMS
</h1>
```
or
```
<h1>DRAGONS
AND
CHERRY-BLOSSOMS</h1>
```

srjfoo:
Testing, it worked as described, but before I approve it, I've got a question.

If the heading does not include no-wrap markup, I assume that the PPer is okay with the title wrapping.

If the PPer leaves no-wrap markup, my assumption would be that the PPer does not want the book title to wrap. As coded, it will wrap.

windymilla:
> If the heading does not include no-wrap markup, I assume that the PPer is okay with the title wrapping.
> 
> If the PPer leaves no-wrap markup, my assumption would be that the PPer does not want the book title to wrap. As coded, it will wrap.

Now you mention it, it would be better if it just always added `<br>` to preserve the line breaks from the original. Usually that is what the PPer will want, and they can always remove it if they don't. They normally have to tweak the front matter considerably anyway.

srjfoo (approved):

PR #753, opened 2025-02-11
Image viewer & next/prev page improvement

windymilla:
Two changes which are independent to improve the interplay between the text window prev/next page and the image viewer prev/next file.

For discussion as to whether either, neither or both of these are improvements:
1. When using prev/next image buttons in status bar (or equivalent buttons in Search-->Goto menu) step back/forward to the next page boundary, even if that means the insert cursor doesn't move. This happens if there are two or more page breaks at the same point in the file, due to full page illo or blank page.
2. Don't turn off Auto Image when prev/next file buttons are used in image viewer. Instead, just pause it until the user moves the mouse out of the image viewer. So the PPer can easily look back/forward a page or two around their current insert location, then when they want to go back to the text to continue editing, auto image will resume. Of course, if the user doesn't want auto image they can always turn it off.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -27,6 +27,7 @@ from guiguts.mainwindow import (
     statusbar,
     ErrorHandler,
     process_accel,
+    mainimage,
 )
 from guiguts.misc_dialogs import (
     PreferencesDialog,
@@ -188,7 +189,10 @@ class Guiguts:
     def auto_image_check(self) -> None:
         """Function called repeatedly to check whether an image needs loading."""
         if preferences.get(PrefKey.AUTO_IMAGE):
-            self.mainwindow.load_image(self.file.get_current_image_path())
+            # Image viewer can temporarily pause auto image viewing,
+            # but still need to schedule another call to this method.
+            if not mainimage().auto_image_paused:
+                self.mainwindow.load_image(self.file.get_current_image_path())
             root().after(200, self.auto_image_check)
 
     def show_image(self) -> None:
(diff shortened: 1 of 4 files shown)


windymilla:
Additional note: with or without these changes (but particularly with the autoimg pausing change) it would be possible to add a "See" button to the image viewer, which would move the text window to the location corresponding to the current image - kind of a reverse one-off auto image operation.

rtonsing:
Seems to work well. I have a case where a paragraph starts on Pg 21 and ends on Pg 27, with 5 illos moved after the paragraph end. The arrow buttons in the image viewer and the status bar both properly page through the images, while moving and leaving the text cursor at the start of Pg 27.

tangledhelix (approved):
Seems to do what it says on the tin. I like it, but will leave this one remark:

> 2. Don't turn off Auto Image when prev/next file buttons are used in image viewer. Instead, just pause it until the user moves the mouse out of the image viewer.

This works fine with the mouseover model, but I can think ahead to a time where someone wants to do this navigation with keyboard shortcuts, at which point the mouse cursor position tells us nothing about the application state. However, this PR involves no keyboard shortcuts, so it's not a concern at present.

Caveat: potential for user to be confused as they navigated to a certain imaged, moused out, and the page flips back and they don't realize it happened (as with your luggage, "many pages look alike").

srjfoo (approved):

tangledhelix:
> Perhaps if we do set up a keyboard shortcut in the future, we could make it so the keyboard shortcut only works when the mouse is within the image viewer

In that case, maybe we'd also want to add a thin border around the entire image widget and change its color on Enter and again on Leave, or something...

windymilla:
> In that case, maybe we'd also want to add a thin border around the entire image widget and change its color on Enter and again on Leave, or something...

That could be a good idea, even with behavior as it is now.

PR #765, opened 2025-02-14
Highlight border of image viewer when focused

### LINKED ISSUE #759: Visible sign that "focus" is in image viewer
Although it's not true focus, #753 relies on the mouse being inside/outside the image viewer. As suggested there, it would be good to show that with a thin highlight border.

windymilla:
When mouse enters/leaves the image viewer, change the border color. Should work correctly with themes. If you change theme, the border won't change immediately, but will as soon as you enter/leave the image viewer. We already tell the user that they may need to restart to get full effect of a theme change.

This is because if the user has AutoImg on, and they use the image viewer buttons to change the image, when the mouse leaves the image viewer, the image will flip back to the AutoImg image.

Fixes #759 as suggested by @tangledhelix

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -427,6 +427,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_default(PrefKey.IMAGE_FLOAT_GEOMETRY, "400x600+100+100")
         preferences.set_default(PrefKey.IMAGE_DOCK_SASH_COORD, 300)
         preferences.set_default(PrefKey.IMAGE_SCALE_FACTOR, 0.5)
+        preferences.set_default(PrefKey.IMAGE_VIEWER_ALERT, True)
         preferences.set_default(
             PrefKey.SCANNOS_FILENAME,
             str(DEFAULT_SCANNOS_DIR.joinpath(DEFAULT_REGEX_SCANNOS)),
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -28,6 +28,7 @@ from guiguts.mainwindow import (
     ErrorHandler,
     process_accel,
     mainimage,
+    AutoImageState,
 )
 from guiguts.misc_dialogs import (
     PreferencesDialog,
@@ -183,6 +184,7 @@ class Guiguts:
         """Callback when auto_image preference is changed."""
         statusbar().set("see img", "Auto Img" if value else "See Img")
         if value:
+            mainimage().auto_image_state(AutoImageState.NORMAL)
             self.image_dir_check()
             self.auto_image_check()
 
@@ -191,8 +193,13 @@ class Guiguts:
         if preferences.get(PrefKey.AUTO_IMAGE):
             # Image viewer can temporarily pause auto image viewing,
             # but still need to schedule another call to this method.
-            if not mainimage().auto_image_paused:
+            auto_image_state = mainimage().auto_image_state()
+            if auto_image_state != AutoImageState.PAUSED:
                 self.mainwindow.load_image(self.file.get_current_image_path())
+                if auto_image_state == AutoImageState.RESTARTING:
+                    mainimage().alert_user()
+                    mainimage().auto_image_state(AutoImageState.NORMAL)
+
             root().after(200, self.auto_image_check)
 
     def show_image(self) -> None:
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -1,5 +1,6 @@
 """Define key components of main window"""
 
+from enum import Enum, auto
 from idlelib.redirector import WidgetRedirector  # type: ignore[import-not-found]
 import logging
 import os.path
@@ -29,7 +30,6 @@ from guiguts.widgets import (
     themed_style,
     theme_set_tk_widget_colors,
     Busy,
-    theme_set_tk_widget_highlightcolor,
 )
 
 logger = logging.getLogger(__package__)
@@ -195,6 +195,15 @@ class Menu(tk.Menu):
         self.add_button("Select ~All", "<<SelectAll>>", "Cmd/Ctrl+A")
 
 
+class AutoImageState(Enum):
+    """Enum class to store AutoImage states to facilitate pausing
+    and restarting AutoImage when needed."""
+
+    NORMAL = auto()
+    PAUSED = auto()
+    RESTARTING = auto()
+
+
 class MainImage(tk.Frame):
     """MainImage is a Frame, containing a Canvas which can display a png/jpeg file.
 
@@ -234,7 +243,20 @@ class MainImage(tk.Frame):
         self.dock_func = dock_func
         self.allow_geometry_storage = False
 
-        control_frame = ttk.Frame(self)
+        # Introduce an apparently superfluous Frame to contain everything.
+        # This is because when MainImage is undocked, Tk converts it to a
+        # Toplevel widget. Any bindings on the MainImage are then reapplied
+        # to the Toplevel widget, and Toplevel widgets apply their bindings
+        # to all their children. So an Enter/Leave binding on MainImage would
+        # cause an Enter/Leave to trigger when the mouse enters/leaves a button,
+        # for example. Binding to top_frame instead means that the binding
+        # doesn't get reapplied to the Toplevel widget, and it continues to
+        # behave as required.
+        top_frame = ttk.Frame(self)
+        top_frame.grid(row=0, column=0, columnspan=2, sticky="NSEW")
+        self.rowconfigure(0, weight=1)
+        self.columnconfigure(0, weight=1)
+        control_frame = ttk.Frame(top_frame)
         control_frame.grid(row=0, column=0, columnspan=2, sticky="NSEW")
         control_frame.columnconfigure(8, weight=1)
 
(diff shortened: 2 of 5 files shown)


tangledhelix:
I haven't reviewed this yet but just a thought that came to me - since we have the invert mechanism already, another way to easily signal to the user that something happened would be to flash the image by inverting the value of IMAGE_INVERT, then flipping it back say 150ms later. The effect is the entire thing flashes momentarily which is a strong signal that something changed.

(I might just be too paranoid that the current shift is too subtle to catch the user's attention).

I implemented a simple PoC (below), although this one is a bit stupid, it flashes every time you leave the image viewer. It would only be of use if it detected when the image being displayed was about to change as the leave event is processed; otherwise it flashes for no reason.

Thoughts? I'm imagining @srjfoo about to write how the screen would be blindingly bright in dark mode with invert on :) 

```diff
diff --git a/src/guiguts/mainwindow.py b/src/guiguts/mainwindow.py
index d9d6049..a2b865a 100644
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -575,7 +575,14 @@ class MainImage(tk.Frame):
         """Handle leave event."""
         self.handle_configure(event)
         self.auto_image_paused = False
+        # TODO: only do this if the image shown is not the same as what we're about to flip back to
+        self.flash_image()
+        self.after(150, self.flash_image)
         theme_set_tk_widget_highlightcolor(self.canvas)
+    
+    def flash_image(self) -> None:
+        preferences.set(PrefKey.IMAGE_INVERT, not preferences.get(PrefKey.IMAGE_INVERT))
+        self.show_image()
 
     def next_image(self, reverse: bool = False) -> None:
         """Load the next image alphabetically.
```

srjfoo:
> Thoughts? I'm imagining @srjfoo about to write how the screen would be blindingly bright in dark mode with invert on :)

😁 

Eh ... while that's true, I'm more likely to grumble about a flashing screen. I dislike large flashing images intensely. No, I'm fortunately not one of those who has problems with epilepsy given flashing images, but I do turn off automatic animations wherever I can, and a flash for every page forward or backward counts towards that for me. I would imagine that others might feel the same way.

That being said, I agree, the current outline is a bit too subtle. I've played with various combinations of light and dark OS theme and GG2 theme.

Would it be reasonable to:
1. Use either the spotlight color or the default system highlight color (I slightly prefer the spotlight color because it's predictable, and some people may not have a very noticeable system highlight color), and
2. Double the width of the outline? Doubling might not even be needed if the outline is noticeable enough. The other thing I've seen in some places is, in addition to the outline, a (very) slight dimming of the area being outlined. (i.e. slightly lighter on the dark theme and slightly darker on the light theme) But not enough to significantly affect readability.

tangledhelix (changes requested):
When the image viewer is docked, this works like I expect:
- If the mouse enters the image viewer area (by which I mean the image plus its controls), the border is applied.
- I can use next/prev, zoom, etc., and everything works fine.
- If the mouse leaves the image viewer area, the border vanishes and the image changes back to what auto-image wants to show.

But if the image viewer is **not** docked, it's buggy:
- If the mouse enters the image viewer area, the border correctly appears.
- If the mouse leaves the area, the border correctly vanishes.
- However, the auto-image handling isn't quite right. I can click on next or prev to change the image (and that works), but in this case as soon as the mouse leaves **the button**, the auto-image goes back into force.

I don't see the issue in the code; it must be the way the binds are handled is different between the two modes, though?

srjfoo:
> * However, the auto-image handling isn't quite right. I can click on next or prev to change the image (and that works), but in this case as soon as the mouse leaves **the button**, the auto-image goes back into force.

I'm not entirely sure what the issue is here? As of PR #753, Auto Img is not turned off at all, but only paused while the mouse hovers over the image. Am I missing something?

srjfoo:
I've played around with the border around the image, both width and color. The width is set to 2. One of the discoveries that I made is that, once a width is set, the border is there, whether or not it's being used for highlighting, so I think that leaving it at 2 is probably best. One question, though -- can we use padding instead of border? That would make the view port slightly smaller, but would not overlap the image, I think?

As far as color, I like using the `darkorange` spotlight color for both dark and light. It shows up well on either version of the theme, and reasonably well at width 2.

tangledhelix:
> I'm not entirely sure what the issue is here? As of PR #753, Auto Img is not turned off at all, but only paused while the mouse hovers over the image. Am I missing something?

But that's not the behavior when un-docked. As soon as the mouse is no longer over the button you pressed (either Next or Previous) it will flip back to auto-image, you don't have to leave the image area, only the single button's area. The behavior when docked is fine.

srjfoo:
> As soon as the mouse is no longer over the button you pressed (either Next or Previous) it will flip back to auto-image, you don't have to leave the image area, only the single button's area.

Ah, okay. Now I see what you were saying. I wasn't looking at the right thing when I was testing it undocked. Yes, I'd consider that behavior broken.

windymilla:
> > I'm not entirely sure what the issue is here? As of PR #753, Auto Img is not turned off at all, but only paused while the mouse hovers over the image. Am I missing something?
> 
> But that's not the behavior when un-docked. As soon as the mouse is no longer over the button you pressed (either Next or Previous) it will flip back to auto-image, you don't have to leave the image area, only the single button's area. The behavior when docked is fine.

I'm not refusing to fix this under this PR, and it's certainly a bug. However, it's not caused by this PR and behaves exactly the same in `master`.

windymilla:
I'm wondering if a compromise would satisfy the flash-o-phobes and the flash-o-philes ;) 
1. Under normal circumstances have no border (like it is in `master`)
2. When the mouse leaves the image viewer, instead of flashing the whole image using Invert, just show, then hide a 2 (or more) pixel wide darkorange border.
If/when I get time later, I may try implementing that to see what y'all think.

windymilla:
Pretty sure I know what the "Leave when undocked" bug is & an idea of how to fix it. The image viewer is a Frame when docked and has a binding for the Leave event. When you undock it, Tk changes the Frame into a tk.Toplevel widget and tells the window manager to manage it, and (presumably) transfers the bindings that were on the frame to bindings on the Toplevel widget. However, if you add a binding to a top level widget, it applies the binding to the widget and all its children. So when you leave a button, the leave handler is called. 

See this from our own ToplevelDialog class:
```
        # Bind to top_frame being destroyed because if binding is to dialog,
        # function will be called for every child of dialog due to their
        # "top level" bind tag
```

windymilla:
OK, hopefully latest commit will be a reasonable compromise. It fixes the "undocked" bug reported by @tangledhelix, and only flashes the border, rather than the whole image under the following circumstances:
1. Auto Image is on
2. User has manually viewed another image using the prev/next buttons
3. Mouse leaves the image viewer window
4. Auto Image loads a different image to the one currently in image viewer

So the flash of orange on the border is an alert to the user that the image they chose has been overridden by AutoImg.

tangledhelix:
> OK, hopefully latest commit will be a reasonable compromise. It fixes the "undocked" bug reported by @tangledhelix, and only flashes the border, rather than the whole image under the following circumstances:
> 
> 1. Auto Image is on
> 2. User has manually viewed another image using the prev/next buttons
> 3. Mouse leaves the image viewer window
> 4. Auto Image loads a different image to the one currently in image viewer
> 
> So the flash of orange on the border is an alert to the user that the image they chose has been overridden by AutoImg.

First things first - everything seems to work, so my new review is "all good, ship it"

Minor nit; for me the highlight is still a bit subtle and I feel like I wouldn't always notice it. This will of course vary from person to person with their particular eyesight, screen, colorscheme, etc.

I played with various values, as high as 40 (which is definitely noticeable but a touch ridiculous). My proposal is 4, which I felt worked well but remains appropriately understated. (Like I said, these things are just a matter of opinion...)

srjfoo:
I agree that the thin line is not quite as visible in the light theme as it is in the dark theme, when it's flashed on and back off again. It's plenty good enough at 2 in dark mode, but 4 would be okay, too.

I still don't like the flash. It gives me the creeps at 150ms. When I increased the interval to 500ms, it was okay. But that may just be me.

That being said -- I think this is something that should be a preference, because I'm pretty sure there are some people who will want to turn it off and take their chances. If they know that if they have auto img turned on, once the mouse leaves the image, the image will revert to where it was before, then they can just deal with it.

> I played with various values, as high as 40 (which is definitely noticeable but a touch ridiculous). My proposal is 4, which I felt worked well but remains appropriately understated. (Like I said, these things are just a matter of opinion...)

😁 -- I went to 60, just to make it really, really obvious. And yes, definitely ridiculous. 99.995% of the time, it won't matter, but some proofing images are already really, really close to the edge of the page, or possibly even slightly cut off. Anyway, that's why I asked if we could use padding instead. But I was wrong -- I just checked and I was getting margin and padding backwards.

---

I still think I'd rather have the option of a split image screen, as we do with text. I'm much more likely to want to have two images open to compare than to want to just take a quick look and then revert to the main image. But I can also just open the image in a different application.

My apologies if sending us down this route was the wrong approach (i.e making it possible to keep auto img turned on instead of disabling it when the user scrolled to an image of a page other than what was displayed).

windymilla:
Pushed commit to address review comments:
1. Add checkbox to Prefs dialog (with tooltip)
2. Increase border width to 4
3. Increase border flash time to 500ms

windymilla:
> I still think I'd rather have the option of a split image screen, as we do with text. I'm much more likely to want to have two images open to compare than to want to just take a quick look and then revert to the main image. But I can also just open the image in a different application.

#768 added (I also considered whether I need to add a "Distant Future Feature" label in addition to the "Future Feature" we already have 😜

tangledhelix (approved):
Looks good! Tested the preference as well, works as advertised.

srjfoo:
> > I still think I'd rather have the option of a split image screen, as we do with text. I'm much more likely to want to have two images open to compare than to want to just take a quick look and then revert to the main image. But I can also just open the image in a different application.
> 
> #768 added (I also considered whether I need to add a "Distant Future Feature" label in addition to the "Future Feature" we already have 😜)

Totally fair. As I said, I (and other users) can always bring up an external image viewer if they don't like the way this works.

srjfoo (approved):
It'll be interesting to see other users' reactions.

Issue #771, opened 2025-02-16
Search highlighting / Find All spotlighting priority clash


windymilla:
1. Search for any string that occurs multiple times in the file.
2. Click Find All
3. Click an entry in the Find All dialog - The entry you click on does not result in an orange spotlight because the search highlight takes priority. 
4. Close the Search dialog (not the Find All) - The orange spotlight appears.

The above would be a common use case for the Find All dialog. Not being able to see which Find All entry you have clicked on seems undesirable. 

Suggest swapping the order of SEARCH & SPOTLIGHT highlights.

windymilla:
I'm happy to do this if you think it is sensible @tangledhelix

tangledhelix:
I think this is fine. Trying to recall why it is the way it is - it may be because the original version of `HighlightTag.SEARCH` highlighted both the active and background matches, before I changed to the current version which uses `sel` for the active highlight. If you swap those two, `sel` will remain above both of them, which is still correct.

Issue #757, opened 2025-02-12
Consider HTML syntax highlighting


windymilla:
This would make it easier in the later stages of PPing when the user is making manual edits to the autogenerated HTML.

NTS: Use the HTML parser (like HTML link checker) - could probably get away with doing a section around the currently displayed page, like the search string highlighting - send the string to the parser and use Tk tags to mark up the HTML tags, possibly different colors for the common tag types, like `<p>, <span>, <div>, etc.`

tangledhelix:
Just thinking ahead; highlighting in a rolling window is a good idea if you want live highlighting, but you will end up with unpaired tags almost certainly... at least sometimes. Particularly long-running block elements like `<p>` and `<div>`, but in some projects maybe `<table>` too. What I've done in other highlighters is just consider a decent off-screen window, say 80 lines above and below the viewport, but it's hard to know how much is enough.

So, the question is what the parser returns if, for example, the string it analyzes begins with text followed by `</p>` because it cannot see far enough back to see the `<p>`. Will it happily match and we can highlight, or does it do anything toward validating the markup?

windymilla:
I've yet to try it for this purpose of course, but from what I see in how it works for the link checker, the simple parser just finds all the open & close tags, without any judgement as to whether they are ordered/paired in a valid way. So, I think it's fine.

PR #772, opened 2025-02-16
Add simple HTML syntax highlighting

windymilla:
Highlight HTML-style tags with color.

Note that the simple python HTML parser does fully detect invalid tags and/or may be confused by things that might also confuse a browser, e.g. unescaped `<`.

I have chosen a few common tags (a,p,span,div) to have their own color. All other valid tags get purple. All unrecognised tags are shown in red.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -63,7 +63,7 @@ from guiguts.misc_tools import (
     check_curly_quotes,
 )
 from guiguts.page_details import PageDetailsDialog
-from guiguts.preferences import preferences, PrefKey
+from guiguts.preferences import preferences, PrefKey, PersistentBoolean
 from guiguts.root import root
 from guiguts.search import show_search_dialog, find_next
 from guiguts.spell import spell_check
@@ -467,6 +467,7 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         preferences.set_callback(
             PrefKey.HIGHLIGHT_QUOTBRAC, self.highlight_quotbrac_callback
         )
+        preferences.set_default(PrefKey.HIGHLIGHT_HTML_TAGS, False)
         preferences.set_default(PrefKey.COLUMN_NUMBERS, False)
         preferences.set_callback(
             PrefKey.COLUMN_NUMBERS, lambda value: maintext().show_column_numbers(value)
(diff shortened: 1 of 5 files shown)


windymilla:
Notes on colors. It's very easy to make any specific tag type have its own color (as long as someone tells me what color they want). We can have several tags with the same color, all different, etc. Very easy to configure in the code.

windymilla:
Fixes #757

srjfoo:
The purple and blue are too dark on macOS default-dark:

image

The yellow is way too light on macOS default-light

image

And perhaps it's just that I need cataract surgery, but the purple on macOS defalut light is so close to black that I can't tell the difference when looking at the screen in the app (the screenshot is a bit more distinguishable).

I think, too, that it may have something to do with me just being more used to the bolder appearance of light text on a dark background -- all of the colors look washed-out to me on a light background.  I'm game to go with whatever makes @tangledhelix happy.

rtonsing:
In Dark mode, the yellow is hard to distinguish from white. Dark blue is unreadable.
In Light mode, purple is hard to distinguish from black. Whatever span is, it is indistinguishable.
The colors are much more obvious in dark mode.

Overall, I don't find this useful, it looks confusing. I would prefer some kind of matching start/end tag highlighting for the current line. In VSCode, if the cursor is in a tag, it highlights it and the matching one.

windymilla:
> Overall, I don't find this useful, it looks confusing. I would prefer some kind of matching start/end tag highlighting for the current line. In VSCode, if the cursor is in a tag, it highlights it and the matching one.

Yes, and that could possibly be added, though not so simply as this was (using a standard simple parser module), but VSCode colors the HTML as well, at least it does on my system (see below). There is also already an "unmatched HTML tags" feature to help with checking for that.

image

windymilla:
Also fixes #771, if you could confirm that please @srjfoo

tangledhelix:
> In VSCode, if the cursor is in a tag, it highlights it and the matching one.

That would be similar to the bracket highlights we're doing now, although a bit more complicated of a matching algorithm to put together...

For me VScode doesn't have that behavior, at least not in the base HTML syntax mode. You may have a plugin or something though? VScode is closer to what Nigel's done; although it differentiates the attributes and attribute values in differing colors instead of coloring the whole tag and its contents together. But as far as I can tell, surrounding tag around the cursor isn't different than any other tag on the screen. (Not that important; I'm just pointing out what you're describing may be coming from a third-party plugin.)

tangledhelix:
> For me VScode doesn't have that behavior, at least not in the base HTML syntax mode. 

Apologies for mis-reading, you're talking about the cursor being in a tag - not between tags.

srjfoo (approved):

tangledhelix (approved):
It works; we can tweak colors some other time.

PR #776, opened 2025-02-18
Add buttons to checker dialogs

### LINKED ISSUE #752: Checker dialog ease-of-use improvements
Where checker dialogs have the ability to Shift, Ctrl, and left/right click messages, should there be equivalent buttons that allow the user to click them without holding keyboard modifier keys?

Notes:
Ctrl means "fix"
Shift means "do same to all matching messages" (idea of "matching" may vary from one checker to another)

There would be up to six buttons needed, because simple left-click to select a message doesn't need a button, and Shift-left-click doesn't make sense.
1. Ctrl-left --> "Fix"
2. Ctrl-Shift-left --> "Fix all matching"
3. Right --> "Remove" (and select next message)
4. Shift-right --> "Remove all matching" (similar to Skip All in spelling)
5. Ctrl-right --> "Fix & remove" (and select next, so similar to Replace & Search in S/R)
6. Shift-Ctrl-right --> "Fix & remove all matching" (similar to Replace All in S/R)

Although most checker dialogs wouldn't have 6 new buttons, most would have at least 2 or 3. Up to 6 additional buttons would make the UI more cluttered, so if this is a good idea, I think it should be controlled by a preference. That preference could be a specific "Hide shortcut buttons in checkers" checkbox, or a more generic one that controls a simple/advanced interface switch or maybe something relating to accessibility.

windymilla:
There are either 2 or 6 buttons, depending on whether the checker is able to Fix a reported problem. The buttons should resize with the dialog width.

Each button has a tooltip, and clicking on each one is equivalent to using Ctrl/Cmd/Shift Left/Right mouse buttons in the message list, but just using a simple mouse click.

Fixes #752

I have decided for now not to add a Preference setting to control whether these buttons are shown.

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -320,20 +320,20 @@ class CheckerDialog(ToplevelDialog):
         if generic_buttons:
             rem_btn = ttk.Button(
                 message_controls_frame,
-                text="Remove",
+                text="Hide",
                 command=lambda: self.remove_entry_current(all_matching=False),
             )
             rem_btn.grid(row=0, column=0, sticky="NSEW")
-            ToolTip(rem_btn, "Remove selected message (right-click)")
+            ToolTip(rem_btn, "Hide selected message (right-click)")
             remall_btn = ttk.Button(
                 message_controls_frame,
-                text="Rem. All",
+                text="Hide All",
                 command=lambda: self.remove_entry_current(all_matching=True),
             )
             remall_btn.grid(row=0, column=1, sticky="NSEW")
             ToolTip(
                 remall_btn,
-                "Remove all messages matching selected one (Shift right-click)",
+                "Hide all messages matching selected one (Shift right-click)",
             )
             if process_command is not None:
                 fix_btn = ttk.Button(
@@ -357,7 +357,7 @@ class CheckerDialog(ToplevelDialog):
                 )
                 fixrem_btn = ttk.Button(
                     message_controls_frame,
-                    text="Fix&Rem.",
+                    text="Fix&Hide",
                     command=lambda: self.process_remove_entry_current(
                         all_matching=False
                     ),
@@ -365,11 +365,11 @@ class CheckerDialog(ToplevelDialog):
                 fixrem_btn.grid(row=0, column=4, sticky="NSEW")
                 ToolTip(
                     fixrem_btn,
-                    f"Fix selected problem & remove message ({cmd_ctrl_string()} right-click)",
+                    f"Fix selected problem & hide message ({cmd_ctrl_string()} right-click)",
                 )
                 fixremall_btn = ttk.Button(
                     message_controls_frame,
-                    text="Fix&Rem. All",
+                    text="Fix&Hide All",
                     command=lambda: self.process_remove_entry_current(
                         all_matching=True
                     ),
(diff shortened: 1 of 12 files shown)


srjfoo:
Looking at
image
(the spell checker) and comparing the tooltip with that last row of buttons, I don't think that last row should even be there. "Fix" doesn't make sense in the spell checker, does it? All of the buttons in the second row correspond to the actions in the tooltip, don't they? Or am I missing something?

windymilla:
> comparing the tooltip with that last row of buttons, I don't think that last row should even be there. "Fix" doesn't make sense in the spell checker, does it? All of the buttons in the second row correspond to the actions in the tooltip, don't they? Or am I missing something?

Yes and no ;)

The whole point of the last row of buttons is precisely to provide equivalent buttons that correspond to the actions in the tooltip. In general for checker dialogs this is to support users for whom holding Ctrl/Cmd/Shift while clicking the mouse is not easy.

However, in this specific dialog, the key point in favor of your suggestion to remove the last row of buttons is that they are
(poorly worded) equivalents of the buttons on the row above:
Remove = Skip
Remove All = Skip All
Variants of Fix = Add to Proj Dict

So, I've disabled the "generic" buttons in the Spell Dialog. I think they are probably relevant in the other checker dialogs though.

srjfoo:
Part 1 of nits (to be continued, because it's time to stop for the night, but I should go ahead and ask about these):
* I'd rather see "Remove" spelled out in "Rem. All". Most of the buttons have everything spelled out, even with text that is much longer than "Remove All" would be.
* If you have to abbreviate button labels (and "Prev." and "Col" are good places to), standardize: either punctuate all abbreviations, or don't punctuate any.
* Actually, "Hide" makes more sense to me than "Remove" but I suppose "Remove" has the weight of "Tradition" behind it, so ... 🤷‍♀️ 😁 
* Curly Quotes check doesn't have the "shift" version of the tooltip when hovering over the results, but has the buttons for it.
* I know that I didn't like the generic buttons in the spelling checker, especially because there were already custom buttons there, but I just noticed that the custom buttons say "skip" and "skip all" instead of "Remove" and "Remove All" (the tooltip uses "Remove"). Should they be one or the other?

windymilla:
> Part 1 of nits (to be continued, because it's time to stop for the night, but I should go ahead and ask about these):
> 
> * I'd rather see "Remove" spelled out in "Rem. All". Most of the buttons have everything spelled out, even with text that is much longer than "Remove All" would be.
> * If you have to abbreviate button labels (and "Prev." and "Col" are good places to), standardize: either punctuate all abbreviations, or don't punctuate any.
> * Actually, "Hide" makes more sense to me than "Remove" but I suppose "Remove" has the weight of "Tradition" behind it, so ... 🤷‍♀️ 😁
> * Curly Quotes check doesn't have the "shift" version of the tooltip when hovering over the results, but has the buttons for it.
> * I know that I didn't like the generic buttons in the spelling checker, especially because there were already custom buttons there, but I just noticed that the custom buttons say "skip" and "skip all" instead of "Remove" and "Remove All" (the tooltip uses "Remove"). Should they be one or the other?

* I like the word Hide - it's a better description of what is happening, and is shorter
* Removed period after "Prev." & "Chap."
* Changed Remove to hide in many tooltips
* Changed Remove to Skip for the Spelling dialog

srjfoo (approved):

PR #777, opened 2025-02-18
Strip footnote anchors from auto ToC

### LINKED ISSUE #775: HTML generation - chapter headers with footnotes
I've just run HTML auto-generate on a file with a footnote anchor in a chapter header. The footnote is handled correctly except that the anchor (which should be in the chapter header) is instead added to the entry in the auto-generated table of contents. I assume that this is because linking to the first anchor (and not creating duplicate anchors) would usually be the preferred behaviour for footnotes with mutiple anchors.

In an ideal world, GG2 would not include the footnote anchor in the ToC entry at all, i.e. the entry for that chapter would show up as "POPULAR LEGAL FALLACIES." not "POPULAR LEGAL FALLACIES.[1]" so I suspect (without having looked at the code!) that it might be easier to fix the ToC generation to strip out the footnote anchor than try to determine whether the first footnote anchor appears in a ToC entry or not.

(I've attached a zip file with the relevant .txt and .json files from immediately before auto-generate is run. There is only one footnote so it should be easy to see what's going on. Let me know if anything else would be helpful.)

[18861106-htmlbak.txt.zip](https://github.com/user-attachments/files/18847801/18861106-htmlbak.txt.zip)

windymilla:
Problem occurs if chapter heading has FN anchor.

Because only one anchor can be linked from a footnote, this should be the one in the chapter heading itself, not one in the Autogenerated ToC.

Fixes #775

### DIFF
--- a/src/guiguts/html_convert.py
+++ b/src/guiguts/html_convert.py
@@ -685,6 +685,8 @@ def html_convert_body() -> None:
                 maintext().replace(
                     f"{line_start}-2l lineend", f"{line_start}-1l lineend", "</h2>\n"
                 )
+                # Don't want footnote anchors in auto ToC
+                chap_heading = re.sub(r"\[.{1,5}\]", "", chap_heading)
                 auto_toc += f'<a href="#{chap_id}">{chap_heading}</a><br>\n'
                 in_chap_heading = False
             continue


windymilla:
Testing notes - see linked issue for test file.

src/guiguts/html_convert.py
```diff
@@ -685,6 +685,8 @@ def reset_ibs_dict() -> None:
                 maintext().replace(
                     f"{line_start}-2l lineend", f"{line_start}-1l lineend", "</h2>\n"
                 )
+                # Don't want footnote anchors in auto ToC
+                chap_heading = re.sub(r"\[.{1,5}\]", "", chap_heading)
```
srjfoo, line 689: Should there be a warning in the manual that square brackets in the auto-ToC will be treated as footnote anchors, for the very rare cases where there might be legitimate square brackets in a title that is included in the autogenned ToC?

srjfoo (approved):

PR #780, opened 2025-02-19
Allow user to configure external viewer

windymilla:
1. Add checkbox for "External" in Prefs dialog
2. Add field for image viewer/command to be used to Prefs dialog
3. Also button to allow user to browse to find image viewer
4. Add "High contrast" checkbox to Prefs dialog which causes internal image viewer to use white/black backgrounds
5. If External viewer is set, then run external viewer to show each page image.
6. User can also choose View Menu->Show Image Viewer which will mean the internal image viewer will be shown the next time a file is loaded.


windymilla:
Testing notes: After you've enabled external viewer and got the path right, you will need to click See Img to see an image, or if you are using AutoImg, you'll need to go to a new page, so that the image will be loaded.

srjfoo:
On macOS, browse to Preview generates:
```
18:55:25: ERROR - mainwindow.py:515 - Unable to execute "/System/Applications/Preview.app"
on file "/Users/sharonjoiner/Desktop/guiguts-testing/_projects_for_testing/blunderland/projectID620c5a7c98ea1/projectID620c5a7c98ea1_images/001.png"
Try configuring viewer in Preferences dialog.

```

For `open -a Preview`:
```
18:59:38: ERROR - mainwindow.py:515 - Unable to execute "open -a Preview"
on file "/Users/sharonjoiner/Desktop/guiguts-testing/_projects_for_testing/blunderland/projectID620c5a7c98ea1/projectID620c5a7c98ea1_images/005.png"
Try configuring viewer in Preferences dialog.
```

Everything else I tried, even with another viewer, got a variation on the above errors. I was not able to get the external viewer working at all. Maybe @tangledhelix will have better luck -- I was pretty much just flailing.

tangledhelix:
# Linux

I'm not sure what to do on Linux exactly. There is an `open` command, but it doesn't work the same way Mac's does where you give it an application; you give it only a URL, like this:

```
open file:///home/ubuntu/DP_presidents/pngs/005.png
```

And then it has some sort of default-app registry, I assume, that determines which application is used to open that type of file. That may be good enough, at least for now.

The image viewer it's using on my Ubuntu VM is `eog` (Eye of GNOME). By default it opens a new window for each image. If we don't use `open` but instead call `eog` directly, we can give it a flag to re-use the window. That works, but unfortunately it's blocking, so GG freezes up until `eog` is closed. I can't find any flag to tell it not to block. So the situation on Linux is not great by default. Probably some different viewer would work better, but I don't know what's available (and since the built-in viewer will probably work well for most people, I'm not inclined to spend a bunch of time figuring it out unless someone inquires about it on Linux.)

# Mac

The approach you have won't work because the first thing in the list passed to `Popen()` is the executable it will call. Since there's no executable named `open -a Preview` it won't work as written. We'd need to do this instead:

```python
if is_mac():
    cmd = ["open", "-g", "-a", viewer_path, self.filename]
```

@srjfoo and I are still testing this. It's working as expected for me, but not for her. We aren't sure why that is, yet...

I tested this with Preview (included with the OS) and Pixea (3rd party commercial download).

tangledhelix:
I'm not sure whether you saw this already but I wanted to point it out. From the [subprocess documentation](https://docs.python.org/3/library/subprocess.html):

> The recommended approach to invoking subprocesses is to use the `run()` function for all use cases it can handle. For more advanced use cases, the underlying `Popen` interface can be used directly.

You may have looked at that and chosen `Popen` anyway, but FWIW, I tested with `run()` and it worked perfectly fine.

tangledhelix:
With the latest push, this is working on Mac (with one minor oddity noted below).

- External viewer navigates to the proper image. I tested this with Preview and Pixea.
- Internal viewer, if currently visible, navigates to the same image as external
- Unchecking the "Use External Viewer" box disables the external viewer

The nit: if you leave "Use External Viewer" checked, but erase the contents of the viewer app box, it will continue to use the already-known viewer, which makes sense in a way but is also a bit odd. If you close the Preferences pane and re-open it, although it clearly knows which app it's using to navigate, it has not re-populated the value. Perhaps setting this value to empty should auto-uncheck the "Use External Viewer" box?

I'll test on Linux and post separately about that.

tangledhelix:
> The nit: if you leave "Use External Viewer" checked, but erase the contents of the viewer app box, it will continue to use the already-known viewer

I just noticed your message in Slack, and I think I just misunderstood what was actually going on. I think it's actually working as designed. If I set it to Pixeam and it's using Pixea, then empty the field, it switches to Preview (which is the default app).

So on Mac this looks like it's all working.

tangledhelix:
# Linux

It "works" on Linux but this is probably the least optimal of the platforms. As I noted in my last message, calling with `open` will open a new window for each image. That is with the default viewer, anyway (`eog`). I see no way to pass arguments to `open` that get passed on to the viewer. One option is we could default to using `eog -w <file>` on Linux, but that's making some assumptions that I'm not sure are correct. I'm testing on one specific distribution of Linux, and who knows what people are using.

If you move forward from image to image (1 to 2 to 3, ...) it works, albeit opening new windows. If you reverse direction (3 to 2 to 1), it gets odd. Then new windows don't open, also the already-open windows don't pop to the top. Instead what it seems to do is give you a notice like this. I'm not sure what this means or whether it's normal.

Screenshot 2025-02-20 at 7 32 03 PM

I'm not a routine enough Linux desktop user to say whether this is typical behavior a Linux user expects. We may want to find a proper Linux tester to look at some of these things. In any case, it sort of works in Linux, sort of not. Personally I don't find this usable and I would use the built-in on Linux.

A different image viewer might work better. Perhaps we can ask the users if anyone has a suggestion.

srjfoo (approved):
Looks good to me.

windymilla:
> A different image viewer might work better. Perhaps we can ask the users if anyone has a suggestion.

My view is that we should do one of the following:
1. (My slight preference, given normal level of Linux user tech skills) Leave responsibility with the user to find a viewer that has config options such as "only open one window". XnView Classic has this, but is Windows-only. XnViewMP is available for Linux but I don't know if it has such features (\*_searches whole internet_\*... https://newsgroup.xnview.com/viewtopic.php?p=120622#p120622 - looks like it possibly does).
2. Add something to allow the user to specify an app with arguments (perhaps split contents of the app field at space characters, so we can pass them as separate elements in the list passed to run/popen). This would probably have to be platform-specific, at least non-Windows, since apps there regularly have space in their path. If you are able to check via a quick hack that changing line 512 of mainwindow.py to the following makes eog work, @tangledhelix, I'd feel more positive towards this option, if you prefer it. Thanks.
```
                    cmd = ["eog", "-w"]
```

srjfoo:
> 1. ... XnViewMP is available for Linux but I don't know if it has such features (*_searches whole internet_*... https://newsgroup.xnview.com/viewtopic.php?p=120622#p120622 - looks like it possibly does).

I haven't tried XnViewMP as an image viewer since we decided that, at least on a Mac, it did not work. Yes, you could specify it, but I have a vague memory of it always bringing up a separate window. Maybe we just didn't know what options to specify, though. It's a pretty complex program, not just image viewing, but image management as well as image display and editing. At one point, we had it as default viewer for GG1, and it just didn't work out as invoked.

windymilla:
> > 1. ... XnViewMP is available for Linux but I don't know if it has such features (__searches whole internet__... https://newsgroup.xnview.com/viewtopic.php?p=120622#p120622 - looks like it possibly does).
> 
> I haven't tried XnViewMP as an image viewer since we decided that, at least on a Mac, it did not work. Yes, you could specify it, but I have a vague memory of it always bringing up a separate window. Maybe we just didn't know what options to specify, though. It's a pretty complex program, not just image viewing, but image management as well as image display and editing. At one point, we had it as default viewer for GG1, and it just didn't work out as invoked.

That was my recollection too, though I don't use XnViewMP myself, which is why I tried to find the single window option. I think the new GG2 scheme is better, which is to use the OS default viewer unless the user specifies an app - it means we don't need to pick a default app for each OS.

tangledhelix:
> If you are able to check via a quick hack that changing line 512 of mainwindow.py to the following makes eog work, @tangledhelix, I'd feel more positive towards this option, if you prefer it. Thanks.
> ```
>                     cmd = ["eog", "-w"]
> ```

I didn't check just now but I did that test the other day. (You need to also pass the filename as the third argument, of course...)

tangledhelix (approved):

PR #782, opened 2025-02-20
Fix couple nits in PPtxt

### LINKED ISSUE #778: PPtxt: Minor suggestion - do not flag thought breaks
Currently `      *       *       *       *       *` is flagged under Character checks for `*`. Since this is standard practice for thought breaks, it should not be considered suspicious.

But not a big deal either way.

### LINKED ISSUE #779: PPtxt: Minor: Ampersands reported in two different checks
They are reported under "Character checks" as '&', then again under "Ampersand character in line".
Suggest excluding it from "Character checks".

`'&' (2)
4028.70: “Yes, when the house of Fausset was in its infancy the house of Felix &
4041.13: Lyons. Felix & Sons had grown old-fashioned. They had forgotten to

...

Ampersand character in line (excluding '&c.' and unicode numeric character references).
4028.70: “Yes, when the house of Fausset was in its infancy the house of Felix &
4041.13: Lyons. Felix & Sons had grown old-fashioned. They had forgotten to
`

windymilla:
1. Don't report ampersand in character checks because there's a better check later in "Ampersand character in line". Fixes #779 
2. Don't report asterisk in `       *       *       *       *       *` in character checks - it's a standard thought break. Fixes #778

### DIFF
--- a/src/guiguts/tools/pptxt.py
+++ b/src/guiguts/tools/pptxt.py
@@ -612,7 +612,9 @@ def weird_characters() -> None:
         # Update dictionary with the weirdos from the line.
         for weirdo in weirdos_list:
             # Skip exceptions
-            if weirdo == "*" and re.fullmatch(r"( {3,10}\*)\1\1\1\1", line):
+            # Thoughtbreak consists of 5 asterisks equally spaced,
+            # and indented by at least 3 spaces.
+            if weirdo == "*" and re.fullmatch(r" {3,}\*( {3,10}\*)\1\1\1", line):
                 continue
             if weirdo in weirdos_lines_dictionary:
                 weirdos_lines_dictionary[weirdo].append(line_number)
--- a/src/guiguts/tools/pptxt.py
+++ b/src/guiguts/tools/pptxt.py
@@ -608,9 +608,12 @@ def weird_characters() -> None:
     line_number = 1
     for line in book:
         # Get list of weirdos on this line. That means any character NOT in the regex.
-        weirdos_list = re.findall(r"[^A-Za-z0-9\s.,:;?!\\\-_—–=“”‘’\[\]\(\){}]", line)
+        weirdos_list = re.findall(r"[^A-Za-z0-9\s.,:;?!&\\\-_—–=“”‘’\[\]\(\){}]", line)
         # Update dictionary with the weirdos from the line.
         for weirdo in weirdos_list:
+            # Skip exceptions
+            if weirdo == "*" and line == "       *" * 5:
+                continue
             if weirdo in weirdos_lines_dictionary:
                 weirdos_lines_dictionary[weirdo].append(line_number)
             else:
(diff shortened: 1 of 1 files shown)


src/guiguts/tools/pptxt.py
```diff
@@ -608,9 +608,12 @@ def weird_characters() -> None:
     line_number = 1
     for line in book:
         # Get list of weirdos on this line. That means any character NOT in the regex.
-        weirdos_list = re.findall(r"[^A-Za-z0-9\s.,:;?!\\\-_—–=“”‘’\[\]\(\){}]", line)
+        weirdos_list = re.findall(r"[^A-Za-z0-9\s.,:;?!&\\\-_—–=“”‘’\[\]\(\){}]", line)
         # Update dictionary with the weirdos from the line.
         for weirdo in weirdos_list:
+            # Skip exceptions
+            if weirdo == "*" and line == "       *" * 5:
```
rtonsing, line 615: I think it would be better to use a regex, and not assume the # of spaces is always the same.

windymilla: Except this way it would warn you if your tb wasn't standard, eg had some spaces missing.

windymilla: What kind of variation were you thinking of?

rtonsing: Probably OK as-is. The spacing between should at least be consistent, but I'm not sure how to specify that in regex.

windymilla: I can do that in a regex, if you specify what spacings you want to allow

rtonsing: GG uses 7 spaces on tb conversion, maybe 3 to 10?

rtonsing (changes requested):
Oops - thought break is still reported in character checks if it is centered, which I and others do. 19 leading spaces in my current case, but may vary according to set margins.

windymilla:
> Oops - thought break is still reported in character checks if it is centered, which I and others do. 19 leading spaces in my current case, but may vary according to set margins.

Further commit should now permit that. Essentially it will not report if there are 5 equally spaced asterisks which can be indented by any number of spaces greater than or equal to 3 to allow for centering or additional indentation due to blockquotes etc. The spacing between asterisks can be from 3 to 10 spaces. 
So the minimum width is:
```
   *   *   *   *   *
```
and the maximum is 
```
     <any number of spaces>    *          *          *          *          *
```
Hope that covers all the necessary cases.

rtonsing:
Still reported for centered cases.

Did a fresh clone and "gh pr checkout 782"

[tb_test.txt](https://github.com/user-attachments/files/18913401/tb_test.txt)

windymilla:
> Still reported for centered cases.
> 
> Did a fresh clone and "gh pr checkout 782"
> 
> [tb_test.txt](https://github.com/user-attachments/files/18913401/tb_test.txt)

Oops - forgot to push my change to Github. Sorry. Should be there now if you refresh your branch

rtonsing (approved):
OK now.

PR #787, opened 2025-02-23
Add link to PPWB

### LINKED ISSUE #451: Add CSS Validator
Is this necessary, or is it now covered by the Nu HTML checker?

windymilla:
HTML5 and CSS validators can be opened via options in the HTML menu. These are marked "(www)" to indicate that they will open a browser window. Also added "(www)" to a couple of Help menu options.

Fixes #451

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -853,6 +853,11 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         html_menu.add_button("HTML5 ~Validator (online)...", html_validator_check)
         html_menu.add_button("~CSS Validator (online)...", css_validator_check)
         html_menu.add_button("HTML ~Link Checker...", html_link_check)
+        html_menu.add_separator()
+        html_menu.add_button(
+            "~PP Workbench (www)",
+            lambda: webbrowser.open("https://www.pgdp.net/ppwb/"),
+        )
 
     def init_view_menu(self) -> None:
         """Create the View menu."""
@@ -898,10 +903,10 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         """Create the Help menu."""
         help_menu = Menu(menubar(), "~Help")
         help_menu.add_button(
-            "Guiguts ~Manual", self.show_help_manual, "F1", force_main_only=True
+            "Guiguts ~Manual (www)", self.show_help_manual, "F1", force_main_only=True
         )
         help_menu.add_button("About ~Guiguts", self.help_about)
-        help_menu.add_button("~Regex Quick Reference", self.show_help_regex)
+        help_menu.add_button("~Regex Quick Reference (www)", self.show_help_regex)
         help_menu.add_button(
             "List of ~Compose Sequences", ComposeHelpDialog.show_dialog
         )
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -849,10 +849,11 @@ Fifth Floor, Boston, MA 02110-1301 USA."""
         html_menu = Menu(menubar(), "HT~ML")
         html_menu.add_button("HTML ~Generator...", HTMLGeneratorDialog.show_dialog)
         html_menu.add_button("Auto-~Illustrations...", HTMLImageDialog.show_dialog)
+        html_menu.add_separator()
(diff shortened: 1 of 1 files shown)


src/guiguts/application.py
```diff
@@ -836,6 +836,17 @@ def init_html_menu(self) -> None:
         html_menu.add_button("~Unmatched HTML Tags...", unmatched_html_markup)
         html_menu.add_button("HTML5 ~Validator (online)...", html_validator_check)
```
srjfoo, line 837: Remove this line? I think the HTML validator was the first one you added, and hadn't standardized yet on the (www) suffix?

image

Otherwise, they both worked for me!

I didn't know whether the open file would automatically be uploaded when the option was chosen, or whether it would require the user to navigate to the file. Since it's the latter (i.e. further input is required, unlike with the help options), perhaps the `...` is appropriate to indicate that?

tangledhelix: The first one is the one that submits the HTML file for validation via the API (built-in validation for GG). The second one just opens the W3C page in your browser, like the CSS validator link.

I almost think the first link might be renamed to be more clear. It's not clear from that menu item what it does compared to the (www) link, perhaps. Maybe it could be re-labeled something like "Submit to HTML5 validator" or something along those lines instead...

tangledhelix: Or "Upload to HTML validator" to be really clear?

srjfoo: Maybe "Validate current HTML (online)".

I guess I can sort of see a reason to have both, but I definitely was confused. I assume the CSS validation doesn't have an API component, since there's not a corresponding menu item for validating the current CSS?

tangledhelix: The W3C CSS service does have an API available, although we haven't explored it yet. It's a different style (SOAP/XML) than the Nu validator's API (REST/JSON). It wasn't discussed in #451 but I recall it came up in Slack at some point.

I'm not averse to looking at it, but I wouldn't say it's a priority either.

tangledhelix: Added #789 to look into that at some point.

tangledhelix (approved):

srjfoo (approved):
Now that I better understand what's going on, looks good.

src/guiguts/application.py
windymilla: Maybe we don't need these two "www" buttons at all, especially if the SOAP/XML method works OK?


I originally said we didn't need the CSS validator, since the Nu HTML validator also reports errors in the CSS. But @tangledhelix thought it could be useful in order to be able to control whether validating to CSS3 or CSS2.1. So then I thought I'd add the link to the www version, and while I was doing it thought we might as well have a link to the HTML validator too. Then it was confusing to have two links to the validator, so maybe we don't need the www versions.
("And that's everything that happened in my life right up to the time I got this feedback" - Grandpa Simpson)

tangledhelix: If one can validate "in-app" then IMO there's no reason to have the links.

Actually, maybe the better plan is (now that we seem to be moving to have both submit-via-API tools), drop the links in this PR and instead just have a menu item to go to the PPWB. The PPWB main page links to the HTML and CSS validators, plus PG's ebookmaker, in addition to offerings its own tools.

windymilla:
Now, just adds link to PPWB instead of validators.

tangledhelix (approved):

srjfoo (approved):
It's fine as-is, but if I might make an organizational suggestion -- reorganize the HTML menu to keep all the internal GG2 checks in a block, and put a divider below that for the online CSS and HTML checks that bring up the W3C validators. Also, if they rate the `...`, should the link to ppwb have it, too?

windymilla:
> It's fine as-is, but if I might make an organizational suggestion -- reorganize the HTML menu to keep all the internal GG2 checks in a block, and put a divider below that for the online CSS and HTML checks that bring up the W3C validators. Also, if they rate the `...`, should the link to ppwb have it, too?

I have reorganized a little, but not exactly as you suggested. There are now 3 sections: generating HTML (2 entries); checks (4 entries); external websites (1 entry). If you strongly object, there will be further opportunities to reorganize, for example when completing #795 (prompted by your query about the ellipsis in the link to ppwb).

PR #790, opened 2025-02-24
Add auto-fit-to-(height, width) in image viewer

### LINKED ISSUE #788: Image viewer: sticky auto-size modes
Add ability in image viewer to activate auto-sizing modes:
- Fit image to viewer width
- Fit image to viewer height

Obviously these modes are mutually exclusive, both cannot be active at the same time. But neither is required, i.e. both may be off (which will be the default).

It seems reasonable that many PPers may have a preference for one of these views as their own default (including myself), so make the mode sticky, stored in the prefs file.

It also seems reasonable a PPer may find both modes useful, as well as deactivating both, so it should be as uncomplicated as possible to change the mode.

tangledhelix:
Adds two checkboxes to image viewer, to automatically fit the image to the viewport either by height or by width (of the viewport). The modes are mutually exclusive; only one may be active at a time. However, neither is required so they may both be inactive.

Operation is very simple: check the checkbox and the mode starts working; uncheck to turn off. If you check one and the other was checked, it will be automatically unchecked (they are mutually exclusive).

When either mode is turned on, the resize happens immediately. As image navigation happens, resize happens on load of the next/prev image.

If zooming an image manually, the manual zoom is kept until the viewer moves to a different image. If neither mode is active, the manual zoom level is left in place as you move from image to image.

The mode is sticky between restarts, stored in the user preferences.

Fixes #788

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -31,6 +31,8 @@ from guiguts.mainwindow import (
     process_accel,
     mainimage,
     AutoImageState,
+    image_autofit_width_callback,
+    image_autofit_height_callback,
 )
 from guiguts.misc_dialogs import (
     PreferencesDialog,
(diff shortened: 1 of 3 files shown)


tangledhelix:
> Suggestion: Remove the label "Fit" completely and let the button icons speak for themselves. Then set `padx=(10,0)` on the 4 checkboxes.

I did something a bit different than that, but took the advice on padding. I removed the separate "Fit" label and built it into the two checkbox labels instead.

Screenshot 2025-02-24 at 9 17 40 PM

For clarity of the entire control system, I added tooltips. After discovering the `use_pointer_pos` argument, it actually works when undocked, unlike the last time I tried tooltips. 🎉 

Screenshot 2025-02-24 at 9 18 27 PM

srjfoo (approved):

windymilla (approved):
LGTM thanks

PR #791, opened 2025-02-24
Add CSS Validator to HTML menu

windymilla:
Use SOAP/XML interface to validator and display
results in a checker dialog.

Allow user to select CSS 2.1 or CSS 3 & store in prefs.


windymilla:
Testing notes: if you wan t something that is valid in CSS3 but not in 2.1, you could try adding
`border-radius: 5px;` to the CSS for the `p` element (around line 25 in a standard GG-generated HTML file.

tangledhelix (approved):

tangledhelix:
Looks good! I wish their API returned the column as well as line, but so be it.

srjfoo (approved):
Very slick!

PR #797, opened 2025-02-26
Don't re-show image viewer when closed

windymilla:
If AutoImg is on, and the External Viewer is turned off, then closing the image viewer just made it reappear again.

Now, if External Viewer is off, when image viewer is closed, we turn off AutoImg.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -1067,7 +1067,12 @@ class MainWindow:
 
     def hide_image(self) -> None:
         """Stop showing the current image."""
-        # preferences.set(PrefKey.AUTO_IMAGE, False)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
1. Turn off External Image Viewer in Prefs
2. Turn on AutoImg - internal viewer should show the current image
3. Hide the internal viewer (using the cross or View-->Hide Image Viewer)
In `master` the viewer reappears and AutoImage remains on.
In this branch, the viewer does not reappear, and AutoImage is turned off.

The above should be the case whether image viewer is docked or floating.

If External Image is turned on, then there is no change in behavior - if you hide the internal viewer, AutoImage remains on, and the internal viewer remains hidden until you Show it again.

srjfoo (approved):

PR #798, opened 2025-02-28
Match the OS-level dark/light theme on all platforms

### LINKED ISSUE #785: Improve OS light/dark detection
As discussed in Slack, [darkdetect](https://github.com/albertosottile/darkdetect) seems to work on all platforms to detect whether the OS is in light or dark mode. Change the theme selection algorithm like so:

- Rename the GG "Default" theme to "Auto"
- The Tk theme(s) behind "Default" such as `aqua` or `vista` would no longer be used
- Options for users to choose from: "Auto", "Light", "Dark"
- Base default for new users: "Auto"
- If the preference system detects an invalid theme name, switch to "Auto" automatically
- "Light" theme maps to Tk theme "awlight"
- "Dark" theme maps to Tk theme "awdark"
- "Auto" chooses among "awlight", "awdark" automatically based on the current OS light/dark mode, as reported by `darkdetect`
- In a timer routine similar to `auto_image_check`, periodically check for a theme update:
    - If the GG theme is not "Auto", exit; there's nothing to do
    - Check the current report of `darkdetect`
    - Gather the current GG theme name
    - If the two don't match, change the GG theme to the current OS theme

Doing this in the global `after_idle` redraw routine is probably too aggressive. Use an independent checker similar to `auto_image_check` and a specific timer, which need not be super up-to-date. 1000ms could be a good starting point, maybe even 1500-2000?

tangledhelix:
On all supported platforms, when Guiguts theme is set to "Default," detect the OS-level light/dark mode and dynamically switch Guiguts to match the OS.

Discontinue use of the Mac/Windows-styled "Default" theme. "Default" now means to match the OS light or dark, using the themes "Light" or "Dark".

Reduced contrast for background, text, and input-line highlight slight in the default themes. All other colors remain as they were.

Promoted the Image Viewer "High Contrast" mode to be a global preference, now applied to both the text area as well as images. This means there are now four possible theme choices: Light, Dark, and high-contrast variations of both.

Preference `IMAGE_VIEWER_HI_CONTRAST` replaced with `HIGH_CONTRAST`. This preference has moved to the top section, no longer Image Viewer-specific.

(Unrelated to dark/light mode): Fixed a mypy error on Mac related to the `idlelib.redirector` import.

Fixes #785


windymilla (approved):
The "blueness" was a bit of a shock to me too, but I'm happy for the reasons you said for us to let it go to users and see whether this theme (with the hi-contrast option) is OK for people generally.

Just one suggestion: move the Hi Contrast checkbox in the prefs dialog up, so that it sits directly below the theme selection, since the two belong together really.

srjfoo (approved):

tangledhelix:
> The "blueness" was a bit of a shock to me too, but I'm happy for the reasons you said for us to let it go to users and see whether this theme (with the hi-contrast option) is OK for people generally.

It's been changed to a pure gray of a similar brightness based on this comment and also conversation with @srjfoo.

> Just one suggestion: move the Hi Contrast checkbox in the prefs dialog up, so that it sits directly below the theme selection, since the two belong together really.

Done; rebased, squashed, pushed.

windymilla:
@tangledhelix I don't see any change in blueness - it's possible we were talking at cross-purposes, or I've not ended up refreshing my local version of the branch properly. I meant that in dark mode, the background is quite blue. As I said, if you think it's better, I'm happy to give it a try, but I thought I should check that I wasn't missing something that you had intended to change to "a puer gray".

srjfoo:
> @tangledhelix I don't see any change in blueness - it's possible we were talking at cross-purposes, or I've not ended up refreshing my local version of the branch properly. I meant that in dark mode, the background is quite blue. As I said, if you think it's better, I'm happy to give it a try, but I thought I should check that I wasn't missing something that you had intended to change to "a pure gray".

What I was talking about was specifically removing the blue from the text, so that's all I looked at when I was playing around. If you think the background is too blue, I can look at that, too.

windymilla:
No, happy to go with that if that was intended - I misunderstood. If all the users say "too blue", we'll reconsider :)

PR #803, opened 2025-02-28
Change Help About Guiguts dialog to text widget

### LINKED ISSUE #802: Change Help-->About GG to be a read-only text
Noticed during #799, it would be easier for the user to report version numbers if the Help About box was a read-only text, rather than a label.

windymilla:
In order to allow user to easily copy version information to report a bug, a label wasn't really suitable. Text widget allows them to
1. Use the button provided!
2. Click & drag to select, then use Ctrl-C to copy
3. Right click to use context menu to select/copy

Fixes #802

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -7,9 +7,6 @@ import logging
 import importlib.resources
 from importlib.metadata import version
 import os.path
-import platform
-import sys
-from tkinter import messagebox
 from typing import Optional
 import unicodedata
 import webbrowser
@@ -45,6 +42,7 @@ from guiguts.misc_dialogs import (
     ComposeHelpDialog,
     UnicodeBlockDialog,
     UnicodeSearchDialog,
+    HelpAboutDialog,
 )
 from guiguts.misc_tools import (
     basic_fixup_check,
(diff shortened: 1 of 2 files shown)


rtonsing (approved):

srjfoo (approved):

windymilla:
When Tk copies text to the clipboard, it doesn't update
macOS's clipboard counter. Some apps use this counter
to detect when the clipboard has changed, so don't
realize there is new text there.

The fix is to use macOS `pbcopy` command to "touch" the
clipboard by non-Tk means.

This was previously implemented for copy/cut within the
main text window. However, the Copy Results buttons in
checker dialogs and WF, and the Copy Version Info in
Help About Guiguts didn't have the fix.

Code altered to override the default clipboard append
routine for maintext, and also to use maintext's version
for the other cases.

Hopefully this means that anyone in the future looking
how to copy to the clipboard will use the new functions,
and thus circumvent the bug.

Ready for retesting please, @srjfoo, in the following areas, each time checking that the cut/copied text actually ends up in the clipboard buffer:
1. Normal Copy/Cut from maintext or peer window.
2. Same, but column Copy/Cut
3. Copy Results in any one checker dialog (e.g. Find All)
4. Same in WF dialog
5. Version Copy in Help About dialog.
Thanks

srjfoo:
> * Normal Copy/Cut from maintext or peer window.
-- works fine

> * Same, but column Copy/Cut
-- works fine

> * Copy Results in any one checker dialog (e.g. Find All)
-- Spelling -- drag to select isn't an option at all. Paste works fine with copy button.
-- Bookloupe, ditto -- no drag to select, paste works fine with copy button.
-- Find all, ditto.

> * Same in WF dialog
-- Unpredictable if trying to use drag-to-select; reliable if using copy button.

> * Version Copy in Help About dialog.
> -- Unpredictable if trying to use drag-to-select; reliable if using copy button (I think).

So, it seems that in the widgets, the copy button will allow paste just fine (I haven't checked `master` yet), but trying to drag-to-select and then copy (where it's an option) is wonky and unreliable.

srjfoo (approved):
It looks like selecting and copying from the help-about text and pasting into BBEdit now works as expected. One thing I noticed with this round of testing -- WF no longer seems to allow drag-to-select. Just verifying that that was intended.

PR #805, opened 2025-03-01
Track image auto-fit with change in viewer size

### LINKED ISSUE #804: Image auto-fit doesn't redraw if image viewer resizes
When auto-fit mode is on, auto-fit is redrawn when the image changes to a different image, but not when the viewer changes its size:

- Drag Image Viewer pane boundary between image and text
- Resize application main window
- Dock / undock of Image Viewer
- Resize undocked Image Viewer window

tangledhelix:
Bug-fix: when auto-fit modes are on, and the height or width of the viewer changes (whichever is relevant), the image was not redrawn until a new image was loaded.

Testing notes:

Using image viewer, either docked, or undocked, turn on an auto-fit mode and then resize the viewer. That can be one of:
- resize undocked window
- resize main window with docked viewer
- resize image viewer using the divider between it and the text pane(s)

It should dynamically resize to match the mode you've chosen (either by height or by width).

If the image changes (using either status-bar `<` `>` buttons or `Auto Image`) then it should also pop to the auto-fit size. If you manually zoom in/out it should remain as you zoomed it unless you either resize the window, or change to a different image.

If you navigate to a different image with the image viewer `<` `>` buttons and then leave the viewer with Auto Image on, then the zoom will reset as the image resets.

Fixes #804

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -7,7 +7,7 @@ import os.path
 import subprocess
 import time
 import tkinter as tk
-from tkinter import ttk, messagebox
+from tkinter import ttk, messagebox, EventType
 from typing import Any, Callable, Optional
 
 from PIL import Image, ImageTk, ImageChops
@@ -361,10 +361,10 @@ class MainImage(tk.Frame):
             _, ce = process_accel("Cmd/Ctrl+equal")
             _, cm = process_accel("Cmd/Ctrl+minus")
             _, c0 = process_accel("Cmd/Ctrl+0")
-            widget.bind(cp, lambda _: self.zoom_in_btn.invoke())
-            widget.bind(ce, lambda _: self.zoom_in_btn.invoke())
-            widget.bind(cm, lambda _: self.zoom_out_btn.invoke())
-            widget.bind(c0, lambda _: self.fth_btn.invoke())
+            widget.bind(cp, lambda _: self.image_zoom(zoom_in=True))
+            widget.bind(ce, lambda _: self.image_zoom(zoom_in=True))
+            widget.bind(cm, lambda _: self.image_zoom(zoom_in=False))
+            widget.bind(c0, lambda _: self.image_zoom_to_height(disable_autofit=True))
 
         self.hbar = ttk.Scrollbar(top_frame, orient=tk.HORIZONTAL)
         self.hbar.grid(row=3, column=0, sticky="EW")
@@ -447,6 +447,10 @@ class MainImage(tk.Frame):
         Args:
             zoom_in: True to zoom in, False to zoom out.
         """
+        # If user zooms manually, turn off the autofit modes to avoid
+        # unexpected weird behaviors and confusion.
+        preferences.set(PrefKey.IMAGE_AUTOFIT_WIDTH, False)
+        preferences.set(PrefKey.IMAGE_AUTOFIT_HEIGHT, False)
         if zoom_in:
             if self.image_scale < 3:
                 self.image_scale *= self.scale_delta
@@ -466,10 +470,13 @@ class MainImage(tk.Frame):
         ) / (bbox_image[2] - bbox_image[0])
         self.image_zoom_by_factor(scale_factor)
 
-    def image_zoom_to_height(self) -> None:
+    def image_zoom_to_height(self, disable_autofit: bool = False) -> None:
         """Zoom image to fit to height of image window."""
         if not self.imageid:
             return
+        if disable_autofit:
+            preferences.set(PrefKey.IMAGE_AUTOFIT_WIDTH, False)
+            preferences.set(PrefKey.IMAGE_AUTOFIT_HEIGHT, False)
         bbox_image = self.canvas.bbox(self.imageid)
         scale_factor = (
             self.canvas.canvasy(self.canvas.winfo_height()) - self.canvas.canvasy(0)
@@ -484,8 +491,6 @@ class MainImage(tk.Frame):
         """
         self.image_scale *= scale_factor
         preferences.set(PrefKey.IMAGE_SCALE_FACTOR, self.image_scale)
-        self.canvas.xview_moveto(0.0)
-        self.canvas.yview_moveto(0.0)
         self.show_image()
 
     def show_image(self, internal_only: bool = False) -> None:
@@ -658,7 +663,7 @@ class MainImage(tk.Frame):
             pass
         self.allow_geometry_storage = True
 
-    def handle_configure(self, _e: tk.Event) -> None:
+    def handle_configure(self, evt: Optional[tk.Event]) -> None:
         """Handle configure event."""
         if preferences.get(PrefKey.IMAGE_WINDOW_DOCKED):
             try:  # In case unlucky timing means it tries to configure during undocking & finds sash doesn't exist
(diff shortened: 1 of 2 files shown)


src/guiguts/mainwindow.py
```diff
@@ -672,6 +672,11 @@ def handle_configure(self, _e: tk.Event) -> None:
                 preferences.set(PrefKey.IMAGE_FLOAT_GEOMETRY, tk.Wm.geometry(self))  # type: ignore[call-overload]
             except tk.TclError:
                 pass
+        if _e.type == EventType.Configure:
```
windymilla, line 675: I love the way it resizes the image as I resize either the docked or floating image viewer now - big improvement!

However, I'm afraid this change has fallen into a trap I (not so) carefully laid - sorry. The argument `_e` has the underscore to signify that it isn't used (recognised by pylint as such), and in `file.py::check_save()` line 363, I took advantage of that by passing a minimally constructed `tk.Event()` into a call to `handle_configure` to force a save of the current image viewer geometry.

So, with your change, if you load a file, make a change, then try to exit the program, when it calls `check_save` to see if you want to save the file before exiting, it calls `handle_configure` with a tk.Event that doesn't have a `type` attribute, and causes an exception.

Suggestions:
1. (Best plan if it works) I tried just deleting the call to `handle_configure` in `check_save` on the basis that it may have been necessary when I worked on image viewer previously, but probably wasn't with more recent improvements. It seemed to be OK, but if you go with that fix, you'd just want to check that image viewer geometry is being saved properly.
3. (What I should have done previously) Make the event argument type `Optional[tk.Event]`, which then permits you to pass `None` instead of the badly formed `tk.Event()` when calling from `check_save`. In `handle_configure`, where you check the event's `type` attribute, you would then have:
```
if _e is not None  and _e.type=EventType.Configure:
```

Whichever of the above is done, the argument `_e` should definitely be renamed, e.g. to `evt` (or something without an underscore) since it is no longer an unused variable.
image

windymilla (changes requested):

src/guiguts/mainwindow.py
tangledhelix: > 1. (Best plan if it works) I tried just deleting the call to `handle_configure` in `check_save` on the basis that it may have been necessary when I worked on image viewer previously, but probably wasn't with more recent improvements.

It looks like I need to check a handful of things, and some of them I'm confused about, actually. For example, looking at what `handle_configure` does (before my change), its purpose is clearly to save the geometry of the viewer. But in that event (har har), why is it invoked on `<Leave>` and `<Enter>`? Neither of those change the window geometry; if the geometry changed then `<Configure>` should have fired and been handled already. Do you remember whether those other events need this code or was it more of a belt-and-suspenders thing?

tangledhelix: Removing the handler from `check_save` broke the geometry save, as far as I can tell, so I went with option 2.

There's no issue or exception with this callback firing on `<Leave>` and `<Enter>`, although I can't quite understand why it's being called. I left those as-is since they were already there and there's no clear reason to remove it.

tangledhelix: It seems they were added in 930ab3d5ac1e49364cd4723eea005e189ac19e0e so I'll assume they are of use and should remain. So I think this is ready for final review.

srjfoo:
Not entirely sure if I've duplicated @windymilla's steps to try to reproduce the exception, but I haven't been able to trigger one.

Question -- when I resize, no matter where I am on the page, the image is displayed at the top of the page, not centered on where it used to be. That is, 

image

 becomes

image

It's not the end of the world if the answer is "so be it", but I thought I'd mention it.

srjfoo (approved):

tangledhelix:
> Not entirely sure if I've duplicated @windymilla's steps to try to reproduce the exception, but I haven't been able to trigger one.

The exception only triggered when trying to quit the app (or probably close an unsaved file). But is fixed now.

> Question -- when I resize, no matter where I am on the page, the image is displayed at the top of the page, not centered on where it used to be. That is,

Just pushed new code to fix that. Good catch, I hadn't noticed that.

srjfoo:
Okay, another question (or group of questions -- and we may have already discussed this, and I just wasn't listening very closely) -- "fit to X" is on, the user zooms in on the image. Is the assumption that if the PPer re-sizes the image pane, they want it to go back to its pre-zoom state? Because that's what it's doing.

I can understand it going back to "fit to X" upon changing to a new image, but if they've zoomed in on "this" image, and decide to make the viewport larger so they can see more of it, chances are they want to keep that same zoom -- for "this" image. Is there a way to turn off autofit for "this" image if zoom (either button or keyboard shortcut) is used, and revert to autofit when the image is changed? Is that even a good idea?

tangledhelix:
> Okay, another question (or group of questions -- and we may have already discussed this, and I just wasn't listening very closely) -- "fit to X" is on, the user zooms in on the image. Is the assumption that if the PPer re-sizes the image pane, they want it to go back to its pre-zoom state? Because that's what it's doing.

It is always hard to guess what people will desire, but I can certainly imagine that scenario and that a PPer would be annoyed by it.

> I can understand it going back to "fit to X" upon changing to a new image, but if they've zoomed in on "this" image, and decide to make the viewport larger so they can see more of it, chances are they want to keep that same zoom -- for "this" image. Is there a way to turn off autofit for "this" image if zoom (either button or keyboard shortcut) is used, and revert to autofit when the image is changed? Is that even a good idea?

I think that implementing it, with some more tracking of state, should be possible. I wonder what the user expectation would be. This is one approach; another might be that if the user zooms, we simply turn autofit modes off until they turn one back on. Which behavior is desirable might depend on the person, but having autofit turn off if you manually zoom is possibly the best fit to the "principle of least surprise".

What do you think? @windymilla any thoughts about it?

srjfoo:
> I think that implementing it, with some more tracking of state, should be possible. I wonder what the user expectation would be. This is one approach; another might be that if the user zooms, we simply turn autofit modes off until they turn one back on. Which behavior is desirable might depend on the person, but having autofit turn off if you manually zoom is possibly the best fit to the "principle of least surprise".

I thought of just turning it off, too, but didn't know what y'all would think best. These are the fiddly bits. We don't need to make it a full-featured viewer, just grind most of the rough edges off I think? It's quite usable the way it is!

windymilla (approved):

windymilla:
I don't think it's necessary to store the zoom-to-fit state for each image - if you were going to do that, then you might as well store the zoom factor per image, but I think it's overkill for what is needed. As we've said before, the majority of projects have the majority of scans more or less identical sizes, so the current "keep the same scale factor" works for those. 

If y'all are happy, I'll merge this - I'll wait to hear.

tangledhelix:
> I don't think it's necessary to store the zoom-to-fit state for each image

I don't think that's what she was suggesting; just storing (not persistently) whether the current image had been manually zoomed, and in that case, don't re-autofit if the window geometry changes; but resuming autofit if a different image loads.

I'm not sure how confusing I want the behavior to become. I think rather than track that state and try to get too clever, it might be more intuitive for the user if manual zoom simply turns off the auto-fit mode (if one was active) and leave it at that. As you said, most images are about the same scale within a project anyway so it's entirely possible most users won't use autofit in the first place.

tangledhelix:
Latest push includes the following:
- If user zooms manually, autofit modes are turned off
- Zoom can mean +/- buttons, or the zoom in/out hotkeys, or the cmd/ctrl-zero (fit-to-height) hotkey.
- Changed hotkey bindings to call underlying functions rather than `{button}.invoke()` because behavior was broken otherwise, at least for cmd/ctrl-zero

srjfoo (approved):
Excellent!

src/guiguts/mainwindow.py
tangledhelix: I removed those calls. I agree they had no purpose, and my testing indicates the resizing still works, as does the on-leave orange-blink hightlight border.

srjfoo:
Did another bit of testing to make sure the latest commit didn't break anything. It looks really good!

PR #808, opened 2025-03-04
Select lines by dragging over line-number gutter

tangledhelix:
Clicking and releasing on the line number gutter selects the clicked line. Clicking and dragging in either direction selects the range of lines.

Caveat: due to the implementation of the line gutter, currently dragging beyond the visible range of lines is not supported. It may be possible with further work, but I didn't attempt it.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -212,6 +212,11 @@ class TextLineNumbers(tk.Canvas):
         self.bind("<<ThemeChanged>>", lambda event: self.theme_change())
         self.text_color = themed_style().lookup("TButton", "foreground")
 
+        # Drag along the line number gutter to select lines
+        self.drag_begin_index = IndexRowCol(1, 0)
+        self.bind("<Button-1>", self.drag_handler_start)
+        self.bind("<B1-Motion>", self.drag_handler)
+
     def redraw(self) -> None:
         """Redraw line numbers."""
         # Allow for 5 digit line numbers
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

src/guiguts/maintext.py
```diff
@@ -260,6 +265,53 @@ def theme_change(self) -> None:
...
+            # we were dragging downward
+            self.textwidget.mark_set(tk.INSERT, f"{cursor.index()} lineend")
+            self.textwidget.mark_set(TK_ANCHOR_MARK, _range.start.index())
+        elif self.drag_begin_index.row > cursor.row:
+            # we were dragging upward
+            self.textwidget.mark_set(tk.INSERT, _range.start.index())
+            self.textwidget.mark_set(TK_ANCHOR_MARK, f"{_range.end.row}.0")
+        else:
+            # clicked only a single line
+            self.textwidget.mark_set(tk.INSERT, f"{_range.end.index()}-1c")
```
windymilla, line 312: Similarly, avoid the insert position being at the end of the single selected line (which isn't a "whole line" select) and putting it at the start of the next line, this should be
```python
self.textwidget.mark_set(tk.INSERT, f"{_range.end.index()}")
```

windymilla (commented):
With a couple of tiny tweaks to the postion of `tk.INSERT` to put it at the start of the next line instead of the end of the preceding line, I think this is great. With those two changes, it seems entirely self-consistent to me, whether you click & drag several whole lines, or just click in the line number to select 1 line. If you then do shift+up/down arrow, it continues to behave as expected, selecting/deselecting whole lines.

src/guiguts/maintext.py
```diff
@@ -260,6 +265,53 @@ def theme_change(self) -> None:
...
+            IndexRowCol(sorted_rows[0], 0), IndexRowCol(sorted_rows[1] + 1, 0)
+        )
+
+    def set_insert_index(self, evt: tk.Event) -> None:
+        """Set the insert index so the line number is highlighted properly."""
+        cursor = self.cursor_index(evt)
+        _range = self.drag_range(evt)
+        if self.drag_begin_index.row < cursor.row:
+            # we were dragging downward
+            self.textwidget.mark_set(tk.INSERT, f"{cursor.index()} lineend")
```
windymilla, line 304: My comment here got lost first time - probably failed to click the green button :)

I think this line should be
```python
self.textwidget.mark_set(tk.INSERT, f"{cursor.index()} +1l linestart")
```

src/guiguts/maintext.py
tangledhelix: Made this change (but without the string embed)

src/guiguts/maintext.py
tangledhelix: Made this change

tangledhelix:
Made those changes, and seems to be "right" as you say. Feel free to re-review.

srjfoo (approved):

windymilla:
All looks good to me now, thanks!

PR #809, opened 2025-03-05
Implement Shift-Click/Drag in line number gutter

windymilla:
Convert Shift-click/drag in line number gutter to a shift/click/drag in the first column of the current text line, since that already does what we want.


windymilla:
Since Dan had already done all the hard work, and after our discussions yesterday, it occurred to me that we wanted a Shift click/drag in the line number gutter to behave exactly like a Shift click/drag in the first column of the text window, which causes it to extend/reduce the existing selection. So, a couple of lines of code just generate an event in the text window instead of the line gutter.

tangledhelix (approved):
Seems to work right, and what it does is intuitive... nice!

srjfoo (approved):

srjfoo:
Looks good to me, too.

PR #818, opened 2025-03-08
Increase Mac line spacing in checkers

windymilla:
Match line spacing in maintext since they are the same type of widget.


srjfoo (approved):
Thanks, that does make dense text somewhat easier to read -- at least on a Mac 😁

PR #819, opened 2025-03-09
Add Advanced tab for Preferences

windymilla:
1. Convert frames to tabbed notebook widget
2. Remember which tab user was looking at previously
3. Allow Ctrl-tab navigation between tabs (may need to be disabled if macOS bug is triggered)
4. Add line spacing pref to advanced tab
5. Add cursor width pref to advanced tab


windymilla:
Testing notes: When we attempted to use a tabbed notebook previously, on macOS when you used the keyboard to switch between tabs (Cmd-tab & Shift-Cmd-tab?) it sometimes did not refresh the display of the relevant labels & fields correctly. IIRC correctly, it was OK if the mouse was used to switch tabs. If that bug is happening again, please try removing line 49 of misc_dialogs.py, to see if that fixes things.
```
        notebook.enable_traversal()
```

tangledhelix:
> on macOS when you used the keyboard to switch between tabs (Cmd-tab & Shift-Cmd-tab?) it sometimes did not refresh the display of the relevant labels & fields correctly.

Cmd-tab and Shift-cmd-tab are globally bound by the OS and not available. Ctrl-tab and Shift-ctrl-tab are working to navigate tabs here. It seems to work like I'd expect mostly, except the first keypress doesn't navigate to the next tab, it seems to jump *to* the tab bar, and then subsequent keypresses navigates between the tabs. That might be the normal behavior.

It doesn't worry me at all, though, as I expect nearly all users will use the mouse pointer.

tangledhelix (approved):

srjfoo (approved):
oops -- deleted text that was put here in error.

PR #820, opened 2025-03-09
Add Ebookmaker integration

windymilla:
1. User needs to install ebookmaker themselves using the ebookmaker installation instructions.
2. User needs to tell GG where ebookmaker is installed.
3. Can choose which formats to be created (EPUB2 is always created)
4. Epub files are created in the project folder

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -20,6 +20,7 @@ from guiguts.html_tools import (
     html_validator_check,
     html_link_check,
     css_validator_check,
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes: it may be helpful to read [the manual section](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/HTML_Menu#Ebookmaker) (or press F1 while in the dialog)

PR #821, opened 2025-03-09
Add "Gray out unused view options" button

### LINKED ISSUE #723: Improvement to View Options
[The manual](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/Tools_Menu#Bookloupe_View_Options) suggests a method of making bookloupe messages manageable, which is often used by PPers:
1. Hide all messages
2. Turn on the first message type
3. Look to see if any messages appear
4. If so, fix, or remove the message
5. When you have no messages left, turn on the next message type and go back to point 3
6. When there are no more types left, you are done.

However, in any project, there are normally quite a few message types for which there were no actual messages, hence the need for point 3. It would be nice if either the toggle buttons for that message type were grayed out meaning "no point in you changing this as there aren't any of those", or maybe a single checkbox in the dialog, something like "Disable unused message types" 

Alternative could be to "Hide unused message types". This has the advantage of making the dialog less busy, but the disadvantage that users may get familiar with where specific buttons are, e.g. bottom of the second column. Would certainly need to be a toggle, rather than permanent behavior.

windymilla:
In View Options dialog (from Bookloupe checker) add a checkbox to gray out any view options that are not relevant to the current set of messages, i.e. if no messages of that type are shown in the checker dialog.

Gray-ness should be updated live, i.e. if a user hides the last message of a given type, and the View Options dialog is visible, the view option for that type will be grayed out. And if the user re-runs the tool, causing a message of the type to appear in the list again, the view option should be ungrayed.

This is to support a common way of using View Options as described in #723

Fixes #723

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -172,7 +172,7 @@ class CheckerViewOptionsDialog(ToplevelDialog):
 
             btn = ttk.Checkbutton(
                 check_frame,
-                text=option_filter.label,
+                text=f"{option_filter.label} (0)",
                 command=btn_clicked,
                 variable=check_var,
             )
@@ -205,27 +205,32 @@ class CheckerViewOptionsDialog(ToplevelDialog):
         self.checker_dialog.display_entries()
 
     def refresh_checkboxes(self) -> None:
-        """Set status of checkboxes based on if any messages of that type."""
+        """Set status of checkboxes based on if any messages of that type, and
+        update the (count) at the end of the label."""
         for row, option_filter in enumerate(self.checker_dialog.view_options_filters):
+            matches = self.filter_matches(option_filter)
+            self.checkbuttons[row]["text"] = re.sub(
+                r"\(\d+\)$", f"({matches})", self.checkbuttons[row]["text"]
+            )
             self.checkbuttons[row]["state"] = (
                 tk.DISABLED
-                if preferences.get(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS)
-                and self.filter_unused(option_filter)
+                if preferences.get(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS) and matches == 0
                 else tk.NORMAL
             )
 
-    def filter_unused(self, option_filter: CheckerFilter) -> bool:
-        """Return whether no CheckerEntry messages match the given filter.
+    def filter_matches(self, option_filter: CheckerFilter) -> int:
+        """Return how many messages match the given filter.
 
         Args:
             filter: Filter to be checked.
 
-        Returns: True if no entries would be selected by the filter.
+        Returns: Number of matching messages.
         """
+        matches = 0
         for entry in self.checker_dialog.entries:
             if option_filter.matches(entry):
-                return False
-        return True
+                matches += 1
+        return matches
 
     def on_destroy(self) -> None:
         if self.checker_dialog.winfo_exists():
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -9,7 +9,7 @@ import regex as re
 
 from guiguts.maintext import maintext
 from guiguts.mainwindow import ScrolledReadOnlyText
-from guiguts.preferences import PrefKey
+from guiguts.preferences import PrefKey, preferences, PersistentBoolean
 from guiguts.root import root
 from guiguts.utilities import (
     IndexRowCol,
@@ -192,7 +192,7 @@ class CheckerViewOptionsDialog(ToplevelDialog):
             btn_frame,
             text="Gray out unused view options",
             command=self.refresh_checkboxes,
-            variable=self.checker_dialog.suppress_unused_message_types,
+            variable=PersistentBoolean(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS),
         ).grid(row=0, column=2, padx=(40, 0))
         self.refresh_checkboxes()
 
@@ -209,7 +209,7 @@ class CheckerViewOptionsDialog(ToplevelDialog):
         for row, option_filter in enumerate(self.checker_dialog.view_options_filters):
             self.checkbuttons[row]["state"] = (
                 tk.DISABLED
-                if self.checker_dialog.suppress_unused_message_types.get()
+                if preferences.get(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS)
                 and self.filter_unused(option_filter)
                 else tk.NORMAL
             )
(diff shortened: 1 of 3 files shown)


srjfoo (approved):
I should have looked for a PR first, instead of commenting in the issue.

I'm fine with the way this works, but admit that I'd prefer to have the "Gray out unused view options" checkbox be sticky. And I'd like the manually disabled checkboxes to be sticky between sessions.

Those two together would be more or less equivalent to GG1's "Save" and "Load" defaults, wouldn't it? In addition to which, it looks like once the saved "defaults" are loaded, they're sticky between sessions.

windymilla:
> I'm fine with the way this works, but admit that I'd prefer to have the "Gray out unused view options" checkbox be sticky.

You said in the issue that it shouldn't be sticky 😁

> And yes, I'd recommend it beig a toggle, and not sticky

srjfoo:
> > I'm fine with the way this works, but admit that I'd prefer to have the "Gray out unused view options" checkbox be sticky.
> 
> You said in the issue that it shouldn't be sticky 😁
> 
> > And yes, I'd recommend it beig a toggle, and not sticky

So I did. And after I actually worked with it in the branch, I changed my mind.  😁  (I'm not so proud that I'm unwilling to change my mind -- and no, not just a woman's pregrogative, but anyone's. :D)

At this point I don't know exactly what I was thinking when I said that, but it probably wasn't that the options will all still be visible, regardless of whether or not they're greyed out.

windymilla:
The question of making individual checkboxes be sticky is a bit different. The problem is that if you have an option disabled, then you might think there are no errors of that type, when actually they are just hidden because that what what you did the previous time you used Bookloupe possibly on a different project.
The reason for having the default settings in GG1 was that we couldn't change what GG1 bookloupe reported, and some of the checks were outdated, like non ascii characters, so you found yourself always turning that one off.

windymilla:
Yes, it's no problem to change your mind of course. It was just funny that you said that, so I did something slightly different  in the code to make it non sticky. Then you said it should be sticky. I'll change it, and you can give it another try.

srjfoo:
> Yes, it's no problem to change your mind of course. It was just funny that you said that, so I did something slightly different in the code to make it non sticky. Then you said it should be sticky. I'll change it, and you can give it another try.

You make a valid point about there being the potential for a PPer to think that there aren't results for an option. Part of my concern is that bookloupe is, at least for me, pretty slow on my test file. Granted, it has 50,000+ lines and 923 pngs, but it's noticeably slower in GG2 than in GG1.

And note that I did check GG1 before I made the comment in the PR. So, the same potential for false assumptions exists in GG1.

If I were presented first with the options so that I could choose what to run instead of twiddling my thumbs while it runs, I might choose differently.

WF only runs one test at a time, but that test is sticky across sessions. Sometimes the different tools and the somewhat different ways they work can be confusing 😁

windymilla:
Yes, it is slower in GG2. Since it typically gets run once our twice per project, I don't think it's a problem. If it turns out to be a problem we could look at it. I should also check if I can make it quickest when lots of options are checked, rather than when none are checked (nts - order/parity of checking on/off and regex) 

> And note that I did check GG1 before I made the comment in the PR. So, the same potential for false assumptions exists in GG1.

I think GG1 is different because it's the default group of settings that is sticky. You know what you have turned on or off in the default group, so every time you start GG1 you get those defaults, eg hide non-ascii warnings. Individual check box settings are not saved between runs of GG1. 

Also note that bookloupe runs all the tests, and view options only changed what is visible. That is why changes to view options are pretty rapid - it doesn't rerun bookloupe. So showing the view options before running bookloupe would help much with speed. 

I agree that WF is an outlier. Pretty much all the tests, like pptxt, pphtml, bookloupe, etc., just list all the errors they find, even if there's a mixture of error types. WF is really several separate tests all displayed in the same dialog. 
FWIW, when I was PP/Ving I rarely used view options for bookloupe, but just ran the test and processed the results sequentially. I'm sure there are others who do the same, in contrast to those who turn on one error at a time.

srjfoo:
> Yes, it is slower in GG2. Since it typically gets run once or twice per project, I don't think it's a problem. If it turns out to be a problem we could look at it. I should also check if I can make it quickest when lots of options are checked, rather than when none are checked (nts - order/parity of checking on/off and regex)

That's fair. Testing something repeatedly should not be taken to apply to actual PP'ing.

> > And note that I did check GG1 before I made the comment in the PR. So, the same potential for false assumptions exists in GG1.
> 
> I think GG1 is different because it's the default group of settings that is sticky. You know what you have turned on or off in the default group, so every time you start GG1 you get those defaults, eg hide non-ascii warnings. Individual check box settings are not saved between runs of GG1.

As a Mac user, I especially liked the option to turn off the (obscenities deleted) `No CR` warning. 😁 

> Also note that bookloupe runs all the tests, and view options only changed what is visible. That is why changes to view options are pretty rapid - it doesn't rerun bookloupe. So showing the view options before running bookloupe would not help much with speed.

Ah, okay. Good point. In that case....

> I agree that WF is an outlier. Pretty much all the tests, like pptxt, pphtml, bookloupe, etc., just list all the errors they find, even if there's a mixture of error types. WF is really several separate tests all displayed in the same dialog. FWIW, when I was PP/Ving I rarely used view options for bookloupe, but just ran the test and processed the results sequentially. I'm sure there are others who do the same, in contrast to those who turn on one error at a time.

And I was reasonably happy with that approach, actually. I don't know how I'd use it if I were actually PP'ing for real.

srjfoo (approved):
I've done some more poking, and at this point, I actually think it's time to go with it. With the addition of the results counts, I think that it's time to see how the others who are actually using it like it. All they have to do if they don't like the greyed out option is to not enable it.

src/guiguts/checkers.py
```diff
@@ -135,22 +170,31 @@ def btn_clicked(row: int = row, var: tk.BooleanVar = check_var) -> None:
...
-        ).grid(row=0, column=1, padx=5)
         ttk.Button(btn_frame, text="Show All", command=lambda: self.set_all(True)).grid(
             row=0, column=0, padx=5
         )
+        ttk.Button(
+            btn_frame, text="Hide All", command=lambda: self.set_all(False)
+        ).grid(row=0, column=1, padx=5)
+        ttk.Checkbutton(
+            btn_frame,
+            text="Gray out unused view options",
```
tangledhelix, line 193: Suggest something like "Gray out options with 0 matches" (or "zero")

tangledhelix (approved):
Looks good; only the one suggestion about the text label, as discussed in Slack.

src/guiguts/checkers.py
tangledhelix: or "no matches" might be better.

PR #824, opened 2025-03-11
Add keyboard bindings to change tabs

### LINKED ISSUE #822: Add macOS tab navigation keys to Prefs dialog
Make Cmd+option+right/left arrow switch tabs, wrapping round.

windymilla:
Apply keyboard bindings to tabbed Notebook widget. Only example currently in code is Preferences dialog.

Already inbuilt were (Shift-)Ctrl-Tab to switch to prev/next. Added Cmd/Ctrl-digit, e.g. Ctrl-1 to change to first tab, to match behavior in other apps, e.g. browsers.
On Macs, added Command-Option-Left/Right to switch to prev/next to match behavior in browsers.

Fixes #822 

Testing notes: On macOS, Cmd-3 should change to the third tab, for example, and Cmd-Option-Left/Right should cycle round tabs.

### DIFF
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -28,6 +28,7 @@ from guiguts.widgets import (
     OkApplyCancelDialog,
     mouse_bind,
     Combobox,
+    Notebook,
 )
 
 COMBO_SEPARATOR = "―" * 20
@@ -44,7 +45,7 @@ class PreferencesDialog(ToplevelDialog):
         self.minsize(250, 10)
 
         # Set up tab notebook
-        notebook = ttk.Notebook(self.top_frame, takefocus=False)
+        notebook = Notebook(self.top_frame, takefocus=False)
         notebook.grid(column=0, row=0, sticky="NSEW")
         notebook.enable_traversal()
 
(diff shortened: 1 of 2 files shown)


windymilla:
@srjfoo - I've pushed a commit with some tests to see if one will fix the keyboard traversal problem of the tabs not being refreshed until you move the mouse across the tab boundary. 
The tests are from line 498 in widgets.py. There are 5 independent bits of code. Only one bit should be uncommented at a time - at the moment, they are all commented out. If test 1 fixes things, no need to do any other tests, etc.
If none of the 5 is sufficient to fix the problem, there are a few more examples along the lines of test 5 that I can rustle up, to do with forcing the focus away from then back to the tab.

srjfoo:
> The tests are from line 498 in widgets.py. There are 5 independent bits of code. Only one bit should be uncommented at a time - at the moment, they are all commented out. If test 1 fixes things, no need to do any other tests, etc.
> If none of the 5 is sufficient to fix the problem, there are a few more examples along the lines of test 5 that I can rustle up, to do with forcing the focus away from then back to the tab.

* Test 1 doesn't work; the results are different (tab 2 shows first time, but subsequently none of them do with either cmd+digit or cmd+option+arrow
* Test 2 works with cmd+digit but not with cmd+option+arrow
* Test 3 -- ditto test 2.
* Test 4 -- ditto 2 and 3.
* Test 5 matches test 1.

Defining "does not work" as "works the same as it does in the original branch".

srjfoo (approved):

PR #825, opened 2025-03-12
RTL experiment

windymilla:
Add advanced "Autofix RTL text" preference.
If this is False there is no change in behavior.
Defaults to False on Macs and True on Windows/Linux.

When a file is loaded, sections of Hebrew (ATM) are word-order reversed, i.e. order of chars within word is unchanged, just order of words changed.
When a file is saved, the same happens.
Also, and extra "Paste RTL text" menu option added for pasting in Hebrew from the forums.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -76,7 +76,7 @@ from guiguts.tools.bookloupe import bookloupe_check
 from guiguts.tools.jeebies import jeebies_check, JeebiesParanoiaLevel
 from guiguts.tools.levenshtein import levenshtein_check, LevenshteinEditDistance
 from guiguts.tools.pptxt import pptxt
-from guiguts.utilities import is_mac, folder_dir_str
+from guiguts.utilities import is_mac, is_windows, is_x11, folder_dir_str
 from guiguts.widgets import themed_style, theme_name_internal_from_user
 from guiguts.word_frequency import word_frequency, WFDisplayType, WFSortType
 
@@ -456,7 +456,12 @@ class Guiguts:
             PrefKey.IMAGE_AUTOFIT_HEIGHT, image_autofit_height_callback
         )
         preferences.set_default(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS, False)
-        preferences.set_default(PrefKey.AUTOFIX_RTL_TEXT, not is_mac())
+        if is_windows():
+            preferences.set_default(PrefKey.AUTOFIX_RTL_TEXT, "word")
+        elif is_x11():
+            preferences.set_default(PrefKey.AUTOFIX_RTL_TEXT, "char")
+        else:
+            preferences.set_default(PrefKey.AUTOFIX_RTL_TEXT, "off")
 
         # Check all preferences have a default
         for pref_key in PrefKey:
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -456,6 +456,7 @@ class Guiguts:
             PrefKey.IMAGE_AUTOFIT_HEIGHT, image_autofit_height_callback
         )
         preferences.set_default(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS, False)
+        preferences.set_default(PrefKey.AUTOFIX_RTL_TEXT, not is_mac())
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -562,6 +563,11 @@ class Guiguts:
         edit_menu.add_button(
             "To~ggle Column/Regular Selection",
             maintext().toggle_selection_type,
+        )
+        edit_menu.add_separator()
+        edit_menu.add_button(
+            "Paste ~Hebrew/RTL text",
+            maintext().paste_rtl,
         )
         edit_menu.add_separator()
         case_menu = Menu(edit_menu, "C~hange Case")
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -456,6 +456,8 @@ class Guiguts:
             PrefKey.IMAGE_AUTOFIT_HEIGHT, image_autofit_height_callback
         )
         preferences.set_default(PrefKey.CHECKER_GRAY_UNUSED_OPTIONS, False)
+        # Retain the RTL fix Pref for now, even though it is not in the Prefs dialog
+        # For debugging etc, it can be manually edited in the GGprefs file.
         if is_windows():
             preferences.set_default(PrefKey.AUTOFIX_RTL_TEXT, "word")
         elif is_x11():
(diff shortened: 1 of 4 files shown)


windymilla:
New commit implements char-reverse and word-reverse as separate options. Quick trial on Linux (WSL) seems to show that char-reverse is helpful for that. Native Windows needs word-reverse. I'll await testing from y'all, but for now, I've made the default setting "off" for macOS.

srjfoo:
I know nothing about Hebrew or how to use RTL text within LTR text. I dug up the "Help with: Hebrew" topic and paged back and picked a page at random (I was looking for a specific post that I didn't find, and so just wound up not worrying about it.)

image

The screenshot shows what the line in the forum post showed, and below that, what pasted into GG2 in macOS (I switched the font temporarily to TNR to better match the post). Whether I have RTL Text Autofix off or word, that's what I get.

Notice that the Hebrew phrases are rendered correctly within the phrase, but not placed correctly within the line when mixed with LTR text. This may just be ignorance on my part, so I'm happy to be instructed. (The duplicated second line, I pasted in the hebrew in what I thought was the correct place, and typed the English, then pasted in the second Hebrew phrase.

srjfoo:
I realized that I'd just used normal copy/paste, so I went back and used Paste Hebrew/RTL text, but saw no improvement. I'm getting what seems to me to be bizarre results when trying to mix LTR with RTL.

srjfoo:
Sorry for the barrage, but I decided to try one more thing. This screenshot is of the same line of text pasted into the Apple Pages app, and matches the forums post:

image

srjfoo:
Last screenshot for the night, I promise!

image

Top is the text from Tony's post in the forums. Bottom (line 3) is when I just copy and paste.

Middle (line number 1), which matches the post in the forums, I manually inserted `U+2067` before the Hebrew text and `U+2069` following the Hebrew text before typing the first bit of English, and again `U+2067` before the Hebrew text and `U+2069` following, before pasting the rest of the line in.

Note that I did not have to do that in Pages or TextEdit, but then, they are Mac apps. BBEdit gets it wrong with a straight copy/paste.

I tried again with the "Paste Hebrew/RTL text" menu option (no screenshot), thinking that perhaps that would wrap the isolates around the Hebrew text, but that turned out not to be the case. It's just as messed up for me, just in a different way.

windymilla:
@srjfoo - Just checking that you had turned on the "Fix RTL" Pref before using the "Paste Hebrew/RTL text"? It sounds like you had because you said it behaved differently to ordinary paste.

The only difference you might expect with the Fix RTL pref is that chunks of RTL text that are pasted will either be reversed character-wise or word-wise. It doesn't insert isolating marks at all.

Do either of you have any suggestions of things that should be changed or added, such as a "Surround selected RTL text with isolate marks" feature?

srjfoo:
> @srjfoo - Just checking that you had turned on the "Fix RTL" Pref before using the "Paste Hebrew/RTL text"? It sounds like you had because you said it behaved differently to ordinary paste.
> 
> The only difference you might expect with the Fix RTL pref is that chunks of RTL text that are pasted will either be reversed character-wise or word-wise. It doesn't insert isolating marks at all.

I definitely played around with the prefs pane, but I'll have to go check that. If you mean "Did I change the default Mac behavior", yes, I did. If there's another pref checkbox that I didn't notice, no, I didn't :D 

> Do either of you have any suggestions of things that should be changed or added, such as a "Surround selected RTL text with isolate marks" feature?

What do PPers who include RTL text within LTR text do? Do they use the isolating marks during post processing and then have those converted to the correct HTML tags? Or something else?

windymilla:
>  If you mean "Did I change the default Mac behavior", yes, I did. If there's another pref checkbox that I didn't notice, no, I didn't :D

Yes, that is what I meant, and no there isn't another checkbox that you didn't notice :D

> What do PPers who include RTL text within LTR text do? Do they use the isolating marks during post processing and then have those converted to the correct HTML tags? Or something else?

I don't know whether PPers end up having to include isolating marks or not. Since there haven't been lots of complaints from GG1 users and I don't recall seeing it discussed in the forum, my guess is that it's either a very rare situation, or that browsers cope fine as long as each stretch of text is self-consistent.

tangledhelix:
> 3. I'm not sure what the Linux default situation is like - if character order is wrong on Linux, it would be relatively easy to adjust the current reversal code to reverse all characters instead of (or as well as) reversing word order.

> 5. Hebrew should continue to work on macOS if the Pref setting is False. You may want to try setting it to True, in order to see what effect it has - it should display the words in the wrong order, so it will give an idea of what it's doing for Windows. Best to change the Pref setting between saving and loading, although since it's only reversing, it doesn't matter if you end up saving it the wrong way, you can just load with the setting off, then turn the setting on before you save, since reversing is reversible!

Regardless of the setting, Hebrew appears to display correctly in the text window. But the behavior does change on saving the file, and I get the predictable outcomes. The correct outcome (on Mac) is when the setting is off.

Screenshots below. My test text was: מי כּמוך בּאלים יי

The first 3 screenshots are the GG2 window with different modes. I set the preference before loading the file in each case. It always seems to display as it should, regardless.

The last screenshot is 3 saved files (in TextEdit, not GG2) showing the outcome as saved to disk. The correct version is when the RTL autofix was turned off.

Screenshot 2025-03-14 at 3 14 21 AM
Screenshot 2025-03-14 at 3 14 55 AM
Screenshot 2025-03-14 at 3 15 34 AM
Screenshot 2025-03-14 at 3 26 10 AM

srjfoo (approved):
After some instruction in Slack, I've retested (hopefully the right way) and it seems to be working as expected.

tangledhelix:
Test on Mac with python.org Python v3.13. Screenshot below. This is on `master` so without any of the RTL code. It seems that as with last night's testing, which was Homebrew Python 3.11, all works properly out of the box. Here is forums on left, GG2 in center, Textedit (to show saved results) on the right. Will test with RTL branch and comment on that as well.

Screenshot 2025-03-14 at 10 14 45 AM

tangledhelix:
More Mac results, now on `rtl` branch using python.org's Python 3.13 package.

By default, that is with autofix=off, everything seems to be correct using either paste, and on save. All good.
Screenshot 2025-03-14 at 10 23 52 AM

With autofix=word, the regular paste displays correctly in GG, but is backward on save. The RTL paste displays backward in GG, but is correct on save.
Screenshot 2025-03-14 at 10 25 12 AM

Similar with autofix=char -- regular paste displays properly in GG, but is wrong on save, and RTL paste displays wrongly in GG, but is corrected on save.

windymilla:
Changes made as suggested:

1. Remove Pref from Prefs dialog, but retain the Pref setting. In emergency/debugging, can edit Prefs file.
2. Make "Right-to-left Text" a submenu of Edit
4. Add "Surround selected text with RLM...LRM markers" to new menu
5. Add special paste button on Windows/Linux that applies the reversing algorithm (Windows=word, Linux=char)

tangledhelix:
Not testing Linux since @windymilla got there first.

tangledhelix (approved):

srjfoo (approved):
still looks good to me on macOS

PR #831, opened 2025-03-17
Fix line num click/drag bug & improve shift-drag

windymilla:
With split text window, using click & drag in the line numbers in one window, then immediately in the other got a bit confused about where the selection should be.

Also, to select down to a line using Shift-click you had to click in the line below the one you wanted to be the last selected line.

Also, the highlighted line number was often the line below the selection region which looked a bit odd.

Removed the old click & drag code and used the mechanism employed by shift-click & drag (generating events in the main text window). I'm not sure it's any better or simpler than the previous code, but at least all the line number clicking & dragging uses the same mechanism now, and the bug is fixed :)


windymilla:
Abandoning this method in favor of #833

PR #845, opened 2025-03-22
Add delete eol space & eop blank line features

windymilla:
Called automatically on file save, and also available in the Edit menu.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -618,6 +618,14 @@ class Guiguts:
             lambda: maintext().transform_selection(str.upper),
             "",
         )
+        edit_menu.add_separator()
+        edit_menu.add_button(
+            "Remove End-of-page Blank Lines",
+            lambda: maintext().delete_eop_blank_lines(),
+        )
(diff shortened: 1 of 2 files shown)


tangledhelix:
If they're automatically done on file save, what's the point of cluttering the menu with it?

windymilla:
> If they're automatically done on file save, what's the point of cluttering the menu with it?

In case you want to run them before saving? Maybe it's not necessary to be able to do that.
@okrick - what do you think? It would be nice not to have the in the menu, and just know that whenever they are saved, it will be after trimming trailing whitespace on each line and trailing blank lines on each page.

okrick:
@windymilla I don't need a menu item if they are done during saves.

windymilla:
> @windymilla I don't need a menu item if they are done during saves.

Excellent - thanks

windymilla:
As suggested by @tangledhelix & confirmed by Rick, removed entries from menu. Both types of blank removal are still run immediately prior to save.

srjfoo (approved):

PR #851, opened 2025-03-24
Implement autosave

### LINKED ISSUE #839: Add Auto Save
Implement the following auto-save and backup features within the `GG1 Preferences=>Backup` menu to enhance data safety and prevent data loss:

- `Enable Auto Save`: A toggle option to activate automatic saving of the current work file.
- `Set Auto Save Interval`: A field to allow users to define the frequency (e.g., in minutes) of automatic saves.
- `Keep a Backup Before Saving`: A toggle option to create a backup copy of the file before each manual or automatic save.

To maintain clear distinctions between backup types:

- Hard Backups (created when `Keep a Backup Before Saving` is enabled): Use the .bak file extension.
- Auto-Save Backups: Use the .bk1 and .bk2 file extensions.

These features are crucial for safeguarding user data against unexpected events such as file corruption, system updates, and power outages, significantly reducing the risk of losing valuable work. The distinct file extensions provide easy identification and management of different backup versions.

windymilla:
1. Pref to Save backup file when user saves manually. Backup file has extension `.bak`
2. Pref to autosave file every N minutes. When autosaved, backup files with extensions `.bk1` and `.bk2` are created. `.bk2` is a copy of the old `.bk1`, and `.bk1` is a copy of the old file, just before the new version is saved.
3. Note that although the user can force a manual save even if they have made no edits, an autosave will not happen if no edits are made.
4. Also, no autosave is done when the currently loaded file has no name, i.e. has never been saved.
5. Default autosave time period is 5 minutes. User can choose between 1 and 60 minutes.

Fixes #839

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -476,6 +476,9 @@ class Guiguts:
         preferences.set_default(PrefKey.EBOOKMAKER_EPUB3, True)
         preferences.set_default(PrefKey.EBOOKMAKER_KINDLE, False)
         preferences.set_default(PrefKey.EBOOKMAKER_KF8, False)
+        preferences.set_default(PrefKey.BACKUPS_ENABLED, True)
+        preferences.set_default(PrefKey.AUTOSAVE_ENABLED, True)
+        preferences.set_default(PrefKey.AUTOSAVE_INTERVAL, 5)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -119,6 +119,7 @@ class File:
         self.page_details = PageDetails()
         self.project_dict = ProjectDict()
         self.mainwindow: Optional[MainWindow] = None
+        self.autosave_id = ""
 
     @property
     def filename(self) -> str:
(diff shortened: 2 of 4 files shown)


windymilla:
Testing notes:
1. Suggest reducing time interval to 1 "minutes" (sic)
2. The two types of backup should work independently of one another, i.e. you can have either, neither or both active.
3. The bin files should be backed up at the same time as the main file.
4. If you want to restore a backup, you need to manually rename `myfile.txt.bak` to `myfile.txt`, and `myfile.txt.json.bak` to `myfile.txt.json`, for example.
5. It strikes me it would be nice to provide some kind of feature to facilitate the restoration of backups, but I'm not sure of the best way to make it work - would be a Future Feature, of course. One issue is not making it too easy for the user to say "restore last backup" only to find that it's worse than their current version which they have now overwritten.

srjfoo:
My testing so far has been with new text files. That is, starting with empty files that I then typed text into. I went ahead and set the auto-backup time to 1 minute.

Test 1: `bk2` was created before `bk1` was.
Test 2: Seemed to work as expected.
Test 3: `bk1` was created first, but `bk2` was not created at the 1-minute interval. In fact, I'd given up after 7 or 8 minutes and started writing this comment when I noticed that it had finally been created.

In both tests 1 and 3, I did a save (to create the initial `.bak` file before the auto-save ran for the first time.

So, definitely, creating the `.bak` file works as expected, but I'm wondering what's going on with auto-save, or if it's just me not understanding something I was supposed to do to set it up.

windymilla:
> My testing so far has been with new text files. That is, starting with empty files that I then typed text into. I went ahead and set the auto-backup time to 1 minute.
> 
> Test 1: `bk2` was created before `bk1` was. 

Just to clarify the algorithm on each autosave: if there's no filename (i.e. never saved) or the file hasn't been edited since the last save, it does nothing; if the filename `myfile.txt.bk1` exists, it is renamed to `myfile.txt.bk2` (overwriting the `bk2` file if it exists), then if `myfile.txt` exists it is renamed to `myfile.txt.bk1`, then the currently loaded text is saved as `myfile.txt`.

There are two possible meanings to this comment, one of which is  "impossible" (unless I'm misreading my code) and the other is correct behavior:
1. A `bk2` file appeared in the directory before a `bk1` file. Assuming the above algorithm is working correctly, this can't happen, because the only way to create a `bk2` is to rename a `bk1`
2. The "creation date" on the `.bk2` file is earlier than on the `.bk1` file. This is correct behavior, because the `bk2` file is the older saved version, the `bk1` is the more recent saved version, and the `txt` is the most recent saved version.

> Test 2: Seemed to work as expected. 
 
Good

> Test 3: `bk1` was created first, but `bk2` was not created at the 1-minute interval. In fact, I'd given up after 7 or 8 minutes and started writing this comment when I noticed that it had finally been created.

During this test, did you edit the file again after the bk1 file was created? The bk2 wouldn't be created if you didn't. Also, how were you checking for the files being created? I find in "File Explorer" (like Finder?) I sometimes have to manually refresh it to see files that have been created. Of course in a shell, you could use `ls` repeatedly to list the dir contents, something like:
```
while true; do ls -l; sleep 5; done
```
Another way to monitor what's happening is to look at the banner of GG - after you edit the file it says "edited", and immediately after it autosaves, the word "edited" is removed from the banner.

> In both tests 1 and 3, I did a save (to create the initial `.bak` file before the auto-save ran for the first time.

Since you are creating a file from scratch, you would have to do a save (to give it a filename) before you would expect autosave to do anything.

> So, definitely, creating the `.bak` file works as expected, but I'm wondering what's going on with auto-save, or if it's just me not understanding something I was supposed to do to set it up.

Hopefully the comments above will help work out what's going on.

srjfoo:
I'll do some more testing. Unfortunately, I can't show you evidence of (1) because the `bk1` file was eventually created. I suppose that it's always possible that watching a finder window was the wrong way to watch whether or not a file was created, but generally it does work, so .. dunno.

Very possibly the "no edit" was the reason for (3). I'll definitely test again!

srjfoo (approved):
Further testing shows everything okay.

PR #852, opened 2025-03-25
Fix Footnotes Autoset Chap LZ bug

windymilla:
Running Autoset Chap LZ in Footnote Fixup caused repeated refreshing of the dialog (probably forever) that could only be stopped by destroying the dialog, which then caused an exception.

Although fixing the first means the exception doesn't happen, it could probably be triggered by killing the dialog while it is mid-operation, so fixed exception too.

Testing notes:
Almost any file with footnotes should show the problem, I think. [This one](https://github.com/user-attachments/files/19447637/music.zip) certainly does.

### DIFF
--- a/src/guiguts/footnotes.py
+++ b/src/guiguts/footnotes.py
@@ -379,9 +379,10 @@ class FootnoteChecker:
         return -1
 
     def enable_disable_buttons(self) -> None:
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
I couldn't duplicate the error on macOS in `master`. However, the new code undoubtedly works.

PR #857, opened 2025-03-28
Fix several small bugs/annoyances

### LINKED ISSUE #836: PPtxt reports "suspects" erroneously
PPtxt is not supposed to report anything as a "suspect". 

That flag appears to depend on the setting of the `error_prefix`, which is wrongly set in PPtxt  for multiple occurrences of a character. So, `* * * * *` generates a warning about asterisks occurring `(x5)` - this multipliers is what triggers the "suspect" flag.

Behavior of "suspects" flag needs to be looked at carefully - what do we want it to do?

### LINKED ISSUE #854: PPhtml suggestion: target ID not found
`Target FNanchor_1 not found`
This is a false error due to the tag containing the ID being wrapped, same as the workbench pphtml. But it says so on the main page.

I suggest adding a disclaimer to the output, if you don't want to fix it.

Note that the HTML Link Checker reports no issues, maybe remove the check from PPhtml, or combine the link checker code.

[pphtml.txt](https://github.com/user-attachments/files/19468468/pphtml.txt)

### LINKED ISSUE #855: Miscellaneous small changes
1. Make tools that generate no errors output a message, such as "None found", or "Check completed", so that user knows the check actually ran. Example is HTML tags.
2. HTML & CSS Validators dialogs do not show the correct manual page when F1 is pressed
3. CSS Validator section in manual is missing
4. HTML & CSS Validator manual sections should make it clear this is equivalent to manually uploading the file to the online validator.
5. PPhtml reports `--- [info] 29 <h2> tags; 28 "class=chapter" attributes ---`, which should be `--- [info] 29 <h2> tags; 28 class="chapter" attributes ---` i.e. quotes round "chapter"
6. PPhtml reports, "Defined classes" can start with a comma, e.g. `, balance, bb, bbd, bl, bt, center, chap, chapter, cols,` ([example file](https://github.com/user-attachments/files/19488139/75730-h.zip))
7. Ctrl-click on a Spelling error adds to Proj Dict. It should remove the entry (all matching entries?) from the list, the same as the "Add to Proj Dict" button does
8. Change the default for Autosave to False - it's disconcerting if you weren't expecting it.
9. Document behavior of Autosave and "backup before save"

windymilla:
Fix erroneous "suspect" reports by PPtxt

PPtxt does not have suspect and non-suspect messages as some other tools do. The `(x5)` prefix showing that an asterisk occurred 5 times on the same line was
mistakenly being identified as a suspect flag.

Fixes #836

PPhtml: Detect ID attributes across line breaks

If `id="my_id"` is split across a line break, it wasn't detected by PPhtml, so might be reported as
`Target my_id not found.`

Use HTML parser instead of single line checks.
Note this error exists in other instances of pphtml.

Fixes #854

Output "Check complete" message in checkers

So that user knows check has run

Link HTML & CSS validators to man page

Manual page link was broken for these two tools.

Minor wording change to PPhtml message

Count of class="chapter" attributes now has quotes round "chapter"

Fix minor pphtml output bug

If CSS contained `.class1.class2`, the list of defined classes began with a superfluous comma, due to a
faulty regex.

Remove spelling when Ctrl-clicked

Cmd/Ctrl-click on a spelling adds it to the project dictionary, but didn't remove it from the spelling error list, unlike the "Add to Proj dict" button.

Fixes #855

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -477,7 +477,7 @@ class Guiguts:
         preferences.set_default(PrefKey.EBOOKMAKER_KINDLE, False)
         preferences.set_default(PrefKey.EBOOKMAKER_KF8, False)
         preferences.set_default(PrefKey.BACKUPS_ENABLED, True)
-        preferences.set_default(PrefKey.AUTOSAVE_ENABLED, True)
+        preferences.set_default(PrefKey.AUTOSAVE_ENABLED, False)
         preferences.set_default(PrefKey.AUTOSAVE_INTERVAL, 5)
 
         # Check all preferences have a default
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -77,7 +77,8 @@ class CheckerEntry:
             section: Section entry belongs to.
             initial_pos: Position of entry during initial creation.
             entry_type: Type of entry: header, footer, content, etc.
-            error_prefix: Prefix string indicating an error - empty if no error.
+            error_prefix: Prefix string indicating an error - empty or begins with
+                parenthesis if no error, e.g. "(x5)".
         """
         self.text = text
         self.text_range = text_range
@@ -92,7 +93,7 @@ class CheckerEntry:
             if text_range is None
             else (
                 CheckerEntrySeverity.ERROR
-                if error_prefix
+                if error_prefix and error_prefix[0] != "("
                 else CheckerEntrySeverity.INFO
             )
         )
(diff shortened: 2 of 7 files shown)


windymilla:
@rtonsing - particularly fixes #854 and #836 which I think you first noticed

rtonsing (approved):

srjfoo (approved):
Should WF have the "Check complete" message, too. Other than asking about that and the duplicate "complete" message in the spell checker, looks good to me.

windymilla:
I removed the duplicate "test completed" from Spell Check, but left the initial "test started" message because that can say "(selected text only)", which is useful to know. 
I decide WF didn't need the message, but thanks for thinking of it.

Issue #829, opened 2025-03-16
Further ebookmaker investigation/fixes


windymilla:
1. If ebookmaker is installed on Windows, it then appears to work from in Guiguts.
2. So far, it has not been possible to do the same on macOS or Linux
3. It's possible it's related to having several versions of python installed, but the situation is currently unclear

Release 2.0.0-alpha.16 will have the feature, documented that it maybe only works on Windows, in the hope that others may be able to try to get it working on other platforms.

This issue is a place to add more info as it is gathered, and hopefully work out how to get it working across all platforms.

susanskinner:
I have installed ebookmaker (v 0.13.6 on Windows 11) and can successfully launch it as an external program from GG1 as described at https://github.com/gutenbergtools/ebookmaker/blob/master/README.md However, GG2 appears to be looking for ebookmaker in the local project directory because when I try to run it, I get the error message:

===== Ebookmaker errors =====
C:\Users\susan\AppData\Local\Programs\Python\Python311\python.exe: can't open file 'C:\\Users\\susan\\Dropbox\\DP\\PP\\chambers\\1886\\18861120\\ebookmaker': [Errno 2] No such file or directory

Do I need to tell GG2 where to look for ebookmaker and, if so, how?

susanskinner:
Also is "Ebookmaker check" the right label for the menu option? I expected this to ask me where an already generated epub file was, similar to the "W3C EPUBCheck" option in GG1, not to try to generate the epubs. Would "Ebookmaker generate/check" be better? (Assuming that it's going to do both.)

jjzsdp:
I think ebookmaker will be looking for the virtualenvs location, something like:
C:\Users\myname\.virtualenvs\ebookmaker-cgaQuYhi

tangledhelix:
I spent a lot of time trying to get Ebookmaker working with GG2 on Mac and Linux, and was repeatedly stymied by the way pipenv installs. I got it to work on both platforms in about the same way @susanskinner seems to have: it works fine when launched from my terminal window, but I haven't convinced GG2 to find ebookmaker's virtual environment so far.

@windymilla got it working on his machine, and it seems @jjzsdp may have as well, but I'm not surprised @susanskinner is having a similar problem to what I've run into, even on Windows where it has worked for at least some users.

I got pulled away from looking at that for a while, but it's still on my list to poke at it more.

windymilla:
[...]

PR #861, opened 2025-03-31
Ebookmaker Mac/Linux support & other changes/improvements

tangledhelix:
This PR has been tested against Mac and Linux (Ubuntu 24).

**Windows testing is needed!**

### Changes
- Adds checkbox to Ebookmaker dialog to control debug output. By default `-v` is used (which will give INFO), with debug box checked, you also get DEBUG (`-v -v`).
- The path to ebookmaker is now expected to be a **directory** rather than a file. It should be the directory where the user ran `pipenv install` to install ebookmaker. **This is not backward-compatible** so users may need to change its value for ebookmaker to work again **even if it was working previously**.
    - Since Ebookmaker's instructions for install are to use pipenv, my viewpoint is to follow that method and assume that's how the user installed it.
- Minor correction of a label, "Epub formats:" is now "Ebook formats:"
- Set `PIPENV_IGNORE_VIRTUALENVS` in the environment to avoid complications with pipenv if running from within an existing virtual environment. This was often the case during development, but it could easily be the case that users run Guiguts from their own virtual environment.
- Improve detection of the Calibre `ebook-convert` program by using `shutil.which()`
- Ebookmaker appears to never output anything to stdout, only to stderr. I simplified the output handler to combine both streams into one and just output that rather than separate them. Previously, everything was grouped under "Errors" even if it was just debugging or general information.

### TESTING NOTES
To install ebookmaker, you can use [their instructions](https://github.com/gutenbergtools/ebookmaker/blob/master/README.md). Here's the TL;DR version:

```
# make sure you have pipenv; if not, then install it.
which pipenv

# put this in whatever directory name you want
mkdir ebookmaker
cd ebookmaker
pipenv install ebookmaker

# now test it out
pipenv run ebookmaker --version

# if you get an error about pkg_resources, then install setuptools too.
pipenv install setuptools

# test again if needed
pipenv run ebookmaker --version
```

If you want to test MOBI generation, also install [Calibre](https://calibre-ebook.com) and make sure the `ebook-convert` program is on your path.

```
# mac/linux
which ebook-convert

# windows
where ebook-convert
```

- Open an HTML file in Guiguts
- In the HTML menu, run the Ebookmaker item
- Click "Browse" and navigate to the directory where you installed Ebookmaker with pipenv, then select it
- Click "Re-Run" and wait
- You should see a number of INFO: lines and now have at least one ebook file. It will be in the same folder where your HTML file is.
- Check "Show debug output" and click "Re-Run".
- The same books should be built, but now you'll have a ton of DEBUG: lines in the output.
- If you installed Calibre, you can check the kf8 or mobi boxes and click "Re-run". Then you should find .mobi file(s) in your project directory after the run finishes.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -476,6 +476,7 @@ class Guiguts:
         preferences.set_default(PrefKey.EBOOKMAKER_EPUB3, True)
         preferences.set_default(PrefKey.EBOOKMAKER_KINDLE, False)
         preferences.set_default(PrefKey.EBOOKMAKER_KF8, False)
+        preferences.set_default(PrefKey.EBOOKMAKER_VERBOSE_OUTPUT, False)
         preferences.set_default(PrefKey.BACKUPS_ENABLED, True)
         preferences.set_default(PrefKey.AUTOSAVE_ENABLED, True)
         preferences.set_default(PrefKey.AUTOSAVE_INTERVAL, 5)
--- a/src/guiguts/html_tools.py
+++ b/src/guiguts/html_tools.py
@@ -9,6 +9,7 @@ import tkinter as tk
 from tkinter import ttk, filedialog
 from typing import Optional, Any
 import xml.sax
+import shutil
 
 from PIL import Image, ImageTk, UnidentifiedImageError
 import regex as re
@@ -29,6 +30,7 @@ from guiguts.utilities import (
     DiacriticRemover,
     IndexRowCol,
     folder_dir_str,
+    is_windows,
 )
 from guiguts.widgets import ToplevelDialog, Busy
 
@@ -1126,14 +1128,14 @@ class EbookmakerCheckerDialog(CheckerDialog):
         )
 
         def locate_ebookmaker() -> None:
-            """Prompt user to give path to ebookmaker."""
-            if file_name := filedialog.askopenfilename(
-                title="Select Ebookmaker program",
+            """Prompt user to give path to where ebookmaker was installed using pipenv."""
+            if dir_name := filedialog.askdirectory(
+                title=f"Select the Ebookmaker 'pipenv install' {folder_dir_str()}",
                 parent=self,
             ):
-                preferences.set(PrefKey.EBOOKMAKER_PATH, file_name)
+                preferences.set(PrefKey.EBOOKMAKER_PATH, dir_name)
 
-        ttk.Label(self.custom_frame, text="Ebookmaker Location:").grid(
+        ttk.Label(self.custom_frame, text=f"Ebookmaker {folder_dir_str()}:").grid(
             row=0, column=0, sticky="NSW", pady=(5, 0)
         )
         ttk.Entry(
@@ -1148,11 +1150,11 @@ class EbookmakerCheckerDialog(CheckerDialog):
             takefocus=False,
         ).grid(row=0, column=2, sticky="NSE", pady=(5, 0))
 
-        ttk.Label(self.custom_frame, text="Epub formats:").grid(
+        ttk.Label(self.custom_frame, text="Ebook formats:").grid(
             row=1, column=0, sticky="NSW"
         )
         format_frame = ttk.Frame(self.custom_frame)
-        format_frame.grid(row=1, column=1, sticky="NSW", columnspan=2, pady=5)
+        format_frame.grid(row=1, column=1, sticky="NSW", pady=5)
         ttk.Checkbutton(
             format_frame,
             text="EPUB 2",
@@ -1175,6 +1177,14 @@ class EbookmakerCheckerDialog(CheckerDialog):
             variable=PersistentBoolean(PrefKey.EBOOKMAKER_KF8),
         ).grid(column=3, row=0, sticky="NSW", padx=5)
 
+        verbose_frame = ttk.Frame(self.custom_frame)
+        verbose_frame.grid(row=1, column=2, sticky="NSE", pady=5)
+        ttk.Checkbutton(
+            verbose_frame,
+            text="Verbose output",
+            variable=PersistentBoolean(PrefKey.EBOOKMAKER_VERBOSE_OUTPUT),
+        ).grid(column=0, row=0, sticky="NSE", padx=5)
+
 
 class EbookmakerChecker:
     """Ebookmaker checker"""
@@ -1190,24 +1200,50 @@ class EbookmakerChecker:
         if not cwd:
             Busy.unbusy()
             return
-        cwd = os.path.dirname(cwd)
         if not os.path.isdir(cwd):
             logger.error(f"{folder_dir_str()} {cwd} does not exist")
             return
 
-        # Set env variable to suppress pipenv whinge about running in virtual environment
-        # (possibly only relevant to devs?)
+        # Don't use an existing virtualenv for pipenv; create a new one.
+        # Using an existing one makes testing from an IDE very hard, and
+        # would also be problematic for anyone running Guiguts in their
+        # own virtualenv (a common practice among Python users).
         my_env = os.environ.copy()
-        my_env["PIPENV_VERBOSITY"] = "-1"
+        my_env["PIPENV_IGNORE_VIRTUALENVS"] = "1"
+
+        # Find the proper path to the venv
+        result = subprocess.run(
+            ["pipenv", "--venv"],
+            capture_output=True,
+            text=True,
+            cwd=cwd,
+            env=my_env,
+            check=False,
+        )
+        venv_dir = result.stdout.rstrip()
+        if not os.path.isdir(venv_dir):
+            logger.error(f"{folder_dir_str()} {venv_dir} does not exist")
+            return
+
+        if is_windows():
+            bin_dir = os.path.join(venv_dir, "Scripts")
+        else:
+            bin_dir = os.path.join(venv_dir, "bin")
+        venv_python_path = os.path.join(bin_dir, "python")
+        venv_ebookmaker_path = os.path.join(bin_dir, "ebookmaker")
+
         # Base command
-        command = ["pipenv", "run", "python", "ebookmaker", "--max-depth=3"]
+        command = ["pipenv", "run", venv_python_path, venv_ebookmaker_path, "-v"]
+        if preferences.get(PrefKey.EBOOKMAKER_VERBOSE_OUTPUT):
+            command.append("-v")
+        command.append("--max-depth=3")
         # Build options
         if preferences.get(PrefKey.EBOOKMAKER_EPUB2):
             command.append("--make=epub.images")
         if preferences.get(PrefKey.EBOOKMAKER_EPUB3):
             command.append("--make=epub3.images")
         # Need Calibre to create kindle versions
-        if "calibre" in my_env["PATH"].lower():
+        if shutil.which("ebook-convert"):
             if preferences.get(PrefKey.EBOOKMAKER_KINDLE):
                 command.append("--make=kindle.images")
             if preferences.get(PrefKey.EBOOKMAKER_KF8):
@@ -1216,7 +1252,7 @@ class EbookmakerChecker:
             PrefKey.EBOOKMAKER_KF8
         ):
             logger.error(
-                "To create Kindle files, install Calibre and ensure it is on your PATH"
+                "To create Kindle files, install Calibre and ensure the ebook-convert command is on your PATH"
             )
             Busy.unbusy()
             return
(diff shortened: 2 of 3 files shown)


tangledhelix:
This might possibly fix #829

tangledhelix:
Possible to-do's to improve this PR:

- Make the changes suggested by @windymilla in #829 
- Look at how we use color in checker items and add colors to INFO and DEBUG tags which are otherwise hard to differentiate as you read
    - If debug output is not active, colors probably aren't useful/needed, so skip them?

tangledhelix:
Another to-do: address the FILESDIR message with a stub config file? Setting it in the environment didn't work. Ebookmaker's documentation suggests a `.ebookmaker` file to set this.

windymilla:
Let me start by saying "Thanks!" It seems there's definitely a way forward for us to have code that runs on all platforms.

Notes (not necessarily in priority order):
1. Installation method (i.e. `pipenv install ebookmaker`) appears to work on Windows, and it creates a pipfile, lockfile & virtualenv dir. However, you can't run ebookmaker with `pipenv run ebookmaker --version` from the dir where you typed the install command (in my case C:\DP\ebookmaker). You have to run 
`pipenv run python C:\Users\myname\.virtualenvs\<name of virtualenv>\Scripts\ebookmaker --version`
i.e. you have to manually specify the location of the virtualenv, rather than relying on pipenv to handle that. This is documented in the ebookmaker install docs in [my notes for Windows users](https://github.com/gutenbergtools/ebookmaker/blob/master/README.md#notes-running-ebookmaker-on-windows-machine-adapted-from-windymilla). This is the reason that the code currently in `master` asks the user to find ebookmaker in the virtualenv dir (the code just uses the dir, not the ebookmaker script, but seemed easier to explain to a user what they should be looking for).
2. Minor: Instead of "directory" in labels, titles, etc., use `folder_dir_str()` so that it will say "folder" on Windows.
3. For the reasons described in (1) the existing code won't work on (at least my, and maybe all) Windows. See screenshot which shows the same message as trying to run `pipenv run ebookmaker` from a shell in the dir you typed the install command.
image. 
Changing the selected dir to the virtualenv dir (e.g. `C:/Users/<username>/.virtualenvs/ebookmaker-cgaQuYhi/Scripts`, like the `master` code uses) also doesn't work. Nor does `C:/Users/<username>/.virtualenvs/ebookmaker-cgaQuYhi` - it says it's creating a virtualenv, but doesn't run ebookmaker. If you then try to re-run, it comes up with the same message as shown in the screenshot.
4. `PIPENV_IGNORE_VIRTUALENVS` seems to work on Windows too, as did `PIPENV_VERBOSITY`, but good to have something that works cross platform
5. New Calibre path checking still works on Windows
6. As noted elsewhere, ebookmaker info/debug/error/critical output is a bit of a mess. Lots of messages are not helpful for the user. I had previously thought it was better not to use the `-v` at all so that you didn't get all of these and could focus on actual errors that needed fixing. However, some users might prefer it because they are used to ebookmaker's verbosity. One possibility would be to change the `Debug` checkbox to `Verbose` (as some other GG tools have) which controls whether a single `-v` flag is passed. If we (devs) need to debug, we can add a second `-v` to the code. Personally, I don't think we should have/need a `Debug` checkbox visible to users. Combining stderr & stdout seems sensible and works fine on Windows.

windymilla:
Notes about making this PR work on Windows:
1. User needs to choose the virtualenv folder (e.g. `C:/Users/<username>/.virtualenvs/ebookmaker-cgaQuYhi/Scripts`) - this is just a documentation issue, not a code change.
2. Base command needs changing from
`["pipenv", "run", "ebookmaker", "-v"]`
to 
`["pipenv", "run", "python", "ebookmaker", "-v"]`
If this works on Mac/Linux, great! If not, we need it to be OS-specific

windymilla:
> Possible to-do's to improve this PR:

> * Look at how we use color in checker items and add colors to INFO and DEBUG tags which are otherwise hard to differentiate as you read
>   
>   * If debug output is not active, colors probably aren't useful/needed, so skip them?

As mentioned above, I don't think we need debug for users, so we won't need color.

tangledhelix:
> Notes about making this PR work on Windows:
> 
> 1. User needs to choose the virtualenv folder (e.g. `C:/Users/<username>/.virtualenvs/ebookmaker-cgaQuYhi/Scripts`) - this is just a documentation issue, not a code change.
> 2. Base command needs changing from
>    `["pipenv", "run", "ebookmaker", "-v"]`
>    to
>    `["pipenv", "run", "python", "ebookmaker", "-v"]`
>    If this works on Mac/Linux, great! If not, we need it to be OS-specific

If you run `pipenv --venv` do you get back the path to the venv? If so, then maybe we can have the user behavior (choose the directory) be the same for everyone, while still making it work correctly. Aside from being more consistent on the user side, it's probably more user-friendly than having to find the venv themselves. (Imagine they have more than one ebookmaker venv on disk at some point, how do they know which one to use?)

My thought here is to capture that output and use the learned path in the command. So instead of `"python", "ebookmaker"` above, you'd have something like

```python
    command.append(f"{venv_path}/bin/python")
    command.append(f"{venv_path}/Scripts/ebookmaker")
```

The real code would probably use `os.path.join()` instead, but this is enough to illustrate the idea, hopefully.

tangledhelix:
@windymilla Same question about `pipenv --py`

tangledhelix:
I see a path to making it more unified, if those commands work on Windows. I can gather information about the Python and Virtualenv in use, e.g.

```
❯ pipenv --py
/Users/dan/.local/share/virtualenvs/eb-hQh4ziwu/bin/python

❯ pipenv --venv
/Users/dan/.local/share/virtualenvs/eb-hQh4ziwu
```

There is also an ebookmaker command, although the path is slightly different (bin, not Scripts):

```
❯ find $(pipenv --venv) -name ebookmaker -type f
/Users/dan/.local/share/virtualenvs/eb-hQh4ziwu/bin/ebookmaker
```

Putting it together in a similar fashion to  Windows, if we account for the path separators with `os.path.join()` and that the scripts directory is different on different platforms, we might be able to use something like this cross-platform:

```
❯ pipenv run $(pipenv --py) $(pipenv --venv)/bin/ebookmaker --version
EbookMaker 0.13.6
```

I wouldn't use those shell expansions, obviously; we could collect those by calling those commands with `subprocess.run()` and collecting their output, then iterpolating the result into the final command. It's a couple of extra process calls, but compared to the runtime of ebookmaker itself it would be negligible.

tangledhelix:
> 4. Change `Debug` checkbox to `Verbose`. Pass single `-v` arg only if verbose. Never pass `-v -v`.

I'm not sure whether I agree, only because `-v -v` is the "always on, no choice" behavior for online ebookmaker. I think it should be possible for users to replicate that. How many knobs we give them is up for discussion, of course.

> 5. In non-verbose mode the user won't get FILESDIR info messages. Maybe nothing else needs doing about that. GG1 does filter out some ebookmaker messages, which is another alternative for us. (View Options is another possibility, but probably overkill)

That can be overcome by including a config file somewhere and pointing ebookmaker to it.

File looks something like
```
[PATHS]
FILESDIR = "/"
```

We could write this with code so it uses `os.path.expanduser("~")` to decide where to point, but for our purposes it only matters that it's set to something. We specify all the paths so I don't think it's actually being used for anything when we run ebookmaker.

Then you point ebookmaker to the file with `--config <config_file>`

(This is the solution that the ebookmaker author offered to PG when they were getting those in the online ebookmaker output.)

I'd prefer if putting `FILESDIR` into the environment worked, but it doesn't seem to.

windymilla:
> There is also an ebookmaker command, although the path is slightly different (bin, not Scripts):

This seems to be the main (only) difference. `bin`/`Scripts` subdir for both python and ebookmaker.

> ```
> ❯ pipenv run $(pipenv --py) $(pipenv --venv)/bin/ebookmaker --version
> EbookMaker 0.13.6
> ```

So, mixing python & shell, instead of `$(pipenv --venv)/bin/ebookmaker` it would need to be something like 
`dirname($(pipenv --py))/ebookmaker` 
if we can assume that ebookmaker and python are in the same dir on all platforms (they are on Windows, and sounds like they are on Macs from what you've said)

windymilla:
> > 4. Change `Debug` checkbox to `Verbose`. Pass single `-v` arg only if verbose. Never pass `-v -v`.
> 
> I'm not sure whether I agree, only because `-v -v` is the "always on, no choice" behavior for online ebookmaker. I think it should be possible for users to replicate that. How many knobs we give them is up for discussion, of course.
> 
Agreed elsewhere to use `-v -v` for Verbose, and `-v` for non Verbose

> > 5. In non-verbose mode the user won't get FILESDIR info messages. Maybe nothing else needs doing about that. GG1 does filter out some ebookmaker messages, which is another alternative for us. (View Options is another possibility, but probably overkill)
> 
> That can be overcome by including a config file somewhere and pointing ebookmaker to it.
> 
> File looks something like
> 
> ```
> [PATHS]
> FILESDIR = "/"
> ```
> 
> We could write this with code so it uses `os.path.expanduser("~")` to decide where to point, but for our purposes it only matters that it's set to something. We specify all the paths so I don't think it's actually being used for anything when we run ebookmaker.
> 
> Then you point ebookmaker to the file with `--config <config_file>`
> 
> (This is the solution that the ebookmaker author offered to PG when they were getting those in the online ebookmaker output.)
> 
> I'd prefer if putting `FILESDIR` into the environment worked, but it doesn't seem to.

Me too - it's a pain to have to have a config file, which we'll have to write and point to. Strongly tempted to just filter out those messages, which would have the same effect, but with less fuss. 👿

windymilla (commented):
Generally very happy, with the points tackled as we've discussed.

tangledhelix:
> So, mixing python & shell, instead of `$(pipenv --venv)/bin/ebookmaker` it would need to be something like `dirname($(pipenv --py))/ebookmaker` if we can assume that ebookmaker and python are in the same dir on all platforms (they are on Windows, and sounds like they are on Macs from what you've said)

Oooh, that might work. I'll have to check Linux but I'm guessing it uses the same naming conventions as Mac.

tangledhelix:
> Me too - it's a pain to have to have a config file, which we'll have to write and point to. Strongly tempted to just filter out those messages, which would have the same effect, but with less fuss. 👿

That's true, it would be easy enough in the output handler. I'll add that as well.

tangledhelix:
Pushed new changes:
- Don't include FILESDIR lines
- Auto-detect pipenv virtualenv location
- One command structure that should work on all platforms
- Fix label of "directory" vs "folder" per platform
- Change "Debug" label to "Verbose"

Tests OK on Mac & Linux.

tangledhelix:
Not addressed:
- attempting to color the lines in the checker list (e.g. CRITICAL)
- not fixing the "busy" flag when ebookmaker location isn't yet set

windymilla (approved):
Works beautifully on Windows!

windymilla:
I updated https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/HTML_Menu#Ebookmaker to tell the PPer to locate the dir where `pipenv install ebookmaker` was typed.

srjfoo (approved):
Tested on macOS 15.4

The `TL;DR` instructions aren't quite as clear as the testing notes in this PR, so I was quite glad to have the abbreviated instructions.

One thing that I learned (at least for me), was that the version of Calibre that I downloaded from their website did not appear to have the command line tools. At least, if they did, `which ebook-convert` didn't find it. I wound up doing a web search for `ebook-convert`, and was directed to homebrew. 😁  -- once I did that, I had no problems.

As @tangledhelix found, the Kindle app doesn't want to open the kindle generated files unless you log into Amazon. Both Calibre and OmniReader on the Mac open them just fine, but I'm not patient enough to do an exhaustive search to find something that will on the iPad (neither one is available for iPad that I could find) :D 

One small niggle -- in the results count in the upper left-hand corner of the widget, I didn't see anything other than a zero there. Is there some meaning to that number?

windymilla:
> One small niggle -- in the results count in the upper left-hand corner of the widget, I didn't see anything other than a zero there. Is there some meaning to that number?

It's not directly related to this PR of course, but in general that count is of messages that point to a specific location in the file, i.e. the message starts with "27.4:", for example.
In the [follow-on issue](#829) to distinguish warnings/errors from debug/info messages, it will hopefully be clearer which issues the user needs to address. I'm not yet sure if those will show as "Suspects", despite not having a line number. I may need to make it possible for messages to be counted and potentially be suspects even when there isn't a line number in the text file to point to (maybe by using a line/col of "0.0".

Issue #762, opened 2025-02-13
Suggestion that keyboard shortcuts be user modifiable


rtonsing:
I have advocated before in GG 1 to use Ctrl-G for "Goto Line Number", to match Notepad++, Visual Studio, VS Code, and others.

I understand why it wasn't accepted, since others are used to using it for "Find next", so I just modified the code for my use. Inconvenient when updating to new versions.

It would be nice to have a means for users to change the shortcut key mappings, saved with other preferences.

windymilla:
I think this is covered by #143. If so, please close. If not, please add a bit more to explain what's needed. Thanks

rtonsing:
I saw that, but I didn't think it clearly said that _any_ shortcut could be reassigned. It talks about using a preceding key combination to the user defined shortcut, which is not what I want. The discussion confused things, to me.

I would add that it would also be nice to (re)assign shortcuts to menu items that don't have default ones. I'm not clear whether #143 also says so.

Maybe this could be closed, if the intent of #143 is clarified, so that it means the same thing.

windymilla:
I've made this a sub-issue of #143 - first time I've used that feature. No idea if it's useful. At least in my head, if not explicitly in #143, user keyboard shortcuts such as you are requesting should added, if we are able to.

Issue #143, opened 2024-02-18
User defined keyboard shortcuts


windymilla:
Relates to #135, since once every command has a unique identifier, it makes user-defined shortcuts for any command possible.
Note that Tk supports multiple keystrokes, so it could be implemented such that the user hits a key combination prior to their chosen shortcut, e.g. `Ctrl+u`, `Ctrl+l` to mean "User shortcut", "Lowercase selection".
This would leave `Ctrl+l` free as a system-defined shortcut (need to check if that's true).

windymilla:
Comment from alpha tester: 

> It would be very helpful to have keyboard shortcuts for changing the case of text in GuiGuts. Specifically, shortcuts for UPPERCASE, lowercase, and Title Case functionality would be much appreciated.  Alternatively, for easier access, consider moving the change case functions to the main Edit menu.

tangledhelix:
> Note that Tk supports multiple keystrokes, so it could be implemented such that the user hits a key combination prior to their chosen shortcut, e.g. `Ctrl+u`, `Ctrl+l` to mean "User shortcut", "Lowercase selection".

User-defined shortcuts aside, this would be of use to us for the defaults as well IMHO. As we've seen, it can be quite challenging to find keys that aren't used already and also work on all the platforms. VSCode went the multi-key route as well (a lot of things are hidden behind Cmd-K, <something>).

windymilla:
Yes, it could. They are a little more awkward to use though, and it seems a shame to say to the many Windows users (for example) that they have to use `Ctrl+u Ctrl+l` rather than just `Ctrl+l` because `Cmd+l` means something on macOS.
It will be a question of balancing our desire to reduce cross-platform differences, and not making things awkward for everyone because they are tiptoeing around the requirements of an OS they don't use.

tangledhelix:
It's not only useful due to conflicts (such as Win vs. Mac) although that might help. What I really meant is might be useful to fit more keys into the overall keymap. There are only so many keys on the keyboard, and we'll end up running out.

srjfoo:
And it's apparently not even just differences between Win and macOS -- witness the problem with `cmd`+`;` for Compose Sequence on Linux.

srjfoo:
> Yes, it could. They are a little more awkward to use though, and it seems a shame to say to the many Windows users (for example) that they have to use Ctrl+u Ctrl+l rather than just Ctrl+l because Cmd+l means something on macOS.
[...]

PR #862, opened 2025-03-31
Implement Command Palette

### LINKED ISSUE #135: Idea: command palette
#133 got me thinking about keyboard shortcuts and where we should and should not be using them. I think I'll err on the side of "no shortcuts" for that use-case, in the spirit of the Apple [HIG][] viewpoints:

1. Support keyboard-only interaction when possible
2. Don't re-purpose standard shortcuts
3. Define custom shortcuts **only for the most common actions** (_emphasis mine_)
4. Prefer `Cmd`, then `Shift+Cmd`. Use `Opt` only sparingly. Avoid `Ctrl` whenever possible.

[hig]: https://developer.apple.com/design/human-interface-guidelines/keyboards#Best-practices

Having fewer shortcuts avoids overloading people, since there is less to memorize in an app. But GG1 has a lot of menu items, and thus we can expect GG2 will eventually have many menu items.

Perhaps we can follow a command-palette model to make the commands findable via keyboard-only interaction, and also more findable in general.

Brief description of a command palette for those not familiar with the term:
- Press a key (usually Cmd-P on Mac, Ctrl-P otherwise)
- A text input popover appears
- As you type, commands matching the input populate a list
- Select a command to execute it

Example: type "low" into palette and it might locate the "lowercase selection" command. Or type "proofer" or "comment" to choose the "Find Next Proofer Comment" or "Find Previous Proofer Comment" commands. The entire catalog of available commands could be available.

To improve discoverability, it wouldn't be a bad idea to have a visible button on the window that triggers the palette, and perhaps some sort of tooltip text to hint what to do next when it appears.

windymilla:
1. In Help menu, or use Cmd+Shift+P/Ctrl+Shift+P
2. Type in dialog to filter list of commands
3. Use Run button, Enter key, or double click to run the selected command
4. Navigate to & within list using up/down arrows or Ctrl+P/N

Fixes #135 

Testing notes: It also supports hidden/orphan commands, i.e. ones that don't appear in a menu. I have added one for you to find. Hint: it's not in a menu, so should be easy to find in the list of commands


windymilla:
Forgot to say, I've only put command buttons in for now as proof of concept. When we've got it working as we want it, we can add checkbox buttons too.

src/guiguts/widgets.py
```diff
@@ -678,6 +679,174 @@ def destroy(self) -> None:
...
+                if metadata := entry.get_submenu_metadata(submenu):
+                    return metadata
+        return None
+
+    def add_button(self, label: str, command: Callable, shortcut: str = "") -> None:
+        """Add a command button to this menu."""
+        button = CommandMetadata(label, command, shortcut)
+        self.entries.append(button)
+
+    def get_all_commands(self) -> list[CommandInfoMetadata]:
```
tangledhelix, line 750: Using vscode as an example, I often want to run the same command I just ran. Perhaps we can store the last-used command somewhere and place it at the top of the stack instead of in its sorted place?

Even better would be to keep the most recent handful (5?) but that would require more storage locations, a rotation and cleanup algorithm, etc... I'd consider just the most recent one a win.

src/guiguts/application.py
```diff
@@ -501,6 +506,11 @@ def init_menus(self) -> None:
         self.init_help_menu()
         self.init_os_menu()
 
+        # Example orphan command
+        menubar_metadata().add_command(
+            None, "Do not run", lambda: logger.error("Especially with scissors!"), ""
```
tangledhelix, line 511: I shall warn Weird Al Yankovic posthaste!

src/guiguts/misc_dialogs.py
```diff
@@ -888,6 +889,140 @@ def insert_char(self, event: tk.Event) -> None:
...
+        self.list.grid(row=1, column=0, padx=5, sticky="NSEW")
+        self.scrollbar = ttk.Scrollbar(
+            self.top_frame, orient="vertical", command=self.list.yview
+        )
+        self.scrollbar.grid(row=1, column=1, sticky="NS")
+        self.list.config(yscrollcommand=self.scrollbar.set)
+
+        # Bind events for list and entry
+        self.list.bind("<Return>", self.execute_command)
+        self.list.bind("<Double-Button-1>", self.execute_command)
```
tangledhelix, line 948: Looking at some other implementations of palettes I'm familiar with (VScode, macOS Spotlight, Alfred on macOS), they all share the common trait that clicking once chooses, not double-click. What do you think of just having click select/execute instead of double-click? If not, is it because you want to reserve click for something in the future?

tangledhelix (approved):
Looks really good! The first thing I tried wasn't there (split window) but that's because it's a checkbox item.

Did you check if there are menu items with duplicate names? What would happen if there were?

tangledhelix:
We might want to review some of the menu names (you may have noticed this already). For example, if I wanted to look for unmatched DP markup (Tools → Unmatched → DP Markup), and I search for "unmatched", the only command is "Unmatched HTML Tags" from the HTML menu. I'm not sure the best way to solve that, but it's something to consider.

src/guiguts/misc_dialogs.py
srjfoo: > Looking at some other implementations of palettes I'm familiar with (VScode, macOS Spotlight, Alfred on macOS), they all share the common trait that clicking once chooses, not double-click. What do you think of just having click select/execute instead of double-click? If not, is it because you want to reserve click for something in the future?

I thought that clicking once to choose was pretty standard across systems, and double-clicking to execute. Or am I missing something?

srjfoo: Oh ... clicking once on a button executes, though, and that's standard across platforms, too.

tangledhelix: In this case, it's a list of things for you to execute, so in my mind, choose == execute. Unless there is a future reason to select one for another purpose that @windymilla has in mind and thus it's kept in reserve for that use.

windymilla: One purpose I mentioned elsewhere (or intended to) is to add an "edit" button that allows you to edit/delete/add the shortcut for an entry.

I suspect it will be mostly used by people who are navigating it via the keyboard, so Return/Enter will execute. Or single click the entry then single click the button. Or double click the entry. Can we see if double click is too annoying, because I'd prefer to use double for execute if it's not.

src/guiguts/widgets.py
windymilla: I'll look into that

windymilla:
> Looks really good! The first thing I tried wasn't there (split window) but that's because it's a checkbox item.
I'll add the checkbox items, now that we've established it's doing vaguely the kind of thing we want.
 
> Did you check if there are menu items with duplicate names? What would happen if there were?
I didn't check. As long as the duplicate menu item is not in the same menu, e.g. two Edit-->Copy entries, I think it would be OK. The display shows which menu it's in, and nothing keys on the text label internally.

windymilla:
> We might want to review some of the menu names (you may have noticed this already). For example, if I wanted to look for unmatched DP markup (Tools → Unmatched → DP Markup), and I search for "unmatched", the only command is "Unmatched HTML Tags" from the HTML menu. I'm not sure the best way to solve that, but it's something to consider.

Two options are
1. Change all the menu buttons in the Unmatched menu to "Unmatched DP Markup", etc. I don't really like that idea, but maybe.
2. When filtering, search the menu name too. One of Chat-GPT's early attempts also searched on the shortcut, but that definitely wasn't helpful. I'll probably implement menu name search and we can see if we think it's acceptable.

windymilla:
1. Filter now also applies to menu name, so "Unmatched" will find items in the Unmatched menu.
2. Checkboxes are now included, as well as Toggle Line Numbers and Column Numbers (from Prefs dialog).
3. Classes reorganized for better efficiency - should also make it easier to do shortcut editing later.
4. Recent-most 5 commands shown at top of list in green (except selected one). Recents should respond to filter correctly.
5. "Orphan" commands added for Toggling Line numbers and Column numbers as sensible examples.

tangledhelix (approved):
The most recent 5 looks good! Just two items:
1. In dark theme the green is hard to read; might need to adjust it to a lighter green, or something.
2. Minor perhaps, but if an item is in the most-recent 5 and you search for the item it shows up twice. Perhaps cosmetic, just mentioning it in case you're concerned users may find it confusing.

I can already see more use of shadow items. I loaded up to test and my image area was blank, so I tried to turn on auto image (but it wasn't there). Then I tried to search "fit" to autosize, which also wasn't there. At some point I tried to change themes, another one that could be added.

But looks really good, in general. I'm looking forward to using this.

tangledhelix:
Your tweaks all look good here. I did run into some weird behavior though I can't explain it after some debugging. Some of the checkbox commands don't seem to work, but others do.

Working:
- Toggle Auto image
- Toggle Highlight Proofer Comments

Not working:
- Toggle Highlight HTML Tags
- Toggle Split Text Window

When I say not working, I mean that you can see them and try to execute but if you try to execute, nothing happens. I don't see why it doesn't work, though.

windymilla:
> Some of the checkbox commands don't seem to work, but others do.

I think the apparently random commands working/not working was down to faulty detection of which line was the fake separator between recents and non-recents. A command could get tagged as a separator, which then made it non-executable, depending on where it appeared in the list and how many recent commands there were. That bug is fixed, and hopefully explains the apparent failure of some commands to work.

tangledhelix (approved):
Indeed the previously/randomly not working commands now work!

They seem to sometimes run slower than I would expect (e.g. 2 seconds for highlights to change for "Highlight Proofer Comments"), but it's not consistent and I'm happy to ignore that as a future optimization task. I have some thoughts about looking at async/await someday to reduce the number of entire-app-blocking operations we have.

srjfoo:
Overall, I like it ... just a few things ...

The following screenshot is at pretty much the default width. Widening the widget will widen the `Command` column, but not the `Menu` column, so both the RTL menu description and convert fractions are truncated.

image

For that matter -- isn't RTL relevant only for Windows/Linux (at least, I don't have an RTL menu of any description)? I.e., should the items in the list be platform-specific?

Btw, when I initially went to take the screenshots, I found that every time I press the `Cmd` key, the list zeroes to the first list item (About Guiguts :D) -- was that intended?

And one thing that became annoying as I was testing -- having to bring it back up again when I wanted to use it. Mind you, that's in testing mode, so it might not be a particular issue most of the time. It's just that, if it's going to replace using a menu, and people decide they like to use it, that means `cmd/ctrl`+`shift`+`p`, plus finding the command, plus either double-clicking, select and enter, or select and run. It would be much more convenient and less work if the widget stays active until dismissed so that commands can just be selected and run. Otherwise, it's probably easier to just go to the menu or the toolbar. Except forthings that don't appear in a menu or in the toolbar.

Now that I understand more what this looks like, I realize it's not what I was thinking about as far as shortcut customization is concerned in macOS, BBEdit and Photoshop. Those are a means for users to create or reassign existing shortcuts. This appears to be a widget in its own right.

windymilla:
> Now that I understand more what this looks like, I realize it's not what I was thinking about as far as shortcut customization is concerned in macOS, BBEdit and Photoshop. Those are a means for users to create or reassign existing shortcuts. This appears to be a widget in its own right.

This isn't addressing the [shortcut customization issue](#143) (nor #762) - this is addressing the [Command Palette issue](#135).
It's possible that this dialog could be used as a basis for shortcut customization when those issues are addressed. For example, if you have a command selected, an "Edit Shortcut" button could allow you to add/edit/remove the shortcut for that command. That's all in the future though.

windymilla:
> And one thing that became annoying as I was testing -- having to bring it back up again when I wanted to use it. Mind you, that's in testing mode, so it might not be a particular issue most of the time. It's just that, if it's going to replace using a menu, and people decide they like to use it, that means `cmd/ctrl`+`shift`+`p`, plus finding the command, plus either double-clicking, select and enter, or select and run. It would be much more convenient and less work if the widget stays active until dismissed so that commands can just be selected and run. Otherwise, it's probably easier to just go to the menu or the toolbar. Except forthings that don't appear in a menu or in the toolbar.

I think that's just a testing annoyance. It's not to completely replace using menus, just to provide an alternative way to access occasional entries, especially for keyboard users, and to access hidden/shadow commands. Looking online at references to command palettes in other apps, it seems that dismissing the palette when the command is executed is normal.

To redo the command you last used, it's really just two keystrokes: Cmd+shift+p then Enter. 

If you want to run another command that you need to search for, then you're going to have to either type, e.g. "Split" to find the split screen command, or you're going to have to scroll down the list of commands to find it. Once you find the command you have to Enter or double click. Given the need for typing/scrolling & clicking, the overhead of the initial Cmd+shift+p is small, IMO.

windymilla:
> Btw, when I initially went to take the screenshots, I found that every time I press the `Cmd` key, the list zeroes to the first list item (About Guiguts :D) -- was that intended?

Not intended. It ignores non-printable keys generally as far as updating the entry field, but was always refreshing the display. Good catch - fixed in next commit.

windymilla:
> For that matter -- isn't RTL relevant only for Windows/Linux (at least, I don't have an RTL menu of any description)? I.e., should the items in the list be platform-specific?

Are you sure you don't have an RTL menu? Look in Edit-->Right-to-left Text (2 up from the bottom), and there should be "Surround selected text with RLM...LRM markers" which is the one you see in Command Palette. You shouldn't see "Paste Hebrew & spaces only (no punctuation)" which is not a button on macOS.

windymilla:
> The following screenshot is at pretty much the default width. Widening the widget will widen the `Command` column, but not the `Menu` column, so both the RTL menu description and convert fractions are truncated.

Yes, I've found with other uses of the Treeview widget (essentially a grid of labels with rows & columns) that there is limited control over the width (without what seems like excessive effort/complexity). You can specify an initial column width; you can make columns so they don't resize at all; or you let them resize in some kind of proportion, i.e. they will all grow/shrink as you adjust the overall width. The menu & shortcut columns are currently set not to resize. I set their width wide enough to display all menus & shortcuts on my computer, but you have to set the width in screen units (pixel-ish) so probably your font is a bit larger than mine. 

One thing you can do (at least I can) is put your mouse between the column headings (or at the end of the Menu heading) and click&drag to change the width. However, as you'll see if you try it, it's a bit awkward to get what you want. Also the widths will be reset next time the dialog pops (though I could save the widths as prefs & restore them if we decide it's really necessary). See this screenshot where I made menu & shortcut wider by dragging:
image

So, I think the best solution is for me to change them so that they can grow/shrink, and see if that's acceptable, because the "manual resize & store in prefs" approach is added complexity for something that is still not ideal.

windymilla:
Commit pushed with those minor changes in:
1. Cmd key shouldn't refresh the list of commands
2. All columns will grow/shrink when dialog is resized
Should be ready for retesting when you are, @srjfoo

srjfoo:
> > For that matter -- isn't RTL relevant only for Windows/Linux (at least, I don't have an RTL menu of any description)? I.e., should the items in the list be platform-specific?
> 
> Are you sure you don't have an RTL menu? Look in Edit-->Right-to-left Text (2 up from the bottom), and there should be "Surround selected text with RLM...LRM markers" which is the one you see in Command Palette. You shouldn't see "Paste Hebrew & spaces only (no punctuation)" which is not a button on macOS.

You're right -- and I was thinking "menu", not "submenu".

srjfoo (approved):

PR #873, opened 2025-04-05
Improve ebookmaker dialog

windymilla:
1. Determine default severity better, and allow calling routine to specify severity when entry is added.
2. Warnings, errors or critical errors from ebookmaker will now have the severity in red at the start of the message.
3. If user chooses to sort by Alpha/Type, sort by severity, preserving message order within severities.
4. Add Suspects Only button.
5. Warn user if they need to select the ebookmaker dir, i.e. if the dialog field is empty.

### DIFF
--- a/src/guiguts/tools/pptxt.py
+++ b/src/guiguts/tools/pptxt.py
@@ -1373,12 +1373,12 @@ def report_multiple_occurrences_on_line(
     error_start = matches[0].start()
     error_end = matches[-1].end()
     # Add record to the dialog with prefix if repeated matches.
+    repeat_str = f"(x{len(matches)}) " if len(matches) > 1 else ""
     checker_dialog.add_entry(
-        line,
+        f"{repeat_str}{line}",
         IndexRange(f"{line_number}.{error_start}", f"{line_number}.{error_end}"),
-        error_start,
-        error_end,
-        error_prefix=f"(x{len(matches)}) " if len(matches) > 1 else "",
+        error_start + len(repeat_str),
+        error_end + len(repeat_str),
     )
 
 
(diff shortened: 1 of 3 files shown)


src/guiguts/tools/pptxt.py
```diff
@@ -1373,12 +1373,12 @@ def report_multiple_occurrences_on_line(
     error_start = matches[0].start()
     error_end = matches[-1].end()
     # Add record to the dialog with prefix if repeated matches.
+    repeat_str = f"(x{len(matches)}) " if len(matches) > 1 else ""
```
tangledhelix, line 1376: I was surprised this file is involved in this PR. Is this part of the ebookmaker change?

tangledhelix (approved):
Looks good in general. Only two remarks:

1. I might consider two different colors for CRITICAL/WARNING, but it's not a major thing
2. The file's unsaved state isn't checked before ebookmaker runs; unsaved changes are not visible to ebookmaker and the user might not realize. I'm happy to see that be a separate PR, though.

srjfoo (approved):
I agree with Dan that the user needs to know that unsaved changes won't be seen by ebookmaker.

One other comment: "suspects only" and Sort options happen live. But re-run does not happen automatically once one adds the path (if it was empty to begin with), nor does it happen automatically on selecting or de-selecting Verbose. Should it? And if not, it probably needs to be documented that it doesn't.

src/guiguts/tools/pptxt.py
windymilla: Yes - it's a consequence of making severities more flexible. Previously the repeat string hijacked the error_prefix mechanism, but that was a bad/lazy decision on my part. So this edit is undoing that laziness, now that severity and error prefix can be set independently.

windymilla:
> 1. I might consider two different colors for CRITICAL/WARNING, but it's not a major thing

Decided against this. In fact, in the generic checker code, we only have error (red), info & other. In the ebookmaker code, critical, error and warning will all be displayed as checker "errors", i.e. red. If there's really a need for distinguishing, we can add more finely grained severities in the future. A user needs to deal with (or at least review) all the red messages, including ebookmaker warnings.

PR #874, opened 2025-04-05
Add pref for whether tooltips are displayed

windymilla:
Added to Advanced tab in Prefs dialog.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -353,6 +353,7 @@ class Guiguts:
             PrefKey.LEVENSHTEIN_DISTANCE, LevenshteinEditDistance.ONE
         )
         preferences.set_default(PrefKey.FOOTNOTE_INDEX_STYLE, FootnoteIndexStyle.NUMBER)
+        preferences.set_default(PrefKey.SHOW_TOOLTIPS, True)
         preferences.set_default(PrefKey.WRAP_LEFT_MARGIN, 0)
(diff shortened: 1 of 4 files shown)


tangledhelix:
Psst... once #862 is merged this could use a shadow entry for it ;)

srjfoo (approved):
Disclaimer: I did not test all widgets, but I tried for a wide variety of types (including WF), and everything looked good.

windymilla:
> Psst... once #862 is merged this could use a shadow entry for it ;)

Maybe, although it's not the kind of thing you're likely to turn on & off often. We'll have to have a careful think about what are our criteria for adding shadow entries. The more there are, the more clutter there is in the code and the command palette. On the other hand, it is annoying not to have a shadow entry for something you want to change.

tangledhelix:
> Maybe, although it's not the kind of thing you're likely to turn on & off often. We'll have to have a careful think about what are our criteria for adding shadow entries. The more there are, the more clutter there is in the code and the command palette. On the other hand, it is annoying not to have a shadow entry for something you want to change.

There is a level of annoyance / maintenance on our side for having too many. Users will (I believe) just search and how many are there don't matter as much. Consider how many are in vscode (hundreds?) yet it remains usable. I suppose if there are some terms that are too generic and would result in not-refined-enough searches that could be a hassle. "git" might be a good vscode example of that, it has a lot of matches, especially if you have any git-related plugins installed.

PR #875, opened 2025-04-05
Improve peer cursor line highlighting

### LINKED ISSUE #863: Split text view cursor line highlight bug
1. Type in a short file:
```
1
2
3
4
5
```

2. Split the text window
3. Position cursor in line 3 in the top half
4. Position cursor in line 5 in the bottom half
5. Depending on which half has focus (use Tab key to switch between halves), the cursor line highlighting jumps from line 3 to line 5 _**in both top and bottom half**_

Problem is caused by the fact that tags (except for the `sel` tag) are shared between maintext and the peer.
I don't see a solution to this (which showed up when using split view for footnotes)

windymilla:
Highlight both the cursor line in maintext and the cursor line in peer. This is an improvement on just one of the cursor lines being highlighted.

Fixes #863

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -375,6 +375,7 @@ class HighlightTag(StrEnum):
     CURLY_SINGLE_QUOTE = auto()
     ALIGNCOL = auto()
     CURSOR_LINE = auto()
+    CURSOR_LINE_PEER = auto()
     COLUMN_RULER = auto()
     SEARCH = auto()
     PROOFERCOMMENT = auto()
@@ -554,6 +555,11 @@ class HighlightColors:
     }
 
     CURSOR_LINE = {
+        "Light": {"background": "#E8E1DC"},
+        "Dark": {"background": "#122E57"},
+    }
+
+    CURSOR_LINE_PEER = {
         "Light": {"background": "#E8E1DC"},
         "Dark": {"background": "#122E57"},
     }
@@ -3704,14 +3710,18 @@ class MainText(tk.Text):
     def highlight_cursor_line(self) -> None:
         """Add a highlight to entire line cursor is focused on."""
         self.tag_remove(HighlightTag.CURSOR_LINE, "1.0", tk.END)
+        self.tag_remove(HighlightTag.CURSOR_LINE_PEER, "1.0", tk.END)
 
         # Don't re-highlight if cursor line highlighting disabled, or if currently a selection
-        if (
-            preferences.get(PrefKey.HIGHLIGHT_CURSOR_LINE)
-            and not self.selected_ranges()
-        ):
-            row = self.get_insert_index().row
-            self.tag_add(HighlightTag.CURSOR_LINE, f"{row}.0", f"{row + 1}.0")
+        if preferences.get(PrefKey.HIGHLIGHT_CURSOR_LINE):
+            if not self.tag_ranges("sel"):
+                row = IndexRowCol(self.index(tk.INSERT)).row
+                self.tag_add(HighlightTag.CURSOR_LINE, f"{row}.0", f"{row + 1}.0")
+            if preferences.get(PrefKey.SPLIT_TEXT_WINDOW) and not self.peer.tag_ranges(
+                "sel"
+            ):
+                row = IndexRowCol(self.peer.index(tk.INSERT)).row
+                self.tag_add(HighlightTag.CURSOR_LINE_PEER, f"{row}.0", f"{row + 1}.0")
 
     def highlight_search_in_viewport(self, viewport: Text) -> None:
         """Highlights current search pattern in designated viewport."""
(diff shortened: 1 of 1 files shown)


windymilla:
See linked issue for how to reproduce bug. With this fix, screenshot looks like this:
image

windymilla:
One possibility would be to have a different color/brightness highlight for the two windows, e.g. peer cursor line is a darker blue, or a purple. Not sure I like the idea, but would be interested in @tangledhelix & @srjfoo opinions.

tangledhelix:
Counter-proposal: have two highlight colors (active/inactive) and color based on which pane is active at the moment.

Screenshot 2025-04-05 at 9 04 25 PM
Screenshot 2025-04-05 at 9 04 43 PM

Probably easier to assess by running it; I have it on a branch `multiple-cursorline` in my fork. To test, just split the text window, put cursor on two different lines that are both visible, and tab to jump between them.

If this looks good to you, the highlight name could probably be changed since "peer" has a meaning in this codebase; maybe it should become `CURSOR_LINE_ACTIVE` and `_INACTIVE` instead.

The colors could probably use some refinement. The light one doesn't work particularly well in high-contrast.

srjfoo:
Here's what the same thing looks like in BBEdit.

image

And only if the line is selected by clicking on the line number.

If you just click in the line, this is what you get:

image

I'm not saying that we should use that model: I don't object to the other lines being highlighted, but to me, one thing that will tell me visually very quickly which is the active pane is the highlighting. The active line in the active pane should be the most noticeable. The greyed-out hightlighting on the line number in the inactive pane will show which has the cursor for that pane in it (i.e. where the cursor will be if you tab between panes). But I'd rather see only the line with the cursor in the active pane have the most noticeable highlighting, if that makes sense.

@tangledhelix -- I tried your branch, but I must be missing something. What I'm seeing is the same as @windymilla's branch, and does not match your screenshot.

tangledhelix:
> @tangledhelix -- I tried your branch, but I must be missing something. What I'm seeing is the same as @windymilla's branch, and does not match your screenshot.

It would help if my branch had my code on it. I pushed it WITH the new code this time.

tangledhelix:
This mechanism could just as easily only highlight the active pane and not the other, with a trivial change. The main change from @windymilla's code is that it highlights based on which pane is active rather than which is the main vs. peer (top/bottom).

tangledhelix:
> This mechanism could just as easily only highlight the active pane and not the other, with a trivial change. The main change from @windymilla's code is that it highlights based on which pane is active rather than which is the main vs. peer (top/bottom).

... although that scenario is exactly what the PR aimed to do away with, so that doesn't make much sense.

srjfoo:
@windymilla -- I like the concept, but definitely prefer the muted highlighting for the non-active pane.

windymilla:
> Here's what the same thing looks like in BBEdit.
> ...
> And only if the line is selected by clicking on the line number.

That's already pretty much what happens in `master` & the PR if you click on the line number.

> If you just click in the line, this is what you get:
> ...
> I'm not saying that we should use that model:

That is what we would like, but it just isn't possible with tkinter.

>  But I'd rather see only the line with the cursor in the active pane have the most noticeable highlighting, if that makes sense.

If I understand you correctly, that's also not possible, and is the whole problem this PR is attempting too mitigate. The highlighting of line 5 (for example) looks the same in the two panes. You can't have it show more noticeably in the active pane and less so in the inactive pane. The only kind of highlighting that can look different in the two panes is the "sel" tag (which is normal text selection). GIven those restrictions, Dan's solution is definitely an improvment, i.e. making the highlighting (in both panes) that relates to the "current line number in the inactive pane" being less noticeable than the highlighting (in both panes) that relates to the "current line number in the active pane". (Use of quotes to clarify that "in the in/active pane" is qualifying "current line number", rather than where the highlighting appears).

windymilla:
@tangledhelix - do you want to put up a new PR with your version. I'm happy with the colors, and am not too concerned about them. 

We also need to remember, that we are developing/testing the fix, not using it for real, so we are deliberately constructing the worst-looking case of having both the active current line and inactive current line visible in both viewports at the same time, and close to one another, and we are looking at both viewpoints.

IRL, it won't be very common to have the same bit of the file visible in the two viewports, because you don't need split screen to look at the _**same**_ bit of the file twice - it's for looking at different bits. Also in general, IRL your eye will be looking more at the active viewport, because that's where you're able to actually make changes, so the more noticeable current line highlight there will be the most relevant one.

In other words, I don't think we need to get too hung up about it, but your version is definitely an improvement on mine.

srjfoo:
>[...] so we are deliberately constructing the worst-looking case of having both the active current line and inactive current line visible in both viewports at the same time, and close to one another, and we are looking at both viewpoints.

Yeah, I was getting too hung up on the testing scenario. I'm fine with @tangledhelix's version.

tangledhelix:
> @tangledhelix - do you want to put up a new PR with your version.

Done, in #878

windymilla:
Superseded by #878

PR #878, opened 2025-04-07
Implement active/inactive pane cursorline

### LINKED ISSUE #863: Split text view cursor line highlight bug

tangledhelix:
Building on the previous work in #875, separate `CURSOR_LINE` highlight into `CURSOR_LINE_ACTIVE` and `_INACTIVE`. The highlights will be swapped between panes depending on which is currently active, rather than using "main" vs "peer" which in this project means essentially "top" vs "bottom" pane.

The "active" highlight is meant to be more prominent than the "inactive". In the case where the viewport is covering the same small area in both panes, the highlights will co-exist and their being different will help to differentiate which is which within each pane.

The "inactive" highlight is lower priority than the "active" so if they cover the same line, the active takes precedence. This means in the pathological case where the cursor is on the same line in both panes, both panes will appear to have the active cursor line. There's no workaround for that edge case (so far as I know).

Fixes #863

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -374,8 +374,8 @@ class HighlightTag(StrEnum):
     STRAIGHT_SINGLE_QUOTE = auto()
     CURLY_SINGLE_QUOTE = auto()
     ALIGNCOL = auto()
-    CURSOR_LINE = auto()
-    CURSOR_LINE_PEER = auto()
+    CURSOR_LINE_ACTIVE = auto()
+    CURSOR_LINE_INACTIVE = auto()
     COLUMN_RULER = auto()
     SEARCH = auto()
     PROOFERCOMMENT = auto()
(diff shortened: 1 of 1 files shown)


tangledhelix:
This is largely the same as the code presented previously, although I cleaned it up slightly (e.g. renaming the highlights to clearer names, etc). It probably doesn't need rigorous re-testing but a quick pass couldn't hurt.

srjfoo:
This is what I see with a light theme on OS dark mode:
image

I thought the last time I looked it looked okay, but this doesn't.

tangledhelix:
The column ruler I can't explain; that doesn't happen for me.

As we discussed, you didn't click into that line, you clicked it in the line number gutter. As such, that line (line 8) is selected, and what you're seeing is that in OS dark mode, the "sel" coloring is using the dark mode colors, not the light mode color.

And line 8 is not highlighted in the top pane because cursorline highlights are not applied when a selection is active. If you click into line 8 instead of select it, then it should look more like what my screenshots show. (Except the column ruler.)

The cursorline not highlighting when a selection is active is by design, not a bug. In that case, we have previously decided that the entire selection is the "cursor" and there is no cursorline in that case. Remember the selection may span multiple lines, and it may be a columnar selection; so we highlight no cursorline at all while select is active.

This color combination is not great IMO. I think it argues for specifying the line number gutter color rather than using the selection colors.

I would even suggest we specify the selection color, as that selection color in light mode is a bit jarring.

srjfoo (approved):

srjfoo (approved):
Once I got it straight that I wasn't supposed to be testing for the line-number highlighting, everything looked great. I couldn't repeat the weird ruler coloring issue.

PR #883, opened 2025-04-08
Add HTML Auto-Table

### LINKED ISSUE #449: Add HTML Auto table feature

windymilla:
In the HTML menu, pops a dialog which behaves
much like the Auto-Table controls in the GG1 HTML
Markup dialog.

1. Select a table that is formatted for text (with or without vertical lines)
2. Check whether table is one line per row, or it is "multiline" with blank lines separating rows.
3. Set default column alignments.
4. Specify column alignments not all the same, e.g. "LCR" will  make the first 3 columns left, center, right aligned, and any more cols will use the default.
5. Click the Convert button.

Fixes #449

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -22,6 +22,7 @@ from guiguts.html_tools import (
     HTMLLinkChecker,
     CSSValidator,
     EbookmakerChecker,
+    HTMLAutoTableDialog,
 )
 from guiguts.illo_sn_fixup import illosn_check
 
@@ -517,6 +518,10 @@ class Guiguts:
         preferences.set_default(PrefKey.ASCII_TABLE_FILL_CHAR, "@")
         preferences.set_default(PrefKey.ASCII_TABLE_RIGHT_COL, 70)
         preferences.set_default(PrefKey.COMMAND_PALETTE_HISTORY, [])
+        preferences.set_default(PrefKey.AUTOTABLE_MULTILINE, False)
+        preferences.set_default(PrefKey.AUTOTABLE_DEFAULT_ALIGNMENT, "left")
+        preferences.set_default(PrefKey.AUTOTABLE_COLUMN_ALIGNMENT, "")
+        preferences.set_default(PrefKey.AUTOTABLE_COLUMN_ALIGNMENT_HISTORY, [])
 
         # Check all preferences have a default
         for pref_key in PrefKey:
@@ -902,6 +907,7 @@ class Guiguts:
         html_menu = Menu(menubar(), "HT~ML")
         html_menu.add_button("HTML ~Generator...", HTMLGeneratorDialog.show_dialog)
         html_menu.add_button("Auto-~Illustrations...", HTMLImageDialog.show_dialog)
+        html_menu.add_button("Auto-~Table...", HTMLAutoTableDialog.show_dialog)
         html_menu.add_separator()
         html_menu.add_button("~Unmatched HTML Tags", unmatched_html_markup)
         html_menu.add_button("HTML ~Link Checker", lambda: HTMLLinkChecker().run())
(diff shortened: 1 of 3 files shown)


srjfoo:
For me, the "`F1`" keyboard shortcut brings up the correct manual page; Choosing "Guiguts Manual" from the menu does not.

I chose as my test table the one from the ASCII tutorial that has all the numbers. I finally figured out why my table cells don't appear to be right aligned -- because I don't have any CSS to match the classes (maybe tomorrow). However, reading the code, that part looks good. What jumped out at me, though, was that all of the non-breaking spaces were stripped.

The other thing, and this may just be because of order of operations, and me misunderstanding the instructions in the tutorial, is that the blank lines in between were not completely blank -- they had spaces and stiles. The stiles were stripped from the table, but were all printed out in a single line above the table.
image

windymilla:
> The other thing, and this may just be because of order of operations, and me misunderstanding the instructions in the tutorial, is that the blank lines in between were not completely blank -- they had spaces and stiles. The stiles were stripped from the table, but were all printed out in a single line above the table.

That looks like a bug. If I can't reproduce it, I'll ask you for instructions/example file.

windymilla:
@srjfoo - I couldn't reproduce that. Please could you give me an example & instructions - thanks

srjfoo:
> @srjfoo - I couldn't reproduce that. Please could you give me an example & instructions - thanks

I re-did it from first concepts. This time I added the CSS to see what the HTML would look like.

This time the extra vertical bars were removed, so I must've gotten something mixed up the previous time (that's a relief). However, the non-breaking spaces are still being removed, so there is nothing to cause the numbers to be aligned properly (or close thereunto). One row sample of the HTML generated:

```
<tr>
<td class="tdl"><i>Nonpareil</i></td>
<td class="tdc">85</td>
<td class="tdc">15</td>
<td class="tdc">29</td>
<td class="tdc">5</td>
<td class="tdc">11.7</td>
</tr>
```

Also, it encoded the empty rows:

```
<td class="tdl"></td>
<td class="tdc"></td>
<td class="tdc"></td>
<td class="tdc"></td>
<td class="tdc"></td>
<td class="tdc"></td>
</tr>
```
I'm using this section of the tutorial, starting with this text:

```
+------------------+---------+--------+-------+--------+---------+
|                  |Length of|Depth of| Rake  |Rake aft|Weight of|
|                  |  Keel   |  hold  |forward|        |masts and|
|                  |         |        |       |        |  yards  |
|                  +---------+--------+-------+--------+---------+
|                  |  feet   |  feet  |  feet |  feet  |ton. cwt.|
|<i>Elizabeth</i>       |   100   |   18   |  36   |  6     |  22.8   |
|<i>Waite Bear</i>      |   110   |   18   |  36   |  6.6   |  24     |
|<i>Merkonour</i>       |   110   |   17   |  37   |  6.6   |  22.13  |
|<i>Ark Royal</i>       |   100   |   15   |  33.6 |  6     |  18.4   |
|<i>Victory</i>         |    95   |   17   |  32   |  5.10  |  18.4   |
|<i>Repulse</i>         |   105   |   16   |  --   |  --    |  20.7   |
|<i>Lion</i>            |   100   |   14   |  31.6 |  5.10½ |  14.14  |
|<i>Nonpareil</i>       |    85   |   15   |  29   |  5     |  11.7   |
|<i>Defiance</i>        |    92   |   15   |  31   |  5.6   |  14.9   |
|<i>Sonne</i>           |    50   |    6   |  11   |  2.2   |   1.2   |
+------------------+---------+--------+-------+--------+---------+
```

And following the instructions to the point where it starts talking about the `HTML->Markup` dialog. At which point, I used `Auto-Table...`. As far as I can tell, the only problem is that the non-breaking spaces are being removed. Am I supposed to leave the replacement characters in place until after I generate the HTML?

srjfoo:
The non-breaking spaces are no longer being stripped 🎉 , but I'm still getting empty rows between number rows.

srjfoo (approved):
🤦‍♀️ -- I finally figured out what I was doing wrong. I didn't have the multiline checkbox checked, and so of course the empty lines were being retained.

PR #884, opened 2025-04-08
Backup Prefs file

### LINKED ISSUE #882: Add feature to auto-backup GGprefs
Suggestion:

1. If there is no GGprefs_week file that is less that 1 week old, rename existing GGprefs_day file to GGprefs_week.
2. If there is no GGprefs_day file that is less that 1 day old, rename existing GGprefs file to GGprefs_day
3. Save GGprefs file as currently.

Reasoning: 

More and more stuff is going in the GGprefs file, and it would be annoying if the user were to lose it, through power outage, code bug, etc. Just doing a `.bk1`, `.bk2` things like file backups is not satisfactory because GGprefs get saved a lot without an explicit user decision. Even restarting the program after a crash might cause GGprefs to be saved. So simple backups of that type may quickly get overwritten before the user has blinked. Saving up to a day-old and a week-old version improves the chances the user will notice before backups are overwritten.

windymilla:
Keep a "day" backup and a "week" backup. These
only get overwritten if they age past their time
period.

Fixes #882

Testing notes:
You may find variants on the following command useful if you do not want to wait for a week to test this:
`touch -m -t 04082015 GGprefs_day.json`
Also, one way to force a save of GGprefs is to load a different file, e.g. from the Recent files menu. Then you can edit the GGprefs file and the two backups and inspect which is the top file in the "recent_files" list. That way you can keep track of when one file is renamed to another.

### DIFF
--- a/src/guiguts/preferences.py
+++ b/src/guiguts/preferences.py
@@ -5,6 +5,7 @@ from enum import StrEnum, auto
 import json
 import logging
 import os
+import time
 import tkinter as tk
 from typing import Any, Callable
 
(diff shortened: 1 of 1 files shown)


srjfoo:
> touch -m -t 04082015 GGprefs_day.json

Actually, on macOS, this created a file with the timestamp of April 8, 2025: macOS, anyway, uses the `[[CC]YY]MMDDhhmm[.SS]` format for `-m -t`. (According to the manual.) 😁

srjfoo (approved):

windymilla:
> > touch -m -t 04082015 GGprefs_day.json
> 
> Actually, on macOS, this created a file with the timestamp of April 8, 2025: macOS, anyway, uses the `[[CC]YY]MMDDhhmm[.SS]` format for `-m -t`. (According to the manual.) 😁

Yes, that's the same for me. That example touch command was to create a file with a time stamp of 8.15pm, which is when I was doing my testing.

srjfoo:
> > > touch -m -t 04082015 GGprefs_day.json
> > 
> > 
> > Actually, on macOS, this created a file with the timestamp of April 8, 2025: macOS, anyway, uses the `[[CC]YY]MMDDhhmm[.SS]` format for `-m -t`. (According to the manual.) 😁
> 
> Yes, that's the same for me. That example touch command was to create a file with a time stamp of 8.15pm, which is when I was doing my testing.

Aaah -- I totally misunderstood. :D

I used it to create an "over a week old" file and an "over a day old" file.

PR #889, opened 2025-04-09
Fix WF bug when underscore follows `7NaAlSiO`

### LINKED ISSUE #887: Word Frequency Alpha/num left click does not find some matches
The Word Frequency Alpha/num check does not find matches for some chemical formulas such as 9WoO, 7NaAlSiO, and 6H. If underscores are the reason for this lack of matching, this unusual behavior should be documented in the GG2 user manual Word Frequency Alpha/num section.

Image

Project file:
[minerals.zip](https://github.com/user-attachments/files/19657389/minerals.zip)

windymilla:
Underscore is no longer treated as a word character, which it was, inconsistently.

Also, remove `001.png` & similar from word list, mostly for ppgen users.

Fixes #887

### DIFF
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -121,6 +121,8 @@ class WFWordLists:
                 word = re.sub(r"(?<!\-)\*", "", word)  # * not preceded by hyphen
                 word = re.sub(r"^\*$", "", word)  # just *
                 word = strip_punc(word)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: Make sure WF still seems to be behaving as expected, particularly when you click a word to jump to it. The edit involved a complete change of routine used for this.

srjfoo (approved):

PR #893, opened 2025-04-10
Add fuzzy matching to command palette

windymilla:
Algorithm:
1. Get score based on quality of match. If button label begins with search string, score=100; if button label contains search string, score =90; if menu label begins with search string, score =80; otherwise fuzzy match "menu+button" labels.
2. If command is a recently used command only put it in the "recent" section at the top if score is > 80, i.e. got a proper match not a fuzzy match. Without this, even when user types accurate search string for a non-recent command, it appears below all the recents. Seems to be roughly how VScode works.


srjfoo (approved):
Works well now :D 

One suggestion. Currently, the list seems to be sorted in ASCII order. That is, `lower case selection` is not alphabetized with the other `l` commands. For that matter, `UPPER CASE SELECTION` is out of order, too, for the same reason. Would it make more sense to do a case-insensitive sort? (Or, contrarily, make both commands title case so that it's not an issue.)

tangledhelix (approved):
I'll have to get a feel for the right patterns to input (`auim` brought up See Image before Auto Image, which I found odd), but once adapted it'll work great, I think.

windymilla:
> Works well now :D
> 
> One suggestion. Currently, the list seems to be sorted in ASCII order. That is, `lower case selection` is not alphabetized with the other `l` commands. For that matter, `UPPER CASE SELECTION` is out of order, too, for the same reason. Would it make more sense to do a case-insensitive sort? (Or, contrarily, make both commands title case so that it's not an issue.)

Yes, it was working like that, then I broke it with the last recent-ness "improvements". I'll fix that.

PR #895, opened 2025-04-11
Add more shadow commands

windymilla:
1. All Prefs dialog toggles added
2. Image viewer Auto-fit toggles added
3. Change theme buttons added
4. "Toggle" before label in command palette list changed to "(toggle)" after label to avoid having lots of commands beginning with Toggle alphabetically, and assist with user-filtering of commands.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -87,7 +87,6 @@ from guiguts.utilities import is_mac, is_windows, is_x11, folder_dir_str
 from guiguts.widgets import (
     themed_style,
     theme_name_internal_from_user,
-    menubar_metadata,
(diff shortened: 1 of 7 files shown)


tangledhelix (approved):
Does what it says on the tin... it is oddly slow, but I can't tell if it's GG or my computer... some other apps seem a bit laggy and slow right now too.

srjfoo (approved):
Suggested elsewhere, but probably fits better here: `UPPER CASE SELECTION` and `lower case selection` are sorted case sensitively, and are thus in unexpected places. Perhaps change to case insensitive sort, or make them title case instead of all UPPER/lower?

PR #896, opened 2025-04-11
Add buttons to Proofer Comment Checker

### LINKED ISSUE #866: Proofer Comments hover text actions need a delete function
Currently the `Search=>Find Proofer Comments` hover text only provides three options:

```
Left click: Select & find comment
Right click: Hide comment
Right click: Also hide all matching comments
```

[That last one is confusing.  Should that be a Control Right click or an Alt Right click?]

We need the ability to easily remove comments, both individually and in bulk. Imagine an author consistently spells 'Shakespeare' as 'Shakespere' and a prrofer added the note '[** Shakespeare]' to each instance. Please add two options:

-     'Delete this comment' to remove a single comment.
-     'Delete all matching comments to remove all comments containing that specific text.

windymilla:
Buttons delete the proofer comment highlighted by
the selected message, or all matching comments
and optionally hiding the message(s).

Also fixed minor errors in tooltips.

Fixes #866

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -62,7 +62,7 @@ from guiguts.misc_tools import (
     FractionConvertType,
     fraction_convert,
     unicode_normalize,
-    proofer_comment_check,
+    ProoferCommentChecker,
     asterisk_check,
     TextMarkupConvertorDialog,
     stealth_scannos,
@@ -710,7 +710,7 @@ class Guiguts:
         search_menu.add_separator()
         search_menu.add_button(
             "Find Proofer ~Comments",
-            proofer_comment_check,
+            lambda: ProoferCommentChecker().run(),
         )
         search_menu.add_button(
             "Find ~Asterisks w/o Slash",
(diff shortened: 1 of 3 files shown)


srjfoo (approved):
Minor nit -- why "delete & hide" -- if the comment is deleted, what's available to hide? Rerun will not bring it back (btw, I do like this!)

windymilla:
To contrast it with just deleting, which is Cmd left, that doesn't hide the message

srjfoo:
> To contrast it with just deleting, which is Cmd left, that doesn't hide the message

Okay, I'll ask the stupid question -- why not?

If you hit rerun, the message does not reappear. If you want to keep the message to do a final run-through to make sure that the comments are gone, you either don't want to hide it, or you'll do a re-run to see just the remaining comments (in which case the messages will be gone).

windymilla:
One reason is that when the current message is deleted, it displays the next one. You may want to see the fix happen with the current message, then delete it, i.e. Delete, then Hide in two operations. I definitely do that for some fixes.

PR #911, opened 2025-04-13
Add image short-name label to image viewer control strip

### LINKED ISSUE #910: Add PNG name to image viewer
When navigating among images in the image viewer, there's no indication (except perhaps a page number as seen in the image) of where you are. Add a label to the navigation bar, perhaps just to the right of the next/previous buttons, to indicate which png is being displayed.

When navigating away from the cursor position in auto-image mode, also update this when the image is reverted to auto-image on leaving the image pane.

I expressly avoid updating the text window's status bar (where the image is also indicated), because that indicator is a) not part of the image viewer UI and not intuitive to the user and b) meant to indicate the image at the current text cursor position, which is a different thing than which image is being shown in the viewer at the moment.

tangledhelix:
Rather than let the user guess which image is being displayed, show it to them. The path leading up to the file, as well as the file extension, are stripped to create a short label (i.e. "001.png" becomes "001", and "washington.jpg" becomes "washington"). The label width adjusts dynamically to the label length.

When no image is loaded, the text `<no image>` is used as a placeholder.

Fixes #910

Testing notes:

Use image viewer as normal; the label should update to match what is displayed.
- When no image is loaded, the label should say `<no image>`
- The "short name" of the image is used as the label: no path, no file extension
- Should work as expected in Auto Image mode
- Should work if you mouse into the image viewer and navigate with the next/prev buttons
- Should update correctly if you do the above, in Auto Image mode, and then mouse out of the pane, triggering a reset to the Auto Image
- Should reset to `<no image>` if you close the currently open file
- Should display `<no image>` if, on startup, Image Viewer is open, but no text/html file is loaded

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -11,6 +11,7 @@ import time
 import tkinter as tk
 from tkinter import ttk, messagebox, EventType
 from typing import Any, Callable, Optional
+from pathlib import Path
 
 from PIL import Image, ImageTk, ImageChops
 import regex as re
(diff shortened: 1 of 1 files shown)


windymilla (approved):
I'm happy with that, thanks

srjfoo (approved):

PR #914, opened 2025-04-13
Add ebookmaker using API

windymilla:
Uses in-development API - not to be merged like this.

### DIFF
--- a/src/guiguts/html_tools.py
+++ b/src/guiguts/html_tools.py
@@ -1121,7 +1121,7 @@ class EbookmakerCheckerDialog(CheckerDialog):
             api: Set to True if API is to be used.
         """
         super().__init__(
-            "Ebookmaker Results",
+            f"Ebookmaker {'(online)' if api else '(local)'} Results",
             tooltip="\n".join(
                 [
                     "Left click: Select message",
@@ -1257,8 +1257,7 @@ class EbookmakerChecker:
             command.append("-v")
         command.append("--max-depth=3")
         # Build options
-        if preferences.get(PrefKey.EBOOKMAKER_EPUB2):
-            command.append("--make=epub.images")
+        command.append("--make=epub.images")  # Epub2 is always turned on
         if preferences.get(PrefKey.EBOOKMAKER_EPUB3):
             command.append("--make=epub3.images")
         # Need Calibre to create kindle versions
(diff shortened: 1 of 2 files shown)


src/guiguts/html_tools.py
```diff
@@ -1319,22 +1321,236 @@ def run(self) -> None:
...
+        try:
+            response.raise_for_status()
+        except (
+            requests.exceptions.HTTPError,
+            requests.exceptions.TooManyRedirects,
+        ) as exc:
+            report_exception("Download request for output.txt was unsuccessful.", exc)
+            return
+
+        # Process output.txt into messages in dialog
```
cpeel, line 1473: Probably overkill, but if this logic is the same in both local and API (and I think it is) you could put it in a function to share it (possible even a new parent class that the local and API classes inherit from).

src/guiguts/html_tools.py
```diff
@@ -1319,22 +1321,236 @@ def run(self) -> None:
...
+        # Create list of formats to build
+        build = []
+        if preferences.get(PrefKey.EBOOKMAKER_EPUB2):
+            build.append("epub")
+        if preferences.get(PrefKey.EBOOKMAKER_EPUB3):
+            build.append("epub3")
+        if preferences.get(PrefKey.EBOOKMAKER_KINDLE):
+            build.append("kindle")
+        if preferences.get(PrefKey.EBOOKMAKER_KF8):
+            build.append("kf8")
```
cpeel, line 1379: Might want to check that `build` isn't empty here if none of the prefs were set. I think the API will gladly let you ask for no `--make` inputs and pass that into ebookmaker which will probably succeed with an error message. Probably better to check that here first.

src/guiguts/html_tools.py
```diff
@@ -1319,22 +1321,236 @@ def run(self) -> None:
...
+
+            buffer.seek(0)
+            return buffer
+
+        # Post zip file to ebookmaker
+        fileobj = zip_html_and_images(file_name, os.path.join(proj_dir, "images"))
+        try:
+            response = requests.post(
+                url,
+                files={"upfile1": ("{file_base}.zip", fileobj)},
```
cpeel, line 1425: Please update with the new field names (`file`, `ebook_number`, `title`).

tangledhelix (approved):

src/guiguts/html_tools.py
windymilla: I fully meant to extract that, but forgot - good catch!

src/guiguts/html_tools.py
windymilla: In fact, in GG2, the Epub2 checkbox is turned on and disabled, so is always built. I'll initialize the array to reflect that.

cpeel (approved):
This looks good to me aside from the URL pending the service update.

srjfoo (approved):

PR #919, opened 2025-04-14
Report save errors more gracefully

### LINKED ISSUE #908: Failure to save file is not graceful
Errors such as a file being locked by another process can cause an exception rather than a graceful error message

Image
Image

windymilla:
Fixes #908

Report via error message rather than allowing exception.

Testing notes:
On Windows, if another application had the text file open, GG could not rename it to ".bak", which caused an exception.
On other systems, you could probably simulate the kind of problems that might occur by disabling write access on the project directory: (`chmod -w proj_dir` ?)

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -303,29 +303,50 @@ class File:
             return f"{self.filename}{ext}", f"{bin_name(self.filename)}{ext}"
 
         binfile_name = bin_name(self.filename)
-        if autosave:
-            # Don't autosave if nothing changed - just reschedule
-            if not maintext().is_modified():
-                self.reset_autosave()
-                return ""
-            backup2_file, backup2_bin = get_backup_names(".bk2")
-            backup1_file, backup1_bin = get_backup_names(".bk1")
-            if os.path.exists(backup1_file):
-                os.replace(backup1_file, backup2_file)
-            if os.path.exists(backup1_bin):
-                os.replace(backup1_bin, backup2_bin)
-            if os.path.exists(self.filename):
-                os.replace(self.filename, backup1_file)
-            if os.path.exists(binfile_name):
-                os.replace(binfile_name, backup1_bin)
-        elif preferences.get(PrefKey.BACKUPS_ENABLED):
-            backup_file, backup_bin = get_backup_names(".bak")
-            if os.path.exists(self.filename):
-                os.replace(self.filename, backup_file)
-            if os.path.exists(binfile_name):
-                os.replace(binfile_name, backup_bin)
-        maintext().do_save(self.filename)
-        self.save_bin(self.filename)
+        try:
+            if autosave:
+                # Don't autosave if nothing changed - just reschedule
+                if not maintext().is_modified():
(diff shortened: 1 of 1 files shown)


srjfoo:
> On other systems, you could probably simulate the kind of problems that might occur by disabling write access on the project directory: (`chmod -w proj_dir` ?)

That didn't work for me on macOS -- after all, the files inside the directory are writeable.

BBEdit cheerfully refreshed when something was saved in GG2 or in TextEdit. The only time I got any reaction from GG2 was when I reloaded a file that was saved in a different application, and that was just the warning about the page locations. TextEdit popped up a warning that the file had been saved in a different app, but BBEdit did not.

@tangledhelix would be better able to test this with VSCode, and @G4OEU and/or @rtonsing on Windows if you want confirmation on Windows.

The code looks good to me, but I realized that since I ordinarily have auto-save off, I didn't test that part of the code. I'll wor on that tomorrow.

windymilla:
I'm happy with the Windows behavior. 

How about this as a way to "lock" a file on macOS @srjfoo?

> To lock a file in macOS Finder, right-click (or two-finger tap) on the file or folder, select "Get Info," and then check the "Locked" box under the General section. This prevents the file from being modified or deleted. To unlock it, simply uncheck the "Locked" box. 

or [there's a command chflags](https://ss64.com/mac/chflags.html) so you could do something like this to lock a file:

`chflags uchg myfile.txt`

and this to unlock it:

`chflags nouchg myfile.txt`

Once you've locked a file, you should be unable to save an update to it from GG2, but you should not get a traceback.

srjfoo:
> Once you've locked a file, you should be unable to save an update to it from GG2, but you should not get a traceback.

In this branch, you mean, but get the traceback in `master`?

windymilla:
Yes, this PR should fix the traceback you would get in master

srjfoo:
> I'm happy with the Windows behavior.
> 
> How about this as a way to "lock" a file on macOS @srjfoo?
> 
> > To lock a file in macOS Finder, right-click (or two-finger tap) on the file or folder, select "Get Info," and then check the "Locked" box under the General section. This prevents the file from being modified or deleted. To unlock it, simply uncheck the "Locked" box.

As far as I could tell, "Get Info" said it was already locked. At least, the little lock was closed.


> or [there's a command chflags](https://ss64.com/mac/chflags.html) so you could do something like this to lock a file:
> 
> `chflags uchg myfile.txt`
> 
> and this to unlock it:
> 
> `chflags nouchg myfile.txt`
> 
> Once you've locked a file, you should be unable to save an update to it from GG2, but you should not get a traceback.

But this worked, so, TIL. It didn't throw the same errors that you got (in `master`), but they were sufficiently similar that in this branch, the proper error messages were displayed.

One comment, though--the actual human-readable error message almost disappears into the partial traceback that's provided. Is there a way of making it more noticeable? Or does that even matter?

srjfoo (approved):

windymilla:
Thanks, I think it's OK for a very rare circumstance

PR #923, opened 2025-04-14
Allow user to select leading vertical table line

### LINKED ISSUE #918: ASCII Table Effects - Select Prev. Line/Select Next Line fails to cycle through all table dividers on a line
```
| abc| def|
| ghi| jkl|
```

If the simple table above is selected then when either Select Prev. Line or Select Next Line buttons are used only the second and third stile lines are selected and highlighted. The stile line to the left of the 'a' and the 'g' is not selected/highlighted.

The same table in GG1.6.3 shows the correct behaviour. Earlier versions of these buttons in GG2 also behaved correctly.

windymilla:
If table has a leading vertical line, i.e. left hand border, it is now possible to select it and move it to the right, then back to the left.

Fixes #918

### DIFF
--- a/src/guiguts/ascii_tables.py
+++ b/src/guiguts/ascii_tables.py
@@ -950,9 +950,7 @@ class ASCIITableDialog(ToplevelDialog):
             ) or maintext().compare(insert_rowcol.index(), ">=", self.end_mark_name):
                 self.selected_column = 0
             else:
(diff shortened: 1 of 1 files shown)


windymilla:
Latest commit should fix #930 & #931

srjfoo (approved):

Issue #956, opened 2025-04-24
Further shortcut improvements


windymilla:
A place to add thoughts about things to add to user-defined shortcut work

1. Set a single shortcut back to its default (not sure this is worth the extra complexity for very little use). Better for user to set it back manually, IMO, since it needs to also take into account that the default may have been re-used for some other action. For example, Ctrl+O is Open File by default. User assigns that to "Toggle bell". User assigns "Ctrl+K" to Open File. Then user wants to restore Open File to its default, but that would clash with their Toggle bell shortcut. Would also need code complication, storing defaults as well as current settings. One advantage would be that the Full Reset  operation wouldn't require a restart of GG, but that's a pretty minor/rare thing.
2. Add "Replace" as a shadow command. Although "Replace & Search" has a shortcut already, it's not implemented as a shadow command, so user can't modify the shortcut. Investigate if that's possible/easy. Would need to allow "Return" as a base key.
3. Consider compound shortcuts, e.g. Ctrl-U, Ctrl-X. Probably too complex for the UI & dangers of things not working to allow general setup of compound shortcuts, but could allow one specific "alert" prefix, like Ctrl-U, which would behave a bit like a modifier key, in that you could add it to any other combination of modifiers, e.g. Ctrl-U, Shift+Ctrl+X, so doubling the number of keystrokes available. Is it worth it, or wait until someone complains they've run out of shortcuts?
4. Add shadow commands for status bar buttons, e.g. next/prev page
5. Add shadow command for "Quit without prompting for Save"
6. Allow Option key as a modifier on macOS. There are no more modifiers available for Windows. The Windows key doesn't act as a modifier, and the Alt Gr key is equivalent to Ctrl+Alt.
7. Set shortcut without the checkboxes, i.e. detect what modifier keys the user has pressed in the Edit shortcut dialog.
8. On Macs, `Cmd+.` does `Cancel` in the current dialog. Can that type of shadow command be implemented? The difference is that it would run on the dialog with focus, rather than a specific dialog, e.g. it could be any of the OkCancel dialogs, of which there may be more than one.

windymilla:
1. Not doing it
2. Done, plus Replace All
3. Not doing it
4. No need, since they all already exist as commands
5. Done
6. Trialled and decided against it
7. Done
8. Not doing it (for now)

PR #962, opened 2025-04-26
Improve setting of shortcut keystrokes

windymilla:
Allow user to press key with modifiers instead of using checkboxes to set the modifiers.

Should also allow use of option key on macOS.

Should address points 6 & 7 of #956

### DIFF
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1023,10 +1023,17 @@ class CommandEditDialog(OkApplyCancelDialog):
             shortcut_entry,
             "Press required modifiers and keyboard key to use for shortcut",
         )
+
+        # Just used for test binding
+        self.dummy_widget = ttk.Label(self.top_frame, takefocus=False)
+
         # Track which modifier keys are currently pressed
         self.pressed_modifiers: set[str] = set()
         self.bind("<KeyPress>", self.key_press)
         self.bind("<KeyRelease>", self.key_release)
+        # Clear modifiers if dialog loses focus, particularly via Alt-tab on Windows
+        self.bind("<FocusOut>", lambda _: self.pressed_modifiers.clear())
+
         self.cmd = EntryMetadata("", "", "")
 
     @property
@@ -1059,6 +1066,20 @@ class CommandEditDialog(OkApplyCancelDialog):
             logger.error(
                 f"Shortcut must include Ctrl or {'Cmd' if is_mac() else 'Alt'}"
             )
+            self.lift()
+            self.focus()
+            return False
+
+        # Bind do_nothing to dummy widget, to test if the bind sequence is legal
+        def do_nothing(_: tk.Event) -> None:
+            """Do nothing."""
+
+        try:
+            test_key = process_accel(new_shortcut)[1]
+            self.dummy_widget.bind(test_key, do_nothing)
+            self.dummy_widget.unbind(test_key)
+        except tk.TclError:
+            logger.error(f"Key combination {test_key} is not supported as a shortcut.")
             self.lift()
             self.focus()
             return False
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1064,8 +1064,10 @@ class CommandEditDialog(OkCancelDialog):
             True if successful.
         """
         new_shortcut = self.shortcut
-        # Don't allow shortcuts that don't use Ctrl/Cmd/Alt
-        if not any(m in new_shortcut for m in ["Ctrl", "Cmd", "Alt"]):
+        # Don't allow shortcuts that don't use Ctrl/Cmd/Alt except for F keys
+        if not any(m in new_shortcut for m in ["Ctrl", "Cmd", "Alt"]) and not re.search(
+            r"F\d+$", new_shortcut
+        ):
             logger.error(
                 f"Shortcut must include Ctrl or {'Cmd' if is_mac() else 'Alt'}"
             )
@@ -1076,6 +1078,15 @@ class CommandEditDialog(OkCancelDialog):
         # Don't allow shortcuts that use Option
         if "Option" in new_shortcut:
             logger.error("Option key may not be used for shortcuts")
+            self.lift()
+            self.focus()
+            return False
+
+        # Plain F1 is reserved
+        if new_shortcut == "F1":
+            logger.error(
+                f"F1 without Shift, Ctrl or {'Cmd' if is_mac() else 'Alt'} is reserved for Help"
+            )
             self.lift()
             self.focus()
             return False
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1054,24 +1054,10 @@ class CommandEditDialog(OkApplyCancelDialog):
             True if successful.
         """
         new_shortcut = self.shortcut
-        single_char_keyname = (
-            len(re.sub(r".*\+", "", self.shortcut_variable.get())) == 1
-        )
-        if single_char_keyname and not (
-            "Ctrl" in new_shortcut
-            or "Cmd" in new_shortcut
-            or "Alt" in new_shortcut
-            or "Option" in new_shortcut
-        ):
+        # Don't allow shortcuts that don't use Ctrl/Cmd/Alt
+        if not any(m in new_shortcut for m in ["Ctrl", "Cmd", "Alt"]):
             logger.error(
                 f"Shortcut must include Ctrl or {'Cmd' if is_mac() else 'Alt'}"
-            )
-            self.lift()
-            self.focus()
-            return False
-        if new_shortcut == "F1":
-            logger.error(
-                f"F1 without Shift, Ctrl or {'Cmd' if is_mac() else 'Alt'} is reserved for Help"
             )
             self.lift()
             self.focus()
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1091,6 +1091,13 @@ class CommandEditDialog(OkCancelDialog):
             self.focus()
             return False
 
+        # Cmd+? is reserved for Help too
+        if new_shortcut == "Cmd+Shift+?":
+            logger.error("Cmd+Shift+? is reserved for Help")
+            self.lift()
+            self.focus()
+            return False
+
         # Other reserved shortcuts
         ctrl_cmd = "Cmd" if is_mac() else "Ctrl"
         display_shortcut = process_accel(new_shortcut)[0]
@@ -1165,6 +1172,10 @@ class CommandEditDialog(OkCancelDialog):
         keysym = event.keysym
         if keysym in self.MODIFIER_KEYS:
             self.pressed_modifiers.add(keysym)
+        elif (
+            event.keysym in ("BackSpace", "Delete") and len(self.pressed_modifiers) == 0
+        ):
+            self.shortcut = ""
         else:
             # Combine the current modifiers with the key
             mods = sorted(set(self.MODIFIER_KEYS[kk] for kk in self.pressed_modifiers))
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1064,10 +1064,11 @@ class CommandEditDialog(OkCancelDialog):
             True if successful.
         """
         new_shortcut = self.shortcut
+        display_shortcut = process_accel(new_shortcut)[0]
         # Don't allow shortcuts that don't use Ctrl/Cmd/Alt except for F keys
-        if not any(m in new_shortcut for m in ["Ctrl", "Cmd", "Alt"]) and not re.search(
-            r"F\d+$", new_shortcut
-        ):
+        if not any(
+            m in display_shortcut for m in ["Ctrl", "Cmd", "Alt"]
+        ) and not re.search(r"F\d+$", display_shortcut):
             logger.error(
                 f"Shortcut must include Ctrl or {'Cmd' if is_mac() else 'Alt'}"
             )
@@ -1076,14 +1077,14 @@ class CommandEditDialog(OkCancelDialog):
             return False
 
         # Don't allow shortcuts that use Option
-        if "Option" in new_shortcut:
+        if "Option" in display_shortcut:
             logger.error("Option key may not be used for shortcuts")
             self.lift()
             self.focus()
             return False
 
         # Plain F1 is reserved
-        if new_shortcut == "F1":
+        if display_shortcut == "F1":
             logger.error(
                 f"F1 without Shift, Ctrl or {'Cmd' if is_mac() else 'Alt'} is reserved for Help"
             )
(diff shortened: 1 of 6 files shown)


windymilla:
Testing notes: anything may have broken, so testing should include
1. All printable ASCII characters should be able to be used as the base key, though not every one has to be tested. Also a few others, such as <code>¬`£§±</code> which are displayed on a Windows UK keyboard or an Apple UK keyboard. "Unknown" keys, i.e. ones I haven't catered or explicitly in the code could/should still work but will appear in the display with the Tk name, rather than the character, e.g. if you can type `Ctrl+¶` on your keyboard, it will probably appear as `Ctrl+paragraph`. This should mean that users with LOTE keyboards can hopefully still use the feature. 
2.  Re-using a shortcut you assigned previously
3. Re-using a shortcut that was assigned to an existing feature by default, e.g. Cmd+O for Open file.
4. Attempting to set up a shortcut without using Ctrl, Cmd, Option or Alt for a key that has a single character name, e.g. "x" as opposed to "Home"
5. F function keys should still work.
6. It's possible you may be able to fool the detection of the modifier keys if you really try, like pressing left Ctrl and right Ctrl, then releasing left Ctrl, then switching focus to another app, then releasing right Ctrl, or other stuff like that, but I hope for "normal" use it will be reliable.

src/guiguts/misc_dialogs.py
```diff
@@ -1146,45 +1118,27 @@ def apply_changes(self) -> bool:
...
-        if self.control_variable.get():
-            shortcut += "Ctrl+"
-        if self.shift_variable.get():
-            shortcut += "Shift+"
-        if self.alt_command_variable.get():
-            shortcut += "Cmd+" if is_mac() else "Alt+"
-        shortcut += self.key
-        self.shortcut = shortcut
+        keysym = event.keysym
+        if keysym in self.MODIFIER_KEYS:
```
tangledhelix, line 1124: Cmd, Ctrl, Shift seem to work as expected on Mac.

Option isn't working. Here you are collecting a chord of keys, and then storing the modifier key(s) before storing the base key. But on Mac, `Opt-c` (where `c` is a non-modifier key) is most often used to input special characters (for example º ™ ¢ € and so on). When I try to use Opt in a chord, I believe it is sending that special character rather than the base character. (Yet, at the same time, it stores Opt as a modifier.)

So for example when I try to bind `Cmd-Opt-k` to a command, it displays this in the dialog:

`Cmd+Option+??`

When I try to save this, an exception is thrown. So I can't save it. I believe it's actually trying to store `Cmd-Opt-˚` where ˚ is the character produced by pressing Opt-k (at least in the US layout I'm using).

Trying with only Opt and no other modifier is even stranger looking. Opt-d as a bind is showing in the dialog as: `Option+partialderiv@ive` (Opt-d produces ∂)

Others are similarly odd, Opt-p `Option+Greek_pi`, Opt-t `Option+dagger` and so on. Some of these actually work when you use them (such as Opt-t) and others don't because you can't even save them, as noted above.

tangledhelix: And wow, just realizing that it took ∂, which is partial derivative, and somehow it replaced the "at" buried in that word with an @ ?? Bizarre! Is that your code doing that somewhere or is tkinter or Python doing it on its own?

srjfoo: I'm trying to do a cross-reference for all the keys showing what I see as the shortcut for Option and Option+shift (since one of the questions is whether Option is going to be useful. If it's useful at all, it should be used only in combination with ctrl or cmd, because of the dead key combinations., and what the keyboard actually types, but it's taking longer than I expected, and (and there were also distractions).

I was puzzled by the ??, too -- and there are several of them.

And yes, I get an exception if I try to assign one of those. But I can assign some of the option keys and get them to work -- I just have to put the edit shortcut dialog in the background, apparently.

windymilla: AFK, but a couple of comments. Tkinter seems to receive the character that would appear when typing, eg percent for shift 5, rather than 5, or other rarer keys if you use Opt. That's the case generally, not specifically to do with Opt.

My code isn't converting at to @

@srjfoo as I mentioned before, if you want the user to be able to type the full shortcut with modifiers in the edit dialog, you don't want it to do the action. That's why you can't do the action when the edit dialog is in the foreground. If it wasn't like that, then when you type the shortcut, trying to set it, it would do whatever the current action was.

windymilla: Overall, it seems to me as though Option is causing more problems than solutions, but it's your call if you want to persevere with it. Personally I think there are enough combinations with 3 modifier keys, like on Windows

tangledhelix:
I immediately get this traceback on Linux:

```
  File "<string>", line 1, in <module>
  File "/home/ubuntu/guiguts-py/src/guiguts/application.py", line 1315, in main
    Guiguts().run()
    ^^^^^^^^^
  File "/home/ubuntu/guiguts-py/src/guiguts/application.py", line 146, in __init__
    self.init_menus()
  File "/home/ubuntu/guiguts-py/src/guiguts/application.py", line 577, in init_menus
    self.recreate_menus()
  File "/home/ubuntu/guiguts-py/src/guiguts/application.py", line 1025, in recreate_menus
    self.add_entries(top_level)
  File "/home/ubuntu/guiguts-py/src/guiguts/application.py", line 1048, in add_entries
    self.add_entries(entry)  # Add its entries
    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/guiguts-py/src/guiguts/application.py", line 1050, in add_entries
    tk_menu.add_button(  # type:ignore[attr-defined]
  File "/home/ubuntu/guiguts-py/src/guiguts/mainwindow.py", line 96, in add_button
    Menu.bind_button(callback, accel, bind_all)
  File "/home/ubuntu/guiguts-py/src/guiguts/mainwindow.py", line 120, in bind_button
    maintext().key_bind(key_event, lambda _: callback(), bind_all=bind_all)
  File "/home/ubuntu/guiguts-py/src/guiguts/maintext.py", line 1405, in key_bind
    self.bind_event(lk, handler, bind_all=bind_all, bind_peer=True)
  File "/home/ubuntu/guiguts-py/src/guiguts/maintext.py", line 1184, in bind_event
    self.bind(event_string, break_func, add)
  File "/usr/lib/python3.12/tkinter/__init__.py", line 1450, in bind
    return self._bind(('bind', self._w), sequence, func, add)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/tkinter/__init__.py", line 1404, in _bind
    self.tk.call(what + (sequence, cmd))
_tkinter.TclError: bad event type or keysym "!"
```

It doesn't like the set of keybinds defined in the `shortcuts` list in `init_bookmark_menu`. I had to comment out all five of them before it would even start.

src/guiguts/misc_dialogs.py
windymilla: Or it could be documented that only some option combinations work, of course with the code trapping exceptions

windymilla: Or maybe there's a way to get the raw keystroke, not the char or charsym that GG receives

windymilla:
That's an oversight. Since the latest version, that should be using ctrl-shift-exclamation and the code converts the exclamation to ! For display purposes.

windymilla:
> It doesn't like the set of keybinds defined in the `shortcuts` list in `init_bookmark_menu`. I had to comment out all five of them before it would even start.

This should be fixed now.

src/guiguts/misc_dialogs.py
windymilla: > ... is whether Option is going to be useful. If it's useful at all, it should be used only in combination with ctrl or cmd, because of the dead key combinations., and what the keyboard actually types

The latest commit enforces that. Every shortcut must contain at least one of Ctrl, Cmd, Alt. 

> But I can assign some of the option keys and get them to work -- I just have to put the edit shortcut dialog in the background, apparently.

Hopefully my earlier comment was clear about the reason for this. If Cmd+K opens the Spell Checker, but you want to reallocate it to run Jeebies, then when you press Cmd+K in the Command Edit dialog, you don't want it to open the Spell Checker, even though it is currently bound to it. Essentially, the Command Edit dialog has to catch all key strokes and not let them execute the command they are normally bound to. That behavior is incompatible with using the shortcut to execute a command while focus is in the Command Edit dialog. So, you have to put it in the background. Unless @tangledhelix has any ideas on how to avoid it.

windymilla: > > But I can assign some of the option keys and get them to work -- I just have to put the edit shortcut dialog in the background, apparently.
> 
> Hopefully my earlier comment was clear about the reason for this. If Cmd+K opens the Spell Checker, but you want to reallocate it to run Jeebies, then when you press Cmd+K in the Command Edit dialog, you don't want it to open the Spell Checker, even though it is currently bound to it. Essentially, the Command Edit dialog has to catch all key strokes and not let them execute the command they are normally bound to. That behavior is incompatible with using the shortcut to execute a command while focus is in the Command Edit dialog. So, you have to put it in the background. Unless @tangledhelix has any ideas on how to avoid it.

@srjfoo - I had an idea about this. I could remove the "Apply" button from the Command Edit dialog (which dates from the previous shortcut setting method). That way, you would have to press "OK" to set the shortcut, which dismisses the dialog, and you can then use it without the shortcut keystrokes being gobbled by the dialog. Do you think that's a good idea?

windymilla:
Latest commit clears modifier flags if dialog loses focus, so that using Alt-tab doesn't leave behind an "Alt is pressed" flag in the dialog.

Also, should now get an error message when trying to bind a new sequence if it is not supported. Code does a test bind to a dummy widget and traps the exception.

src/guiguts/misc_dialogs.py
srjfoo: > > > But I can assign some of the option keys and get them to work -- I just have to put the edit shortcut dialog in the background, apparently.
> > 
> > 
> > Hopefully my earlier comment was clear about the reason for this. If Cmd+K opens the Spell Checker, but you want to reallocate it to run Jeebies, then when you press Cmd+K in the Command Edit dialog, you don't want it to open the Spell Checker, even though it is currently bound to it. Essentially, the Command Edit dialog has to catch all key strokes and not let them execute the command they are normally bound to. That behavior is incompatible with using the shortcut to execute a command while focus is in the Command Edit dialog. So, you have to put it in the background. Unless @tangledhelix has any ideas on how to avoid it.
> 
> @srjfoo - I had an idea about this. I could remove the "Apply" button from the Command Edit dialog (which dates from the previous shortcut setting method). That way, you would have to press "OK" to set the shortcut, which dismisses the dialog, and you can then use it without the shortcut keystrokes being gobbled by the dialog. Do you think that's a good idea?

In actual production use, maybe so. For testing purposes, probably not. That would force me to re-open the "Edit Shortcut" dialog for every test assignment I make, which would be annoying. What about a "Test" button in the dialog? (Either as a temporary expedient, or permanent, if you think it's a useful idea.) But it's not really necessary, I can get used to putting the dialog in the background -- I'm not sure why I thought it worked before, because your explanation makes perfect sense, and I have no idea why that didn't occur to me.

windymilla:
Re-requesting from @tangledhelix due to Linux-specific report of dialog not popping to the front.

src/guiguts/widgets.py
```diff
@@ -129,8 +129,12 @@ def show_dialog(
         if dlg := cls.get_dialog():
             dlg.deiconify()
             dlg.reset()
+            # On Linux, Window Managers stop dialogs being brought to the front by deiconify,
+            # or anything else. Destroying the dialog and recreating it gets round that.
+            if is_x11():
+                dlg.destroy()
```
tangledhelix, line 135: I don't think this is a necessary step. This patch is working for me on Linux to bring the palette back to focus (and in the case of that specific dialog, it focuses the text entry too). For other things like jeebies, spellcheck, etc, it also seems to raise them as expected.

```diff
diff --git a/src/guiguts/widgets.py b/src/guiguts/widgets.py
index 11eb566..3c3ba6c 100644
--- a/src/guiguts/widgets.py
+++ b/src/guiguts/widgets.py
@@ -128,6 +128,10 @@ class ToplevelDialog(tk.Toplevel):
         # Can we just deiconify dialog & reset it?
         if dlg := cls.get_dialog():
             dlg.deiconify()
+            if is_x11():
+                grab_focus(dlg)
+                if dlg_name == "CommandPaletteDialog":
+                    dlg.entry.focus()
             dlg.reset()
         # Or do we need to create it?
         else:
```

windymilla: On WSL, that does put the focus in the Command Palette (in the text entry), but it doesn't raise it to the top of the window manager stack. Does it raise it on your Linux, @tangledhelix?

For me if the Command Palette is hidden behind the main window, you still have the same issue that you originally described - it seems like the shortcut or menu button haven't worked. It looks like it's the kind of thing that will be dependent on the Linux window manager being used.

If this is useful on some window managers, we should keep it. However, we'd need to structure it slightly differently. The TopLevelDialog class is not permitted to know/use the names of specific dialogs. 

Probably we should add a `grab_focus` method, which in TopLevelDialog would do:
```python
if is_x11():
    grab_focus(self)
```
Then in CommandPaletteDialog we'd override `grab_focus` to first call the base class method, then (on all platforms):
```python
    self.entry.focus()
```
Then we would add `dlg.grab_focus()` to TopLevelDialog's `show_dialog` method.

Some other dialogs might benefit from similarly overriding the new `grab_focus` method to force focus to a specific widget within the dialog.

I'm happy to do that work, if you confirm that your code is solving the "raise to the top" issue on your Linux window manager.

windymilla:
@tangledhelix - Please can you confirm if the latest commit (which I think is broadly equivalent to your code) solves the "raising" issue on your Linux. Thanks.

windymilla:
Fixes #970

tangledhelix:
> @tangledhelix - Please can you confirm if the latest commit (which I think is broadly equivalent to your code) solves the "raising" issue on your Linux. Thanks.

Yes, it does. I didn't try every possible tool, but it works for the Palette, Bookloupe, Jeebies, and Spelling, at least.

I don't know if this is expected, but raising the latter 3 mentioned above caused a re-run to happen. That may be normal, but I wasn't expecting it to happen since I didn't tell it to re-run. I would think the user wouldn't want that, because if they had cleared some items out of the list already, then clicked into the window (thus de-focusing the tool) and going back, they may well not want everything to come back that they had already cleared.

windymilla:
> > @tangledhelix - Please can you confirm if the latest commit (which I think is broadly equivalent to your code) solves the "raising" issue on your Linux. Thanks.
> 
> Yes, it does. I didn't try every possible tool, but it works for the Palette, Bookloupe, Jeebies, and Spelling, at least.
> 
> I don't know if this is expected, but raising the latter 3 mentioned above caused a re-run to happen. That may be normal, but I wasn't expecting it to happen since I didn't tell it to re-run. I would think the user wouldn't want that, because if they had cleared some items out of the list already, then clicked into the window (thus de-focusing the tool) and going back, they may well not want everything to come back that they had already cleared.

Just refocusing by clicking the dialog or Cmd tab doesn't rerun, does it? But if you choose the menu option or use a shortcut it should rerun. The menu option means run the tool, not focus/raise the dialog. Focusing and raising is a side effect of running the tool.

srjfoo (approved):

tangledhelix:
> Just refocusing by clicking the dialog or Cmd tab doesn't rerun, does it?

That's correct.

PR #990, opened 2025-05-05
Mark file state dirty when setting bookmark

### LINKED ISSUE #989: Setting a bookmark doesn't mark the save state as dirty
Bookmarks are saved to the JSON file for persistence, but setting a bookmark doesn't mark the file state as dirty. Therefore it's possible to set one or more bookmarks and close/quit without being told to save. If the user didn't happen to save in between, the bookmarks are lost.

tangledhelix:
Setting a bookmark didn't mark the file as edited; therefore you could close or quit the file without saving (or being asked if you want to save) and lose the bookmark permanently.

Testing notes:
- On `master`, set a bookmark and then close or quit without saving. The bookmark should be lost on next startup.
- On this branch, set a bookmark and then close or quit without saving; you should be asked if you want to save.

Fixes #989

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -1072,6 +1072,7 @@ class File:
             f"{BOOKMARK_BASE}{bm_num}", maintext().get_insert_index()
         )
         self.highlight_bookmark(bm_num)
+        maintext().set_modified(True)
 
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

windymilla (approved):
Thanks

PR #991, opened 2025-05-05
Image rotation

### LINKED ISSUE #953: Consider image rotation control
See previous discussion in #483 

* Do we need this at all?
* Do we need more than one button that rotates in one direction by 90º per click?
* If the user rotates an image should we remember that the image has been rotated and store that metadata in the project json file so it doesn't reset every time the user navigates to a different image?

tangledhelix:
Adds a button to the image viewer control frame to rotate the image by 90º counter-clockwise. If an image has been rotated, the rotation will be stored in the JSON file to persist across restarts. Images that haven't been rotated aren't stored in this data and assumed to be rotated 0º.

Testing notes:
- Load a project with images
- Rotate some images however you like
- Navigate away and back again; the rotation should be remembered
- Save the file and restart; the rotations should be remembered
- Rotate more images, without saving, and try to close the file. It should ask you to save.

There's a new key in the JSON file called `imagerotate` that stores the data. By default, images aren't in this data. Once an image has been rotated, it gets added and tracked.

Fixes #953

### DIFF
--- a/src/guiguts/file.py
+++ b/src/guiguts/file.py
@@ -54,6 +54,7 @@ BINFILE_KEY_IMAGEDIR: Final = "imagedir"
 BINFILE_KEY_PROJECTID: Final = "projectid"
 BINFILE_KEY_LANGUAGES: Final = "languages"
 BINFILE_KEY_BOOKMARKS: Final = "bookmarks"
+BINFILE_KEY_IMAGEROTATE: Final = "imagerotate"
 
 PAGE_FLAGS_NONE = 0
 PAGE_FLAGS_SOME = 1
@@ -87,6 +88,7 @@ class BinDict(TypedDict):
     projectid: str
     languages: str
     bookmarks: dict[str, str]
+    imagerotate: dict[str, int]
 
 
 class File:
@@ -247,6 +249,7 @@ class File:
             return
         maintext().set_insert_index(IndexRowCol(1, 0))
         self.languages = preferences.get(PrefKey.DEFAULT_LANGUAGES)
+        mainimage().reset_rotation_details()
         bin_matches_file = self.load_bin(filename)
         self.mark_page_boundaries()
         flags_found = (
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
Very nice!

src/guiguts/file.py
```diff
@@ -508,6 +510,9 @@ def interpret_bin(self, bin_dict: BinDict) -> bool:
         if bookmarks := bin_dict.get(BINFILE_KEY_BOOKMARKS):
             for key, value in bookmarks.items():
                 self.set_bookmark_index(key, value)
+        image_rotations: Optional[dict[str, int]]
+        if image_rotations := bin_dict.get(BINFILE_KEY_IMAGEROTATE):
+            mainimage().rotation_details = image_rotations
```
windymilla, line 515: I think this needs something like the following to clear the previous file's rotations when the newly loaded dict doesn't have any image rotations:
```python
else:
    mainimage().rotation_details = {}
```

windymilla (changes requested):
As described in the comment, I think rotation information is not being cleared at one point when it should be (file.py:515)
1. Load a file, rotate the page 001, save the file (but don't quit GG)
2. Load another file (that's never had page rotations) and view page 001. 
3. For me, it appears to pick up the rotation from the previous file.

srjfoo:
> As described in the comment, I think rotation information is not being cleared at one point when it should be (file.py:515)
> 
> 1. Load a file, rotate the page 001, save the file (but don't quit GG)
> 2. Load another file (that's never had page rotations) and view page 001.
> 3. For me, it appears to pick up the rotation from the previous file.

I've duplicated these results. Further comment:
* Opening another project, whether or not its images have been rotated, that has the first image at normal rotation, and then reopening the sample(s) with no rotated images will display the image properly.
* It doesn't matter whether or not the project is still open. If I save and close the project, then open a project that has never had an image rotated, the first image will be rotated. And any subsequent project that has never had an image rotated, until it's reset.
* I was able to get similar results rotating the second image, as well. It didn't seem to happen every time (that is, not projects that I'd used in the first round of testing), but it did on additional projects that I'd used specifically for testing for the second image as well as the first.

(probably more than you need to know 😁)

src/guiguts/file.py
tangledhelix: Good catch, that's exactly it. New version pushed.

tangledhelix:
Last change didn't take. After a couple more attempts, it should be correct now. Now will reset state on load of a new file (including on application start with `-r1`). Reset happens prior to checking for, or reading, a .json file, so it should work whether the new file has a .json or doesn't.

srjfoo (approved):

windymilla (approved):
LGTM!

PR #992, opened 2025-05-05
Add regex timeout

windymilla:
Some regexes can take effectively for ever, so add a timeout, settable via an advanced Pref.

If regex times out during search, disable highlighting of search string temporarily, or it will keep trying to repeat the failing search.

Error message may appear twice, once for the search and once for the highlighting. I think this is an
acceptable price to pay - alternatives such as setting a flag telling searches to be ignored would run a
risk of breaking things. This is to fix a very rare occurrence in a way that doesn't risk breaking
other user searches or intra-code searching.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -544,6 +544,7 @@ class Guiguts:
         preferences.set_default(PrefKey.SURROUND_WITH_BEFORE_HISTORY, [])
         preferences.set_default(PrefKey.SURROUND_WITH_AFTER, "")
         preferences.set_default(PrefKey.SURROUND_WITH_AFTER_HISTORY, [])
+        preferences.set_default(PrefKey.REGEX_TIMEOUT, 5)
 
         # Check all preferences have a default
         for pref_key in PrefKey:
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2643,7 +2643,21 @@ class MainText(tk.Text):
         if nocase:
             flags |= re.IGNORECASE
 
-        match = re.search(search_string, slurp_text, flags=flags)
+        try:
+            match = re.search(
+                search_string,
+                slurp_text,
+                flags=flags,
+                timeout=preferences.get(PrefKey.REGEX_TIMEOUT),
+            )
+        except TimeoutError:
+            logger.error(
+                "Regex timed out. Try changing the regex or flags;\n"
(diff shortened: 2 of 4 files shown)


srjfoo:
Testing on macOS, I had a slightly different experience. I started with a 20 second timeout, and of course it timed out. The odd thing was that it actually highlighted the first match (highlighted, not spotlighted), which goes from line 4138 to line 4335. I brought the search form back into the foreground, and it continued thes search (I did this twice), until it timed out again (both times). After the third time, it did not continue the search. 

Image

I then extended the timeout to 300 seconds. It still timed out (which, based on the original problem, I'm not surprised about. But with the same results. That is, the first match was actually highlighted/selected: when I shifted focus to the main text and pressed delete, the text was deleted.

Also, before going to try the revised regex, just for fun, I tried "Find all". It worked almost instantaneously. 

image

Notice the position of the horizontal scrollbar. I positioned it so that you could see the length of the selection, second-longest relative to longest. One odd thing that I noticed -- when clicking on the matches in "find all", the first one also spotlighted the blank line before the opening bq tag. The others did not.

One other thing that occurred to me was that, if you need a selection, find all doesn't give you that -- or give you a way to create a selection from the spotlight. Do we need something like that?

---

Confirmed that the revised regex works instantaneously, still (i.e. in the branch).

srjfoo (approved):

windymilla:
> One other thing that occurred to me was that, if you need a selection, find all doesn't give you that -- or give you a way to create a selection from the spotlight. Do we need something like that?

If you are wanting to do a replacement, then you need to go back to the S/R dialog. I'm not sure what else you would use a selection for from the spotlight. Find All is more for looking through matches, where S/R Replace & Search is for going through one at a time making changes.

srjfoo:
> > One other thing that occurred to me was that, if you need a selection, find all doesn't give you that -- or give you a way to create a selection from the spotlight. Do we need something like that?
> 
> If you are wanting to do a replacement, then you need to go back to the S/R dialog. I'm not sure what else you would use a selection for from the spotlight. Find All is more for looking through matches, where S/R Replace & Search is for going through one at a time making changes.

Trying to explain my thought processes: "Find All", unlike WF results (where you can cmd/ctrl-click on an entry in the list to have that specific entry show in the search form) lacks a way to get from the results list directly to that specific instance. And perhaps that's expecting too much. Thanks -- this has helped me see that, while I definitely like the Find All function, it has a different purpose from the basic S/R dialog.

PR #1000, opened 2025-05-07
Add shadow commands for image scrolling

### LINKED ISSUE #483: Add shortcuts to scroll image viewer up/down

windymilla:
Scroll Image Up, down, left & right. User can define their own shortcuts.

Fixes #483 (the only bit left to do)

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -842,6 +842,22 @@ class MainImage(tk.Frame):
         menubar_metadata().add_checkbutton_orphan(
             "Image Fit ↑↓", PrefKey.IMAGE_AUTOFIT_HEIGHT
         )
+        menubar_metadata().add_button_orphan(
+            "Scroll Image ↑",
+            lambda: self.canvas.yview_scroll(1, "units"),
+        )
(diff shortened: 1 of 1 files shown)


srjfoo:
For me, this completely breaks trackpad (two-finger) scrolling, and hamstrings mouse wheel scrolling. For me, it would be a very poor trade-off. 🙁 

Comment -- at first, I set the shortcuts to `cmd`+`ctrl`+`shift`, plus the arrow keys, so as not to break the single mod key versions. This did not work well for me -- something got confused, and occasionally the active pane would switch to the text pane, and start selecting text. When I picked an inverted T of regular keys,it seemed to work just fine, so I thought I'd mention that the arrow keys may be unpredictable.

srjfoo:
> For me, this completely breaks trackpad (two-finger) scrolling, and hamstrings mouse wheel scrolling. For me, it would be a very poor trade-off. 🙁

Correction -- it appears to be Tk9 that broke it. Trackpad scrolling is clunky in Tk8.whatever, but it works.

PR #1001, opened 2025-05-07
Improvements to Stealth Scannos

windymilla:
1. Five scannos with hints "Find a string that starts with xx and change it to yy" are better with a lookahead to check there is a character after it. This stops them matching "tb" (thought break) or "hr" (horizontal rule).
2. Scannos dialog tooltip now includes "Shift" actions
3. Various "Stealth Scannos" shadow commands added, including Replace, Replace All, Prev Scanno, Next Scanno.
4. "guess" exception extended to  include "(mar)quess"
5. Punct at start of line no longer triggered by ellipsis
6. "ii" now only suspected "ll" if preceded by a, e, o, u.
7. Ellipsis followed by lower case no longer reported

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -82,6 +82,7 @@ from guiguts.misc_tools import (
     convert_to_curly_quotes,
     check_curly_quotes,
     indent_selection,
+    ScannoCheckerDialog,
 )
 from guiguts.page_details import PageDetailsDialog
 from guiguts.preferences import preferences, PrefKey, PersistentBoolean
@@ -1104,6 +1105,7 @@ class Guiguts:
         PreferencesDialog.add_orphan_commands()
         CommandPaletteDialog.add_orphan_commands()
         SearchDialog.add_orphan_commands()
+        ScannoCheckerDialog.add_orphan_commands()
         mainimage().add_orphan_commands()
         menubar_metadata().add_button_orphan(
             "Quit Without Saving File", lambda: root().quit()
--- a/src/guiguts/data/scannos/regex.json
+++ b/src/guiguts/data/scannos/regex.json
@@ -106,27 +106,27 @@
 "hint": "Find repeated words or word sequences and remove one"
 },
 {
-"match": "\\bhl",
+"match": "\\bhl(?=\\w)",
 "replacement": "bl",
 "hint": "Find a string that starts with hl and change it to bl"
 },
 {
-"match": "\\bhr",
+"match": "\\bhr(?=\\w)",
 "replacement": "br",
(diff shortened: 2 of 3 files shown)


srjfoo:
(nts: this PR must be checked with auto-advance off 😁)

Scanno search 33/45 identifies page separators. Mind you that's not an issue if page separators have already been removed, but when testing, I often just have the CTFs that I'm working with, not a project at a recognizable interim stage appropriate for doing scanno checks.

Scanno search 41/45 still identifies thought breaks. I don't see how you can stop it from doing so unless you specify that there has to be a word character either before or after the string. For example, "math" does't have any word characters following the th, and "the" doesn't have a word character before, so `(\w?)tb(\w?)` still identifies thought breaks, `tb(\w)|(\w)tb` works, though.

windymilla:
> Scanno search 33/45 identifies page separators. Mind you that's not an issue if page separators have already been removed, but when testing, I often just have the CTFs that I'm working with, not a project at a recognizable interim stage appropriate for doing scanno checks.

We'll have to live with that one during testing, I think

> Scanno search 41/45 still identifies thought breaks. I don't see how you can stop it from doing so unless you specify that there has to be a word character either before or after the string. For example, "math" does't have any word characters following the th, and "the" doesn't have a word character before, so `(\w?)tb(\w?)` still identifies thought breaks, `tb(\w)|(\w)tb` works, though.

I've improved the regex in the latest commit. It has to use lookahead/behind as you don't want the word character to be included in the search, so I went for the lookahead/behind equivalent of your regex: `(?<=\\w)tb|tb(?=\\w)`

srjfoo (approved):
One minor nit, that I don't know if you'd considered:

> "guess" exception extended to include "(mar)quess"

You do realize this means that `marguess` will not be identified by this check (assuming that OCR/proofers might not correctly identify one or the other?

Mind you, spellcheck should find it (but then spellcheck should find a lot of what are considered to be scannos for the purposes of this check).

windymilla:
> You do realize this means that `marguess` will not be identified by this check (assuming that OCR/proofers might not correctly identify one or the other?

This check is to catch inverted `n` in words that contain `ness`, like `kindness`. It would just be a simple check for `uess` but `guess` is a common false positive. Allowing `quess` is just avoiding the `marquess(es)` false positive, which I have also seen. If you used the previous version of this check to try to fix `marguess`, you'd get `margness`.

srjfoo:
> If you used the previous version of this check to try to fix `marguess`, you'd get `margness`.

😁 -- narrow focus -- it'll trip you up every time (yes, I did lose sight of that). Thanks!

PR #1003, opened 2025-05-07
Add Commonly Used Chars to Unicode menu

### LINKED ISSUE #975: Commonly Used Characters is well hidden
I had quite a hunt for the "Commonly Used Characters" list! It was tucked away in the "Unicode Block" section of the GG2 Manual, which wasn't where I'd ever think to look. While I don't necessarily object to its location, I do feel that for a frequently used feature, it would be much more convenient if it were easily accessible on the bottom toolbar like GG1.  Adding to my previous thought about the "Commonly Used Characters" list, I'd also find it incredibly useful to have a dedicated keystroke combination to access it.

windymilla:
It's really the Unicode Block dialog, but with the dropdown block selection set to Commonly Used
Chars.

If you change blocks, it will be stored in the pref, and when you next do Unicode Block, it will show
that block, but if you do Common Chars, it will
force to that. If you change the block to be
Common Chars, that will be stored in the pref,
and that's what you'll get next time you do pop
Unicode Block.

Also added Shift left/right click to status bar, where the character name is shown - instructions in tooltip.

Fixes #975

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -882,7 +882,13 @@ class Guiguts:
         unicode_menu.add_button(
             "Unicode ~Search/Entry", UnicodeSearchDialog.show_dialog
         )
-        unicode_menu.add_button("Unicode ~Blocks", UnicodeBlockDialog.show_dialog)
+        unicode_menu.add_button(
+            "Unicode ~Blocks", UnicodeBlockDialog.show_unicode_dialog
+        )
+        unicode_menu.add_button(
+            "C~ommonly Used Characters",
+            UnicodeBlockDialog.show_common_dialog,
+        )
         unicode_menu.add_button(
             "~Normalize Selected Characters",
             unicode_normalize,
@@ -1240,10 +1246,18 @@ class Guiguts:
 
         the_statusbar.add(
             "ordinal",
-            tooltip="Character after the cursor\nClick to toggle name display",
+            tooltip="Character after the cursor\nClick to toggle name display\n"
+            "Shift left click to show Commonly Used Characters dialog\n"
+            "Shift right click to show Unicode Blocks dialog",
             update=ordinal_str,
         )
         the_statusbar.add_binding("ordinal", "ButtonRelease-1", ordinal_name_display)
+        the_statusbar.add_binding(
+            "ordinal", "Shift-ButtonRelease-1", UnicodeBlockDialog.show_common_dialog
+        )
+        the_statusbar.add_binding(
+            "ordinal", "Shift-ButtonRelease-3", UnicodeBlockDialog.show_unicode_dialog
+        )
 
     def logging_init(self) -> None:
         """Set up basic logger until GUI is ready."""
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -1292,7 +1292,7 @@ class StatusBar(ttk.Frame):
                 self.set(key, func())
         self.after(200, self._update)
 
-    def add_binding(self, key: str, event: str, callback: Callable[[], None]) -> None:
+    def add_binding(self, key: str, event: str, callback: Callable[[], Any]) -> None:
         """Add an action to be executed when the given event occurs
 
         Args:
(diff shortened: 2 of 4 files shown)


jjzsdp:
Is it felt irrelevant to have hints as to what the character is? For example, it is easy to select high reversed quotation mark ‛ when single left quotation mark ‘ is intended. I know there are tips in the Guiguts window itself, but I have always found it helpful to have confirmation before selecting the character.

srjfoo:
> Is it felt irrelevant to have hints as to what the character is? For example, it is easy to select high reversed quotation mark ‛ when single left quotation mark ‘ is intended. I know there are tips in the Guiguts window itself, but I have always found it helpful to have confirmation before selecting the character.

What context? In "Unicode Blocks" and "Commonly Used Characters" I see tooltips if I hover over a key. In the Help menu, there's also a list of compose sequences that includes the character, the sequence, and the Unicode name.

jjzsdp:
I regret I did not spell out the exact context. I thought the discussion was about adding Commonly Used Chars to Unicode. So I thought my question was regarding Unicode blocks/Commonly Used Characters; sorry I got it wrong. When using Unicode blocks/Commonly Used Characters, I see no hover over tooltips. I do not know how to use compose sequences.

windymilla:
That's odd, @jjzsdp - you're quite right that tooltips are very useful for the Unicode Blocks/Commonly Used Characters. I added them in GG2 to the same as in GG1. When you hover over one of the buttons, it should pop up the tooltip showing the unicode code point number and the name of the character. Here's an screenshot of it on my computer. I'm not sure why it's not working on yours - I presume other tooltips work OK? I tried various combinations of using the Unicode Blocks menu entry, the new Commonly Used Characters menu entry (which isn't yet merged in), using the dropdown in the dialog to select a different block, etc., but I couldn't make it go wrong. 

Perhaps if you restart, it will be one of those things that goes away all by itself 🤞 

image

windymilla:
@jjzsdp - one thing worth checking was whether you have turned off tooltips in the Preferences dialog, Advanced tab. "Show Tooltips" must be on for the commonly used characters tooltips to appear.

image

srjfoo (approved):

jjzsdp:
Yes, that was it, dreamt about it last night and woke up wondering if turning off tool tips affected everything ... hmm, needed for some things, not all. That's annoying. Thanks so much.

windymilla:
> Yes, that was it, dreamt about it last night and woke up wondering if turning off tool tips affected everything ... hmm, needed for some things, not all. That's annoying. Thanks so much.

@jjzsdp - yes, at the moment, tooltips are either on or off. I don't think most programs would let you select which tooltips you want and which you don't want. 

There are several types in GG - some let you know what the effect of various clicks would be, e.g. "Shift-click to show/hide line numbers"; some explain the meaning of settings, where there is too much information to fit into the label, e.g. "Index Main Entry Margin" in the Prefs Wrapping tab; some explain where there might be some ambiguity, e.g. "Add Vertical Line" and "Insert Vertical Line" in ASCII Tables, or where the user might need a hint about how to use it, e.g. "Type part of a command to filter the list" in the Command Palette; some explain why something isn't available, e.g. on Mac computers the Tear-off menu preference setting is disabled with a tooltip "Not available on macOS".

I suppose the Unicode/Common characters ones fit into the "too much information to fit into the label" category or maybe the "ambiguity" category in the quotes example you gave.

I would be open to adjusting things if someone comes up with a change that would suit most/all people, and doesn't make the preferences too complicated. 

In the meantime, I have a suggestion for you @jjzsdp:
1. Open Command Palette
2. Type "Tooltip"
3. Edit the "Tooltips (toggle)" command, and assign it a shortcut key, e.g. Ctrl+T or Ctrl+Shift+T
4. Now, when you open a dialog, like Commonly Used Chars, and think, "I'd like tooltips", press Ctrl+T, and it will start to show tooltips for you
5. If you later open another dialog, like Bookloupe, and find the tooltips annoying or distracting, press Ctrl+T, and it will stop showing them.

I hope that helps.

jjzsdp:
Thank you. I'm sure it would help if I were able/willing to use more shortcuts.

windymilla:
@jjzsdp - to try to get something that works for everyone, Sharon has made another suggestion which I will test out. It wouldn't involve tooltips, but would show the character name within the dialog, so I think it would be better than the current tooltip method, as it would be more responsive and no annoying popups.

jjzsdp:
Look forward to seeing how that goes, thank you both.

PR #1005, opened 2025-05-08
Investigate tk9 Mac scroll

windymilla:
Testing notes:

In Tk8, you should see a debug print "Failed to bind TouchpadScroll". It would be interesting to know if Tk9 gives the same message.

In Tk8 and Tk9, attempt to scroll the minimum amount in the image viewer, using the mousewheel.
You should see a debug print `delta: <nn>` in the shell window. When I scroll the mouse wheel one notch, I get a delta of +/- 120.
In Tk8, I think you should get a delta of +/- 1.
It would be interesting to know what Tk9 gives.

It's possible that using touchpad scrolling will give lots of messages `Touchpad: <nn>`. Again, it would be interesting to know what happens there. I wouldn't expect the touchpad to do any scrolling, just output debug messages at the moment.

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -822,6 +822,13 @@ class MainImage(tk.Frame):
             _, cm = process_accel("Cmd/Ctrl+MouseWheel")
             self.canvas.bind(cm, self.wheel_zoom)
             self.canvas.bind("<MouseWheel>", self.wheel_scroll)
+            try:
+                self.canvas.bind(
+                    "<TouchpadScroll>",
+                    lambda e: print(f"Touchpad: {e.delta}", flush=True),
+                )
+            except tk.TclError:
+                print("Failed to bind TouchpadScroll", flush=True)
 
         self.image_scale = float(preferences.get(PrefKey.IMAGE_SCALE_FACTOR))
         self.scale_delta = 1.1
@@ -1044,6 +1051,7 @@ class MainImage(tk.Frame):
 
     def wheel_scroll(self, evt: tk.Event) -> None:
         """Scroll image up/down using mouse wheel"""
+        print(f"delta: {evt.delta}", flush=True)
         if evt.state == 0:
             if is_mac():
                 self.canvas.yview_scroll(int(-1 * evt.delta), "units")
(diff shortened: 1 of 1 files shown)


srjfoo:
Python 3.13 with Tk9:

Touchpad: No scrolling, but lots of command line output. Scrolling up,the numbers were 65535 and up (high of 131071 with an outlier at 196604). Scrolling down, between 1 and 65538. Scrolling right, between -65536 and -1, and Scrolling left, between 65536 and 131072. Those numbers may not be the full ranges. (Actually, I went back and re-did this, and either I didn't record things right the first time, or something. Up and right gave negative numbers, left and down gave positive numbers.)

Mouse: No command line messages. The mouse did scroll, but so coarse-grained up/down as to be unusable (it scrolled from the top of the image all the way to the bottom, with no steps in between for a greatly zoomed-in image. Left-right, still choppy, but there were steps in between (with the pointer hovering over the horizontal scrollbar).

Python 3.11 with Tk8:

After `INFO - application.py:127 - Guiguts started` but before `INFO - application.py:165 - GUI initialized`, "Failed to bind TouchpadScroll". No such message in 3.13/Tk9.

Touchpad: prefix is `delta:`. Scrolling worked to the extent that it did happen, but the deltas mostly varied between 1 - 3 and -1 - -3. I was able to get a couple of 3-digit numbers by scrolling very fast. Scrolling very choppy and laggy, but it was working.

Mouse wheel: Horizontal scrolling didn't work at all. Vertical scrolling gave similar `delta:`-labeled numbers to the ones from the touchpad.

srjfoo:
Still focusing on Tk9 -- progress!!! 🎉 

Two-finger scrolling now works! And it's very fast (maybe too fast). BUT ... (yes, there's a BUT) ... it seems to be ultra-sensitive to multi-dimensional ooches. When I think I'm just scrolling up, I'm apparently also scrolling sideways. So, for instance, the image will go from top left to bottom right, relatively smoothly and quickly (I can see some choppiness, but it's much smaller increments, and moves much more quickly).

srjfoo (approved):
With `scroll_rate_denominator = 3` and `wiggle_tolerance = 2`, I think we're good to go on this one.

PR #1009, opened 2025-05-08
Updates relating to python 3.13

### LINKED ISSUE #1002: Fix whines from Python 13 black/flake/etc

windymilla:
Update various checking tools to more recent versions, and fix issues reported by them in the code.

Once merged, this should allow devs to use python
3.11, 3.12 or 3.13 without flake and its cronies whining.

Fixes #1002

Testing notes - will need `poetry install` and a quick check things still seem to work OK (i.e. don't crash), including rewrapping Word Distance Check. This should only affect devs, not users, so shouldn't matter whether you are using Python 3.11 or 3.13 to test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ python you installed above.
      ```bash
      poetry config virtualenvs.prefer-active-python true
      ```
+     You may also need to tell poetry explicitly which python to use
+     ```bash
+     poetry env use 3.13.2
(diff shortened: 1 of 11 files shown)


srjfoo (approved):
In addition to the two mentioned, I did some random testing both in 3.11 and 3.13, and didn't find any problems.

Issue #1008, opened 2025-05-08
Search & Replace limited tab selections


okrick:
In GG1's Search & Replace window, while the tab order wasn't ideal, it did allow navigation between most fields and selections. 

However, in GG2, the Tab key is limited to cycling only through the Search and Replace input fields, completely skipping checkboxes and button actions. I believe the tab order should prioritize input fields first, followed by buttons or checkboxes for better keyboard accessibility.

In my opinion, addressing this issue can be prioritized towards the end, just before GG2's release.

windymilla:
I've put this as an earlier feature than you suggested, @okrick, because I think it's definitely the kind of thing we want to see whether users like, or if it causes other problems. 
Online comments I've seen about the tabbing order is that it should generally follow the visual flow, typically left to right and top to bottom, so it's predictable for users. 
I don't think it will be much work to make tabbing go to checkboxes and buttons. Once we have the basic case, we can consider if there are special cases where changing the tabbing order from the "obvious" one is something we should do.

srjfoo:
Just a thought -- if the tab order makes more sense not left-to-right-top-to-bottom, would it make more sense to change the button order?

PR #1012, opened 2025-05-09
Allow tabbing to all buttons and checkboxes

windymilla:
To support those who prefer to use the keyboard

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -1128,15 +1128,16 @@ class Guiguts:
             tooltip="Click: Go to line\nShift click: Toggle line numbers\nShift right-click: Toggle column numbers",
             update=rowcol_str,
         )
-        the_statusbar.add_binding("rowcol", "ButtonRelease-1", self.file.goto_line)
+
+        the_statusbar.add_binding("rowcol", StatusBar.BTN_1, self.file.goto_line)
         the_statusbar.add_binding(
             "rowcol",
-            "Shift-ButtonRelease-1",
+            StatusBar.SHIFT_BTN_1,
             lambda: preferences.toggle(PrefKey.LINE_NUMBERS),
         )
         the_statusbar.add_binding(
             "rowcol",
-            "Shift-ButtonRelease-3",
+            StatusBar.SHIFT_BTN_3,
             lambda: preferences.toggle(PrefKey.COLUMN_NUMBERS),
         )
 
@@ -1145,12 +1146,12 @@ class Guiguts:
             tooltip="Click: Go to image",
             update=lambda: "Img: " + maintext().get_current_image_name(),
         )
-        the_statusbar.add_binding("img", "ButtonRelease-1", self.file.goto_image)
+        the_statusbar.add_binding("img", StatusBar.BTN_1, self.file.goto_image)
 
         the_statusbar.add(
             "prev img", tooltip="Click: Go to previous image", text="<", width=1
         )
-        the_statusbar.add_binding("prev img", "ButtonRelease-1", self.file.prev_page)
+        the_statusbar.add_binding("prev img", StatusBar.BTN_1, self.file.prev_page)
 
         the_statusbar.add(
             "see img",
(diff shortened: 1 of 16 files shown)


windymilla:
Assuming this is basically working, I suggest we merge it. There may be specific dialogs where we want to change the tabbing order. However, I suspect that will be very subjective.

I have made some changes to the S/R dialog. Each row contains an entry field, a Search (or Replace) button, a First (or R&S) button and an "All" button. They now tab across each row in a consistent way. Also, all the checkboxes above the entry fields tab in order, and finally, "In selection" and "Count". Whatever order you choose, you can think of situations where you would need to do several tabs to do a common operation.

For example, as I have it, to enter a Search string and a Replace string, you have to tab 4 times to get from Search field to Entry field. However, if you make it so that tabbing goes downward, i.e. from Search field to Replace field 1, then Replace field 2, then Replace field 3, then when you want to do the common operation of typing a search string and searching for it, you would have to tab 4 times to get from the Search field to the Search button. So :shrug:

srjfoo:
> For example, as I have it, to enter a Search string and a Replace string, you have to tab 4 times to get from Search field to Entry field. However, if you make it so that tabbing goes downward, i.e. from Search field to Replace field 1, then Replace field 2, then Replace field 3, then when you want to do the common operation of typing a search string and searching for it, you would have to tab 4 times to get from the Search field to the Search button. So 🤷

For my purposes, I would rather use keyboard shortcuts -- though if there are multiple replace fields, each one would have to have a shortcut, I guess? What about using `<tab>` to move across and `<ctrl>`+`<tab>` to move down, and `<shift>`+ whichever to move backwards?

srjfoo:
> Allow tabbing to all buttons and checkboxes ...

I haven't checked everything, but, so far, the only places I'm seeing tab navigation is in S/R, the Scannos check, and Bookloupe options (which is painful to deal with, because if you check or uncheck something, you're shunted immediately back to the main bookloupe window and either have to click back in the options pane if you want to unclick another, or navigate there by rotating through the windows).

I didn't try all dialogs, but I did try a selection of them including WF, Jeebies, Spelling, Word Distance Check, and the fixup dialogs. Page Separator Fixup appeared to do tab navigation, but the others did not.

Tried both in Python 3.11 and 3.13. Is there something special I needed to do? If so, I totally missed it! 😁

windymilla:
> For my purposes, I would rather use keyboard shortcuts -- though if there are multiple replace fields, each one would have to have a shortcut, I guess? What about using `<tab>` to move across and `<ctrl>`+`<tab>` to move down, and `<shift>`+ whichever to move backwards?

Good idea, but not practical, given how tabbing works, I'm afraid. It just has an order of widgets, and Tab takes you to the next one, and Shift+Tab takes you to the previous one. There is no concept of "across" or "down". If you use Tab a lot for navigating in other contexts, e.g. in the browser, while editing a comment on Github, you are basically at the mercy of that order. For example, to add this comment, which is the thing I am most likely to do next, I would have to press Tab 4 times to get to the Comment button.

I think what I have done in this PR (which is essentially to restore default tabbing behavior) will have to be acceptable as base behavior. In the future, we may want to add specific shadow commands for frequently used dialogs. As you suggest, we could add a shadow command for "Search/Replace, Focus in Search field", and "Focus in Replace Field" (the multiple fields aren't used commonly enough for them to need shadows). I don't think we should rush into that though.

windymilla:
> I haven't checked everything, but, so far, the only places I'm seeing tab navigation is in S/R, the Scannos check, and Bookloupe options (which is painful to deal with, because if you check or uncheck something, you're shunted immediately back to the main bookloupe window and either have to click back in the options pane if you want to unclick another, or navigate there by rotating through the windows).

I'll look at making Bookloupe options retain focus, instead of it focusing on bookloupe main window when you make a change. Although for me it's a simple Cmd/Ctrl+tab to return to the options window. I'm not sure whether I deliberately made it focus on bookloupe main, or if that's an inadvertent side effect, or indeed whether there's a good reason for wanting that.

> I didn't try all dialogs, but I did try a selection of them including WF, Jeebies, Spelling, Word Distance Check, and the fixup dialogs. Page Separator Fixup appeared to do tab navigation, but the others did not.

The problem in checker dialogs is that the focus is in the text window that contains all the messages. Tabbing out of a text window doesn't work by default (general Tk tab issue) so needs a specific bit of code adding. I'll look into doing that.

> Tried both in Python 3.11 and 3.13. Is there something special I needed to do? If so, I totally missed it! 😁
What you've tried is perfect, thanks. Nothing else to check for now.

windymilla:
@srjfoo - I think I've addressed those points, if you can check that they work on macOS, please.

windymilla:
Fixes #1008

srjfoo (approved):

PR #1013, opened 2025-05-10
Update README about Homebrew Python 3.12

windymilla:


README.md
```diff
@@ -263,11 +268,15 @@ previous version of GG2).
...
 brew install python@3.11 python-tk@3.11
 ```
+Alternatively, if you want to install Python 3.12 or later, you should install
+by download from [python.org](https://www.python.org/), not using Homebrew.
+This is because Homebrew installs Tk version 9 with Python 3.12, whereas Guiguts
+is currently configured to work best with Tk 8.6. So ensure that you either
+install Python 3.11, or non-Homebrew Python 3.12.
 2. Type `pip install guiguts` (or `pip install guiguts --upgrade` to upgrade from a
 previous version of GG2).
 3. Type `guiguts` to run GG2.
```
srjfoo, line 282: You need a blank line at the least preceding number 2 on line 289 and the number 3 on line 282. Otherwise everything runs together.

May I suggest slightly simplifying the macOS user install section (this is only a suggestion. If you don't like it, no worries -- the change request is only for adding the blank lines as mentioned above):

1. Install Python 3.11 and python-tk using Homebrew if you don't already have it:
For example:
```bash
brew install python@3.11 python-tk@3.11
```

\-\-***OR***\-\-

Install the current version of Python available for download from
[python.org](https://www.python.org/).

This is because Homebrew installs Tk version 9 with Python versions
later than 3.11, and Guiguts is currently optimized for Tk 8.6.

2. Type `pip install guiguts` (or `pip install guiguts --upgrade` to upgrade from a
previous version of GG2).

3. Type `guiguts` to run GG2.

srjfoo (changes requested):
I keep wanting to add more detail about the Tk8.6 vs Tk9 issue, but I realize that's probably TMI.

I wouldn't even mention Python version 12 at all anywhere, actually -- just say that at this time, homebrew installations should be limited to 3.11, and that at this point in time, any current version from python.org is fine.

README.md
windymilla: Thanks. I used your simplified version, but still referencing 3.11 and 3.12, rather than "current version". We are only recommending 3.11 and 3.12 for users for all platforms at the moment, although we believe 3.13 is OK. At some point we could update the readme to say any version from 3.11 to 3.13 (or possibly 3.14 once it is released in October)

srjfoo (approved):

PR #1027, opened 2025-05-13
Fix tabbing through statusbar/image viewer

### LINKED ISSUE #1026: Tab order is suboptimal when there are split text windows

windymilla:
Fixes #1026

The order is main text, peer window (if visible), status bar, image viewer (if visible and docked), back to main text. If image viewer is floating, it is not included in the tab loop from the main window, but has its own tab loop. You need to Ctrl/Cmd+tab or click in the image viewer to get focus in the window, then tabbing will just cycle through the image viewer buttons. In short, the floating image viewer should behave like a dialog as far as tabbing is concerned.

Testing notes - there are 6 cases to test:
1. No split text or image viewer
2. Just split text
3. Just docked image viewer
4. Just undocked image viewer
5. Split text and docked image viewer 6 Split text and undocked image viewer

Transition between regions is the most likely place for things to go wrong, e.g. at beginning/end of
status bar, or beginning/end of image viewer.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -956,7 +956,7 @@ class Guiguts:
         )
         view_menu.add_separator()
         view_menu.add_checkbutton(
-            "Image Viewer",
+            "Image ~Viewer",
             PrefKey.IMAGE_VIEWER_INTERNAL,
             self.show_image_viewer,
             self.hide_image_viewer,
@@ -1145,6 +1145,7 @@ class Guiguts:
             StatusBar.SHIFT_BTN_3,
             lambda: preferences.toggle(PrefKey.COLUMN_NUMBERS),
         )
+        the_statusbar.set_first_tab_behavior("rowcol")
 
         the_statusbar.add(
             "img",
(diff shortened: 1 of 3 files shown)


srjfoo:
> Testing notes - there are 6 cases to test:
>
> 1. No split text or image viewer

👍 

> 2. Just split text

👍 

> 3. Just docked image viewer

👍 

> 4. Just undocked image viewer

👍 👎  -- Or, at least, not what I was expecting. If I use (macOS) `cmd`+`~` to navigate to the floated window, it works fine. If this is as intended, that's fine.

> 5. Split text and docked image viewer

👍 

> 6. Split text and undocked image viewer

Ditto number 4. 


So, the responses to 4 and 6 were based on my expectation that tabbing would move from text pane to image pane and back without having to `cmd`+`~` between windows, since nothing was mentioned about needing to. In actuality, it's not necessarily a bad idea for them to be separate, it just doesn't match the docked version.

srjfoo (approved):

srjfoo:
> If image viewer is floating, it is not included in the tab loop from the main window, but has its own tab loop.

I obviously missed this when I was reading the instructions -- so, yes, it's behaving as it should :D

Issue #1006, opened 2025-05-08
Image viewer shortcuts


windymilla:
Add zoom-type commands as shadows

Check that it works in both docked and undocked mode.

PR #1035, opened 2025-05-17
Add shadow commands for image zoom

windymilla:
See all the "Image" shadows by typing "Image" into Command Palette.

This adds zoom in/out (Ctrl/Cmd-plus/minus) and
fit to height (but not autofit) (Ctrl/Cmd-zero).
Those were the default shortcuts we already had, but this adds them as shadow commands, so the user can reassign them.

Test in docked and undocked state, and when image
is not shown.


windymilla:
Fixes #1006

srjfoo (approved):

PR #1036, opened 2025-05-17
Alternative apply

windymilla:
Switch to space to activate some Treeviews

In Page Labels, use Space instead of Return.
In Unicode Search allow Space or Return.

Only enable Apply/OK in Page Labels when needed


srjfoo (approved):
Looks good as far as I can tell, thanks!

PR #1072, opened 2025-05-29
Reduce width of active part of checkboxes in SR

windymilla:
So users don't accidentally click them.


srjfoo (approved):
This works great for me. I hope it isn't an inconvenience for others.

PR #1086, opened 2025-06-02
Make Prev/Next FN work if anchor message selected

### LINKED ISSUE #1068: Footnote Fixup: Prev FN and Next FN always result in "No footnote selected".
This file does have some footnotes with multiple anchors, causing SEQUENCE warnings.

The navigation buttons don't work intermittently, usually if the selection is before the warnings, less so after, it seems. Inconsistent either way. But **always** when Next FN is clicked after first opening the dialog.

So best repro:
1. Open file
2. Run Tools⇾Footnote Fixup
3. Click Next FN

- **Side issue:** If the error popup is not closed, it keeps popping up a new one every time the navigation button is clicked, resulting in multiple popups to close. They stay open when the FN Fixup dialog is closed. Perhaps they should be closed on navigation click, definitely on dialog close.

- **And a question:** with the multiple anchors, is there a way to clear the SEQUENCE warnings so that I can reindex. etc.? Or some workaround?

[norse.txt](https://github.com/user-attachments/files/20463648/norse.txt)

Guiguts version: 2.0.0b1

Python version: 3.13.3 (tags/v3.13.3:6280bb5, Apr  8 2025, 14:47:33) [MSC v.1943 64 bit (AMD64)]
Tk/Tcl version: 8.6.15
OS Platform: Windows-11-10.0.26100-SP0
OS Release: 11

windymilla:
Previously if the selected line in the FN dialog was an anchor line rather than a footnote line, you couldn't do Prev/Next FN. Selecting an anchor line is now the same as selecting the footnote it belongs to.

Fixes #1068 - see issue for example file and description of how bug was noticed by user.


rtonsing (approved):
Just a quick check with the original file, but it looks good.

srjfoo (approved):

PR #1087, opened 2025-06-02
Add Browse button to image viewer

### LINKED ISSUE #922: Allow loading of (almost) any image into the image viewer
Particular use case (but there may be others): if a file doesn't have a .json project file, it has no link between pages and png files, so it's not possible/easy to load a png into the viewer. 

Maybe a "Browse" button (between the manual  "<" and ">" buttons?)

Note that if auto image is on, then the automatic "return to auto image" feature might cut in if this feature was (ab)used to load other files, but that's up to the user to turn off auto image if they don't want that behavior.

windymilla:
Button is to the right of the filename label in the image viewer.

Only scans should be inverted, not images, like illos, that are stored as RGB files.

Fixes #922

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -898,6 +898,10 @@ class MainImage(tk.Frame):
             lambda: self.image_zoom_to_height(disable_autofit=True),
             "Cmd/Ctrl+0",
         )
+        menubar_metadata().add_button_orphan(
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Seems to work as advertised.

Possible future enhancement: shadow commands to allow the user to add keyboard navigation for images?

PR #1089, opened 2025-06-03
Drag n drop to main text or image viewer

windymilla:
Intended to work from File Manager/Chooser/Nautilus etc.
If several files are dropped, just the first will be opened.


windymilla:
Too fragile - not supporting Tk9 yet, etc., so closing

PR #1098, opened 2025-06-08
Fix mousewheel scrolling in image

windymilla:
Broken for Mac users in #1005 by comparing Tk version to 3.7 instead of 8.7!

Testing notes: this won't make it any smoother than before, nor fix any other problems, but should revert to the "basically works" situation pre-#1005. Apologies for the typo!


tangledhelix (approved):
As you say - seems to revert back to the previous behavior (which is still broken). I'll see if I can take a stab at that behavior. Based on my observation of the behavior I'm wondering if it's an interaction with when the loop goes idle. If you swipe left/right on an image over and over without stopping nothing happens; if you stop, a beat goes by, then it moves. Almost as if it were waiting for the event loop to calm down.

srjfoo (approved):
Thanks!

PR #1099, opened 2025-06-08
Allow user to customize colors

windymilla:
Add first few colors
1. Main text
2. Main text in hi-contrast mode
3. S&R match highlighting

Incidentally, **_horizontal_** scrolling should now work in Unicode Block and the new Prefs tab by holding Shift while scrolling the mousewheel.


srjfoo (approved):

src/guiguts/misc_dialogs.py
```diff
@@ -431,6 +445,184 @@ def add_orphan_commands(cls) -> None:
...
+
+            desc.bind(
+                "<Button-1>",
+                lambda _, cb=click_callback: cb("foreground"),  # type:ignore[misc]
+            )
+            desc.bind(
+                "<Button-3>",
+                lambda _, cb=click_callback: cb("background"),  # type:ignore[misc]
+            )
+            desc.bind(
```
tangledhelix, line 518: As discussed earlier, right-click on macOS doesn't work. Adding `<Button-2>` binding would fix that. If `<Shift-Button-1>` was only added to address that shortcoming, then potentially it could be removed.

tangledhelix:
Seems to work fine on Linux, including right-click to edit background colors.

src/guiguts/misc_dialogs.py
windymilla: Thanks - we do add the button 2 binding when the code attempts a button 3 binding in some places, but I forgot to include it in this case. I'll add it.

windymilla:
Latest commit binds mouse-1 to choose foreground and Shift-mouse-1 and mouse-3 (plus mouse-2 and Control-mouse-1 on Macs) to choose background.
Also, color chooser dialog title should now read "Choose foreground/background color" as appropriate.

srjfoo (approved):

srjfoo (approved):

PR #1104, opened 2025-06-09
Another attempt at Linux image scroll

windymilla:


srjfoo (approved):

tangledhelix (approved):
On Mac, there seem to be no changes (as expected), all good.

On Linux, the vertical scrolling is working, and works pretty well! But horizontal scrolling does nothing at all. It seems to not lock up the event loop or have negative effect, but it just doesn't work.

I added a print to show me `evt.num` but it doesn't seem to do much if I try to horizontal scroll. It seems to show just 4 or 5 as the value (which, I think, are the values for vertical up/down scroll). It's almost like the event doesn't fire.

I'm going to approve this because it does work for vertical, and maybe that event doesn't work on Linux. If you have any other ideas I'm happy to test, but I'll call it done for now otherwise.

PR #1124, opened 2025-06-13
Fix Shift-tab on Linux

windymilla:
There were several places Shift-tab went wrong on Linux.

Now all attempts to bind to Shift-tab go through a common routine that also attempts to bind to Linux version of the key: ISO_Left_Tab

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -39,6 +39,7 @@ from guiguts.widgets import (
     focus_prev_widget,
     ThemedStyle,
     TreeviewList,
+    bind_shift_tab,
 )
 
 logger = logging.getLogger(__package__)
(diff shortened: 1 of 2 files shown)


windymilla:
@srjfoo, including you as a reviewer so you can check I haven't broken Mac tab/shift-tab behavior, although I know you'll have to leave Linux testing to @tangledhelix

srjfoo:
I just tested in the main window with text (both split and un-split) and image viewer docked, WF, the S/R Find All, and bookloupe. Tab and shift tab seem to work just fine, but I would have expected the heavy border to be for the selected pane -- not the inactive pane. Also, the border looks just a mite heavy for me. In fact, it looks heaver (though that may be an optical illustion) than the heavy border on the main text.

I spotted it first in WF, and checked bookloupe to make sure that it was doing the same for both. And then checked Find All. For all of them, the heavy highlighting is on the results pane for every choice except when it's the currently selected spot. Is this what's desired?

srjfoo (approved):

tangledhelix (approved):
Does what it says on the tin. Tested both Mac & Linux.

PR #1125, opened 2025-06-13
Miscellaneous theme changes

### LINKED ISSUE #1116: Border isn't quite right on main text widget
Try tabbing between the split text widgets and see the border doesn't behave as expected.
Check light and dark modes

windymilla:
1. Main and peer text border highlighting was reversed - now highlights correctly when focused
2. Added some padding to left and right for text window
3. Made focus highlight color more obvious (i.e. light for dark theme, dark for light theme) - test by tabbing to, or clicking on, a button.

Fixes #1116

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -738,6 +738,7 @@ class MainText(tk.Text):
             spacing1=self["spacing1"],
             insertwidth=self["insertwidth"],
             wrap=self["wrap"],
+            padx=self["padx"],
         )
         self.peer.bind(
             "<<ThemeChanged>>",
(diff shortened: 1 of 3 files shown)


srjfoo (approved):
I've gone ahead and approved it (thanks for the padding changes!), but one thing you might want to look at before merging is the light theme. I did check both light and dark (via default, by changing the OS theme) and it seemed to me that, while the solid outline for the selected text pane (or results pane in a widget) is nicely visible in the light theme, the dotted borders for the buttons is not. It's still very subtle, though it's fixed and nicely visible on the dark theme. Is this because light on dark stands out more, but dark on light tends to get washed out by the dark bacground?

tangledhelix (approved):
I agree with what @srjfoo said, it looks great in Dark theme, but the button highlights are a bit subtle in Light mode when you tab-navigate. (Both in Mac and Linux.)

PR #1129, opened 2025-06-16
Add regex ability to store persistent variables

windymilla:
When using `\C...\E` can now use the following 2 functions:
1. cset("key", value) - stores (and returns) the given value using the key string (must be in quotes)
2. cget("key") - returns the previously-stored value

Note Python syntax
`<value if true> if <test> else <value if false>`

Warning: you can execute dangerous stuff if you want to, e.g. `\C exit() \E` will exit the program. How to reformat your hard disk is left as an exercise for the reader.

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -38,6 +38,9 @@ MARK_END_RANGE = "SearchRangeEnd"
 PADX = 2
 PADY = 2
 
+# Passed into eval call when `\C...\E` used in regex replacement
+lglobal = {}
+
 
 class NoMatchFoundError(Exception):
     """Raised when no match is found for the search string."""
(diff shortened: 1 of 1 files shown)


windymilla:
For testing:

[index_example.zip](https://github.com/user-attachments/files/20763856/index_example.zip)

This file contains a multi-volume index. The following S/R converts page numbers to hyperlinks for the correct volume by storing the volume number each time it finds one, then getting it when it wants to create a hyperlink.

Search `(([> ])(i|ii|iii|iv|v)(\.))|(\d+)(?=[-\.,;<]| <i)`
Replace `\C '<a href="vol_' + cget("C") + '#Page_\5">' + '\5' + '</a>' if '\3' == '' else '\2' + str(cset("C", '\3')) + '\4' \E`

srjfoo (approved):

PR #1138, opened 2025-06-18
Make Page Labels dialog refresh when 'Img' column is clicked

rtonsing:


windymilla (approved):
Thanks @rtonsing - it wasn't clear to me initially what this was for, though I understand now.

In `master` if you click on "Img" or "Label" column, nothing happens. With this fix, you jump to the relevant page break in the file.

I will take the liberty of editing the PR title to make it clearer to me at least.

Thanks for fixing this!

srjfoo (approved):
If you're wondering why I never spotted this, I literally never click in either column! I use home or end or scroll with the arrow keys! 😆

PR #1139, opened 2025-06-19
Add File-->Content Providing menu 

windymilla:
1. Import Prep Text Files
2. Export as Prep Text Files
3. CP Character Substitutions

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,6 +18,11 @@ Checkers
     
 .. automodule:: checkers
  
+Content Providing
+=================
+    
+.. automodule:: content_providing
+ 
 File
 ====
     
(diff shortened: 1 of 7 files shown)


srjfoo:
> @srjfoo I've extended the list of character substitutions to match the one you gave me - thanks. Just a quick check would be great.

Looks good -- the only one that didn't get converted was the zero-width space (U+200b), and that one got entered into my test set accidentally (and isn't in the list in the wiki). It would be rejected by `add_files.php` anyway because it isn't in any of the character suites. And it really doesn't make sense for OCR programs to identify anything as a zero-width space!

srjfoo (approved):

PR #1149, opened 2025-06-22
Sort CSS class reports from PPhtml

windymilla:
PPhtml reports classes used but not defined, and defined but not used. Previously, they were just listed alphabetically. This commit lists each on its own line with the location in the file where it first occurs, allowing the user to click on the message and correct the error.


windymilla:
Testing notes: change/add/remove class names from CSS block or HTML elements and see if the correct ones are reported, and correctly linked.

srjfoo (approved):

PR #1154, opened 2025-06-23
Flake8 fixes for Python 3.13

tangledhelix:
Fixes a few Flake8 complaints that happen in Python 3.13.

They might happen in 3.12 too; I'm not sure. They didn't happen in 3.11.


windymilla (approved):
Thanks - sorry about leaving those spaces in. They obviously made it capable of checking the spacing around operators within the curly brackets of an f-string.

PR #1163, opened 2025-06-26
Stop Proofer Comment deletion removing edits

### LINKED ISSUE #1161: "Delete" in Proofer Comment tool can silently undo edits
In the Proofer Comments tool, the "Delete" and especially the "Delete & Hide" button can silently destroy edits and the user may not notice. (Particularly in the "& Hide" usage they may not notice.)

To reproduce, create a small file containing a line like this:

```text
foo bar ba[**baz?] woot
```

Now, position the cursor just left of the opening bracket `[` and type `z`, resulting in:

```text
foo bar baz[**baz?] woot
```

Finally, click "Delete" in the PC tool window. You should be left with this incorrect outcome:

```text
foo bar ba woot
```

It happens with a longer edit the same way, oddly; so I guess it's not merely memorizing the start & end index of the comment and deleting that? If I make this edit:

```text
foo bar ba longer edit[**baz?] woot
```

And click "Delete", then the outcome is the same.

windymilla:
If the PPer corrected the word before a proofer comment by typing letters immediately before (or after) the comment, then used the Delete function of the Proofer Comment dialog, their edit could be lost or corrupted.

Avoid by using RIGHT gravity for the mark before the comment and LEFT gravity afterward, which it the
opposite to default checker behavior.

Fixes #1161

### DIFF
--- a/src/guiguts/checkers.py
+++ b/src/guiguts/checkers.py
@@ -283,6 +283,7 @@ class CheckerDialog(ToplevelDialog):
         show_process_buttons: bool = True,
         show_all_buttons: bool = True,
         match_on_highlight: Optional[bool] = False,
+        reverse_mark_gravities: bool = False,
         **kwargs: Any,
     ) -> None:
         """Initialize the dialog.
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: see linked issue for example.

Also worth checking that manually editing "he" to "be" after a Jeebies check by deleting the "h" and typing "b", that using the Fix button in the Jeebies dialog still works to swap between "he" and "be".

In the proofer comment case, we want to preserve the edit made outside the matched text, but in the Jeebies case, we want to preserve the edit made inside the matched text.

PR #1186, opened 2025-06-29
Don't trap GG shortcut sequences in WF

### LINKED ISSUE #1184: Cmd-F (find/search) when Word Frequency is focused
With Word Frequency focused, Cmd-F does not take me to the search panel as I expected. Instead it ignores the meta-key (Cmd) and hears only "f" and jumps me to what I assume must be the first entry in the list beginning with "f".

I have no objection to the jumping if "f" is pressed, but habitually using Cmd-F while doing other things, this was not expected (and I then had to locate where in the list I had come from, as there was no way to jump back there).

windymilla:
Simple keystrokes navigate the WF list, e.g. "x" goes to the first word beginning with "x".

It was also trapping combinations like "Alt-d" and "Cmd-s". Allow these to pass through for default handling.

Fixes #1184

Testing notes:
1. Edit the main text so that file is modifed and needs saving.
2. Run WF - all words, and see how key strokes navigate up & down the list.
3. With focus still in WF, press Cmd+s to save file

### DIFF
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -631,6 +631,9 @@ class WordFrequencyDialog(ToplevelDialog):
         """
         if not event.char:
             return ""
+        # If modifiers other than Shift & Caps Lock, don't goto word - allow default processing
+        if int(event.state) & ~(0x0001 | 0x0002) != 0:
+            return ""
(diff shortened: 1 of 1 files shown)


tangledhelix (approved):
Does what it says on the tin. Thanks!

Issue #1192, opened 2025-06-30
Command Palette Surround Selection With, Apply Setting is intermittent


okrick:
1. Opened the Command Palette.
2. Set "Surround Selection With" to Ctrl-R.
3. Set "Surround Selection With, Apply" to Ctrl-Shift-R.
4. Used Ctrl-R to set <I> ... </i> as the surround selections.
5. Used Ctrl-Shift-R to apply the settings to some text.  It worked.
6. Did some Search & Replace actions.
7. Used Ctrl-Shift-R again to apply the settings to some text.  It did not work.
8. Checked the setting by pressing Ctrl-R.  It is still set to <I> ... </i> as the surround selections.
9. Tried Ctrl-Shift-R again.  Now it works. 
10. The "Surround Selection With, Apply" only works for consecutive actions, but not after some intervening actions.

okrick:
Yes, I did.  I'll try to remember not to close that window.  I've been in the habit of closing all of the various extra windows in GG2 once they've been used..

PR #1194, opened 2025-06-30
Make Surround Selection Apply work without dialog

windymilla:
Shadow command for Surround Selection did nothing if the dialog had been dismissed. In fact, it could work perfectly well without the dialog, using the most recent before/after text (like Find Next for S/R dialog).

This may have broken shadow commands generally, since it required some under-the-hood changes to make this type of shadow command work without the dialog. It means it will be easy to do similar shadow commands in future.

Things to try: the new Checker dialog commands, like "Hide"; Scanno Checker commands, like "Next Scanno". These do require the dialog to be visible, but should work the same as they do in master.


windymilla:
Fixes #1192

srjfoo (approved):

PR #1208, opened 2025-07-04
Fix HTML hanging

### LINKED ISSUE #1207: HTML Generator never completes with file
I don't see any unmatched markup or other issues, but HTML Generator never completes. Closing editor shows "Python is not responding".

[landscape.zip](https://github.com/user-attachments/files/21048489/landscape.zip)

windymilla:
Minor HTML prettifying bug where it inserted 5 newline characters but only advanced its current location by 3 lines, thus checking the same line for adding the chapter div over and over again.

Fixes #1207


rtonsing (approved):
All good now.

PR #1231, opened 2025-07-08
Output warning if non-utf8 file loaded

### LINKED ISSUE #1229: Request: Notify when file encoding is changed
It appears that:

- If a file is not plain UTF-8, it is automatically converted.
- If a file does not have Windows line endings (CR LF) it is converted.

and I agree with that.

But I would like to be notified when the file is loaded, that it needed conversion.

I don't know how much others would care, but this would be useful to me when whitewashing.

windymilla:
Will never affect most people. Note that warnings are output in the command window and the message log.
There will not be a popup window.

Fixes #1229

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1402,6 +1402,7 @@ class MainText(tk.Text):
             with open(fname, "r", encoding="utf-8") as fh:
                 self.insert(tk.END, maintext().reverse_rtl(fh.read()))
         except UnicodeDecodeError:
+            logger.warning("File failed to open as utf-8, attempting as iso-8859-1")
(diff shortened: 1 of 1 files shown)


rtonsing:
I get a message for an ANSI file, but not UTF-8-BOM.

I guess that is some help, but it means I need to explicitly view the message log or un-minimize the command window.

windymilla:
> I get a message for an ANSI file, but not UTF-8-BOM.
> 
> I guess that is some help, but it means I need to explicitly view the message log or un-minimize the command window.

You didn't mention BOM in your issue ;)

I didn't think it warranted a popup error, but I suppose it could be argued that 90% of users will never experience it anyway.

I'll take another look (tomorrow)

PR #1245, opened 2025-07-11
Fix url-opener command on Linux

tangledhelix:
As [discussed in the DP forums](https://www.pgdp.net/phpBB3/viewtopic.php?p=1370147&sid=8b1785d8837f3d81cc8ec05e27c7f0e2#p1370147), the `open` command doesn't work in all Linux distributions.

My research indicates that `xdg-open` is normally installed on Linux. So this PR sets `xdg-open` as the default for Custom menu commands on addition, and also uses it to open URLs when entered in the Custom menu.


srjfoo (approved):
Confirmed that it doesn't seem to break macOS

windymilla (approved):
Works fine in WSL - I fixed the black complaint and will get it merged

PR #1246, opened 2025-07-11
Update Linux install instructions

tangledhelix:
At least some Linux distributions won't let you do global installs with `pip` so as to avoid conflict with the system packaging machinery. Using pipx works around this by auto-managing a virtual environment. In my research it seems to be the case that for commands you install to run (such as guiguts), this is the preferred approach on modern Linuxes like Ubuntu 24.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ previous version of GG2).
 Python needs to be version 3.11 or above, and GG has been tested in
 versions up to 3.13.
 
-1. Install python, pipx, and Tk. Note that on some Linux distributions, the version number for `idle-python3.12` may differ slightly.
+1. Install python, pipx, and Tk. Note that on some Linux distributions, the version number for `idle-python3.12` may differ slightly, e.g. as of this writing Debian 12 would require `idle-python3.11`.
 
     a. `apt-get update`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -272,8 +264,8 @@ poetry env use /usr/local/bin/python3
    * Example from Ubuntu 22.04 -- adapt to your own Linux distro. Python should
      be version 3.11 or above, and GG has been tested in versions up to 3.13.
      ```bash
-     sudo apt install python3.11 python3-pip python3-tk idle-python3.11 git
-     sudo python3.11 -m pip install poetry
+     sudo apt install python3.11 python3-pip python3-tk idle-python3.11 git pipx
+     sudo pipx install poetry
```
srjfoo, line 268: In line 272 below, you say
> * If this does not work, try using `pipx` in place of `pip` in the above commands.

Do you need that anymore, since you change the command to `pipx`?

windymilla (approved):
Apart from resolving Sharon's comment, this all looks good to me.

README.md
tangledhelix: Good catch; fixed that by deleting that line.

windymilla (approved):

PR #1256, opened 2025-07-12
Don't allow HTML autogen on unnamed file

### LINKED ISSUE #1254: Running HTML autogenerate on a never-saved file causes an exception
1. Open GG
2. Paste in a few lines of text
3. Don't save, so there is no filename
4. DO HTML autogenerate

It should really call `tool_save()` before running autogen

windymilla:
Like other tools, requires user to save file first.

Note that check is disabled when GG is run with `-d` flag. Although I've stopped it from crashing for testers, they won't be able to re-load the backup file, since there won't be one, because GG had no filename to save it as. Testers can use undo under these circumstances, but note the usual warnings about undo and page break locations.

Fixes #1254

### DIFF
--- a/src/guiguts/html_convert.py
+++ b/src/guiguts/html_convert.py
@@ -13,6 +13,7 @@ import regex as re
 from guiguts.data import html
 from guiguts.file import the_file
 from guiguts.maintext import maintext
+from guiguts.misc_tools import tool_save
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Tested both with and without the `-d` flag.

One amusing side-effect that shouldn't have surprised me is that you also can't view the generated HTML file without saving it first. 😁

PR #1276, opened 2025-07-17
Improve Illo/SN movement over multi-blanklines

### LINKED ISSUE #1273: Sidenote fixup at chapter heading leaves spacing wrong
Described [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1370681#p1370681), if a sidenote is moved up to a chapter heading, it ends up above the double blank line instead of below.

(Also in that forum post is an unreproduced exception traceback)

windymilla:
Previously, it skipped over all blank lines in a block when moving illo/sn up. It is less likely that this is the desired position when moving up, i.e. to want the sidenote/illo to have multiple blank lines below it.

So, this commit defaults to just moving to the first blank line upwards. If the user wants to move further they can click "Up" again. Requiring an extra click in the less frequent circumstance seems preferable to not being able to move the illo/sn where you want to.

Although this means several clicks would be necessary to move an illos/sn past a chapter heading, in practice, it is uncommon to want to move them to a different chapter.

Fixes #1273

### DIFF
--- a/src/guiguts/illo_sn_fixup.py
+++ b/src/guiguts/illo_sn_fixup.py
@@ -363,66 +363,6 @@ class IlloSNChecker:
         )
         return
 
-    def advance_up_to_first_bl_of_block(self, line_num: str) -> str:
-        """Move insert point to the top of the block of blank lines.
-
-        The insert point will not move if the 'block' is a single blank line.
-
-        Arg:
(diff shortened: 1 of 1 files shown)


windymilla:
Also now fixes two bugs discovered during review

PR #1279, opened 2025-07-18
Add Dutch, Portuguese & Spanish dictionaries

### LINKED ISSUE #1266: Add more dictionaries?
A new PPer I'm working with asked about LOTE support in Guiguts. Aside from bundling French and German dictionaries, there isn't very much. It seems easy enough to include more dictionaries (it was Dutch in this case).

Surveying current projects in the rounds, for example, there are a number of Dutch projects in F2 that will enter PP next. Some Spanish and Portuguese are also in the rounds currently. Perhaps those are good candidates for languages to include dictionaries for. I'm guessing some of the squirrels (aside from me) have better information on what LOTE tend to show up the most.

(For the time being, I gave the PPer a Dutch dictionary supplied by @windymilla and explained how to install it locally in their ggprefs directory.)

Dan

windymilla:
These cover the main languages being used by known GG2 users.

Total download size of wheel/zip file will be 8.5MB.

Fixes #1266


tangledhelix (approved):
Other than the filename issue (swap `_user` for `_default`), which I think you're aware of already, this looks good.

srjfoo (approved):

PR #1285, opened 2025-07-20
Avoid exception with Cmd-w and pulldown menu

### LINKED ISSUE #1280: Exception on Cmd-w if pulldown list is opened
This is a bit of an edge case, but it did throw an exception so I'll file it. This happened on Mac; I didn't test other platforms.

To reproduce:
* Open the Settings / Preferences dialog
* Click the Theme pulldown to pop up the list
* Do not choose a theme - just leave the list open
* Try to close the window with Cmd-W (or Win/Linux equivalent key for window close)

```
23:55:21: ERROR - root.py:58 - Tkinter Exception
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/tkinter/__init__.py", line 2068, in __call__
    return self.func(*args)
           ~~~~~~~~~^^^^^^^
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/maintext.py", line 1120, in break_func
    func_ret = func(event)
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/maintext.py", line 4660, in <lambda>
    maintext().key_bind(key_event, lambda _: callback(), bind_all=bind_all)
                                             ~~~~~~~~^^
  File "/Users/dan/repos/dp/guiguts-py/src/guiguts/application.py", line 344, in close_command
    focus_widget = maintext().focus_get()
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/tkinter/__init__.py", line 806, in focus_get
    return self._nametowidget(name)
           ~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/tkinter/__init__.py", line 1677, in nametowidget
    w = w.children[n]
        ~~~~~~~~~~^^^
KeyError: 'popdown'
```

windymilla:
This fix should stop the exception, though it may mean that Cmd-w doesn't close the expected window if a pulldown is open - I don't see a way round that.

Fixes #1280

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -341,7 +341,10 @@ class Guiguts:
     def close_command(self) -> None:
         """Close top-level dialog with focus, or close the file if it's
         the main window."""
-        focus_widget = maintext().focus_get()
+        try:
+            focus_widget = maintext().focus_get()
+        except KeyError:  # Weird stuff allowed to happen if dialog being closed
+            focus_widget = None
         if focus_widget is None:
             return
         focus_tl = focus_widget.winfo_toplevel()


srjfoo (approved):

tangledhelix (approved):
Works now on Mac:
- cmd-w closes the window
- but if pulldown is open, it does nothing
- no exception thrown

On Linux it's the same as before, and alt-F4 closes the window whether the pulldown is open or not.

PR #1289, opened 2025-07-21
Prepare for beta-3

windymilla:


srjfoo (approved):
The two comments I've left are mostly important in the announcement in the forums, I think. Definitely not show-stoppers.

Issue #944, opened 2025-04-18
Unicode Blocks jump-to


okrick:
Tools=>Unicode=>Unicode Blocks:  

It would be useful if the "Unicode Blocks" selection in the "Tools=>Unicode=>Unicode Blocks" menu had a jump-to feature. For instance, if I press "s", it would be great if the selection would immediately move to the "Sinhala" block. Even more efficient would be the ability to type the beginning of a block name quickly, so pressing "sup" would take me right to "Superscripts and Subscripts".

PR #1306, opened 2025-07-27
Add keystroke character search to WF & checkers

windymilla:
Typing a single key searched for the first occurrence of a
word beginning with the letter. Extend this to checker
dialogs. Also, if multiple keys typed less than 1 second apart,
search for message beginning with whole string, like an
incremental search.

Since checkers often contain longer messages than WF,
e.g. context of error, as well as headings, add a "Full search"
option (next to up/down arrows) that searches whole of
message text (not row/col, nor red error type) rather than
only matching at the start of the text.

Example: running pptxt, type "ellipsis" to jump to section
describing ellipsis checks.


srjfoo (approved):

windymilla:
My mistake - this does not address #944, but is using the same idea as will be used for that.

PR #1315, opened 2025-07-28
Allow user to drag selected text to another place

windymilla:
Initial commit to gather feedback...

1. Only drags the first bit of a column selection - one possibility instead would be to disable dragging if there's a column selection
2. If you drag below the bottom of the screen, it still works, but appears to select all the intervening text until you release the mouse, when the dragged text should appear in the correct place. Again, may be possible to fix.
3. Dropping should drop immediately before the character the mouse is released over, unless at the very right of that character, in which case it should go to the right.

Testing:
Drag small bits of text left and right, as well as larger chunks up or down.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1032,6 +1032,13 @@ class MainText(tk.Text):
                 "<B1-Motion>", self._autoscroll_callback, add=True, bind_peer=True
             )
 
+        def leave_event(_: tk.Event) -> str:
+            if self.dragging:
+                return "break"
+            return ""
+
+        self.bind_event("<Leave>", leave_event, add=True, bind_peer=True)
+
         # get the current bind tags
         bindtags = list(self.bindtags())
 
@@ -2100,10 +2107,7 @@ class MainText(tk.Text):
         """End dragging of text, dropping it if appropriate."""
         if not self.dragging:
             return ""
-        self.dragging = False
-        self.config(cursor="")
-        self.peer.config(cursor="")
-        self._delete_drag_window()
+        self.cancel_drag_sel()
         idx = event.widget.index(f"@{event.x},{event.y}")
         self.set_mark_position("seldroptarget", IndexRowCol(idx))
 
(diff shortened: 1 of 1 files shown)


srjfoo:
> My take on this may be the unpopular one, but I would be happy to not have this dragging behavior at all. I've always disliked it in every app as it quite often moves things when I just want to select different text. My personal pet peeve, maybe...

I like it, and miss it when I don't have it, but ... having tested it just now, I'm seeing that there is potential for confusion, especially if the users are not familiar with it. It won't break my hear if we opt not to implement it.

tangledhelix:
I'll fully own and admit that I have a personal bias against this one. But, I'll try to set that to the side and think more objectively.

> but ... having tested it just now, I'm seeing that there is potential for confusion, especially if the users are not familiar with it.

This right here is what I would be concerned about. Just like with Ctrl-{D,K,T}, my worry would be that someone not familiar with it would accidentally drag some text from A to B and not notice it happened.

If this is implemented, my suggestion is to lock it behind a checkbox in the Advanced tab and default to off (for that reason). Since it's a mouse operation, it wouldn't be a good candidate to treat like we've done to Ctrl-K and Ctrl-D, using shadow commands and bindings.

windymilla:
I've pushed a few improvements, some of which mitigate the danger of dragging without realizing (points 2, 3 & 4):

1. Make it work in peer window as well
2. Show insert point with insert cursor during dragging
3. Change mouse cursor shape while dragging
4. Drag (potentially truncated) preview of text with mouse
5. Allow Cmd/Ctrl dragging (copy not move) text to location
immediately before/after selection, e.g. `abc` --> `abcabc`
6. Disable dragging if selection is a column selection

On Windows and I believe Macs, most apps allow dragging of text, so I think users will generally be aware of that it's possible. For example, in this window I can select a word and drag to copy it to another place. To drag something a user has to select some text, then separately click inside the selected area, then drag the mouse outside the selected area.

Although I didn't suggest this feature, I think it is a valuable addition if it does not introduce bugs in other drag operations.

tangledhelix:
Doing a basic select & drag seemed to work for me. Also worked in the peer. However, attempting to drag from the main to the peer, or from the peer into the main, made it go kind of crazy. It's scrolling through the whole file, auto-img is going nuts...

Ideally, dragging from one pane to another would "just work," but in case that isn't possible, perhaps the drag operation should be canceled if the cursor crosses the pane boundary?

PR #1316, opened 2025-07-29
Implement "replace identicals" feature

### LINKED ISSUE #609: When using a regex in "Search", add an option to "Replace all identicals"
Enhancement suggestion:

When using a regex to find patterns, it would be a great timesaver in some situations to be able to tell GG to replace all "identical" matches to the string that just matched the current pattern.  For example:

```
S: <i>(([^A-Z]|\n)+?)</i>
R1: <em>$1</em>
R2: <cite>$1</cite>
R3: <i lang="fr">$1</i>
R4: <i lang="la">$1</i>
```
The Search sometimes will find a phrase such as "_en route_" (French), and the appropriate Replace in this case would be R3. The phrase likely occurs frequently, so the user today can either replace each one as it occurs, or suspend the current S&R and create a new one just for `<i>en route</i>`.  A "Replace all identicals" would find and replace all occurrences of `<i>en route</i>` with `<i lang="fr">en route</i>` and return the cursor to the position it held just before the "Replace all identicals" was executed. That would allow the user to continue the normal S&R, without having to individually Replace each "_en route_".

windymilla:
When using the Replace All buttons in the S/R dialog, if the button is Shift-clicked, it only replaces the strings that exactly match the current match. It then moves on to the next match.

"Exactly" means not regex (regardless of dialog setting) because it's an exact match; case sensitive (regardless of dialog setting) - could be argued it should match dialog; whole word matches dialog setting (or chaos would ensue)

Fixes #609

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -255,6 +255,14 @@ class SearchDialog(ToplevelDialog):
             repl_all_btn.grid(
                 row=rep_num + 3, column=4, padx=PADX, pady=PADY, sticky="NSEW"
             )
+            mouse_bind(
+                repl_all_btn,
+                "Shift+1",
+                lambda _e, idx=rep_num: self.replaceall_clicked(  # type: ignore[misc]
+                    idx,
+                    identicals_only=True,
+                ),
+            )
             self.repl_all_btn.append(repl_all_btn)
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Very nice!

PR #1321, opened 2025-07-31
Fix silent crash using Recent Files menu

### LINKED ISSUE #1319: Guiguts quits after loading a recent file
1.  Opened Guiguts
2.  Pressed Alt+F
3.  Pressed U
4.  Pressed 1 (or another number)
5.  Guiguts closes unexpectedly.

This seems to happen intermittently on other files, but it just occurred three times in a row on this one. 

Opening a file using Alt+F then O, or with Ctrl+O, always works without issue.

[palaces.zip](https://github.com/user-attachments/files/21511925/palaces.zip)

windymilla:
When the file name changes, previously the whole menu system was cleared and re-written, just because the Recent Documents menu needed rewriting. Either the repeated clearing and recreating (internal structures or Tk menus) or occasional timing clashes of using the menu and recreating it at the same time caused GG to silently exit.

Fixed by rewriting the labels and commands in the Recent Docs menu and not deleting/recreating everything.

Fixes #1319

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -288,7 +288,7 @@ class Guiguts:
         """Handle side effects needed when filename changes."""
         # Recreate menu to reflect recent files
         self.populate_recent_menu()
-        self.recreate_menus()
+        self.update_recent_menu()
         self.update_title()
         CustomMenuDialog.store_filename(self.file.filename)
         maintext().after_idle(maintext().focus_set)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing Notes:
The Recent Files menu is now updated via completely new code. Please check things like each entry opens the correct file, especially the first and last entries. Also that the menu is updated correctly - normally that will be the newly-loaded file appearing at the top of the menu, and other entries moving possibly moving down one place. Also, if you've renamed a file, but its name is still in the list - it will fail to load, but nothing bad should happen. Thanks

srjfoo (approved):

PR #1322, opened 2025-07-31
Fix exception caused by searching for bad regex

### LINKED ISSUE #1318: Crash during regex search
I received multiple Tkinter exceptsions while trying to perform the following "Find All" regex search: `(</i>)([!?;:])`

I received the same exception (below) about six times, and the application quit immediately after I closed the last exception dialog.

Guiguts version: 2.0.0b1

Python version: 3.11.13 (main, Jun  3 2025, 18:38:25) [Clang 16.0.0 (clang-1600.0.26.6)]
Tk/Tcl version: 8.6.16
OS Platform: macOS-26.0-x86_64-i386-64bit
OS Release: 25.0.0

Image

windymilla:
If the user ignored the red regex warning and searches anyway, the search match highlighting code raised an exception.

Treat like the regex timeout situation - output a single error and turn off match highlights until the user does another search.

Fixes #1318

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2615,6 +2615,13 @@ class MainText(tk.Text):
             )
             self.highlight_search_deactivate()
             return None, 0
+        except re.error as e:
+            logger.error(
+                f"Regex error: {str(e)}\n\n"
+                "Search highlighting turned off temporarily."
+            )
+            self.highlight_search_deactivate()
+            return None, 0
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
1. Enter a bad regex, e.g. `[`
2. Regex should turn red
3. Click Search anyway
4. May need to click in main text window to focus there before the exception would happen in master.

PR #1324, opened 2025-07-31
Provide message box in status bar

windymilla:
1. Show last message in box
2. Fade message after 5 seconds
3. If message box doesn't fit on same line as buttons, drop it (and the "Working" label) to a second line.
4. Take into account whether ordinal button is displaying name or not, i.e. allow more space for it if it is.
5. When the user resizes the main window or changes the ordinal names preference, recalculate whether 1 or 2 lines are required in the status bar.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -188,7 +188,6 @@ class Guiguts:
         root().set_zoom_fullscreen()
 
         self.logging_add_gui()
-        logger.info("GUI initialized")
 
         preferences.run_callbacks()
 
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes:

So far, you can generate messages in the new message box via the following:
1. Generate a message with the Search dialog, like Replace All reports how many replacements, or Count counts matches.
2. Try to import prep text files from a dir with no txt files
3. Change your language to a one that you don't have a DEFAULT dictionary for, then try to spell check, and you should get the warning "No default dictionary for language klingon". You can have a personal dictionary for it. If you have neither dictionary, you will get the warning plus a popup error saying no dictionary for klingon.

Either on this PR or a new one, we can add messages from checker dialogs, like how many Fixes were made.

srjfoo (approved):

PR #1334, opened 2025-08-04
Fix click-on-selection bug

### LINKED ISSUE #1333: Can't click within a selection to place cursor
Select some text. Then position the mouse somewhere in the selection and click, then release without moving.

**Expected behavior:** the current selection is de-selected and the cursor is placed at the position of the  mouse click.

**Observed behavior:** the "drag" cursor appears momentarily and then disappears. The selection remains as it was and the cursor is not moved.

To accomplish the desired result, first the selection has to be de-selected by clicking outside the selected area, or using an arrow key to move the cursor, and then the click is needed to place the cursor at the new position.

windymilla:
If user clicks on selection, don't go into full dragging mode unless they actually move the mouse.
1. Don't change cursor to drag cursor unless mouse has moved
2. When mouse is released over the selected text, treat it like a click, i.e. position insert cursor and clear selection.

Fixes #1333

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2087,7 +2087,7 @@ class MainText(tk.Text):
         sel_start = sel_ranges[0].start.index()
         sel_end = sel_ranges[0].end.index()
 
-        if self.compare(index, "<", sel_start) or self.compare(index, ">", sel_end):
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

tangledhelix (approved):
Looks good!

PR #1356, opened 2025-08-13
Add vowels with macron, breve

### LINKED ISSUE #1095: Commonly Used Characters request
I would like to add vowels with macron, and with breve.

Or add the ability for user's to enter characters, like GG1. But I think these make sense anyway.

srjfoo:
Expanded Commonly Used Characters to include vowels with macrons and vowels with breves. Also added the single left and right angle quotes and the minus sign. (Fixes #1095)


windymilla (approved):
Thanks!

PR #1383, opened 2025-08-27
Don't show busy cursor for palette commands; leave

tangledhelix:
As discussed in Slack.These cursor changes add considerably to the time it takes to run what would otherwise be a sub-second tool run. Without them, executions via the open panel should be much faster than before.

### DIFF
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -41,7 +41,6 @@ from guiguts.widgets import (
     mouse_bind,
     Combobox,
     Notebook,
-    Busy,
     TreeviewList,
     ScrollableFrame,
     themed_style,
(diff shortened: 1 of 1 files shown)


windymilla (commented):
All fine, apart from `Busy` being imported but not used, as reported by flake8
Thanks

windymilla:
Just noting, for future devs - it only appears to be slow on Macs, and is related to changing the cursor in the tree of widgets.

tangledhelix:
> Remind me, please -- this is only if the commands are triggered by running directly from the command palette, not if using keyboard shortcuts, or does it affect keyboard shortcuts, also?

In my testing it was only when using the Palette, not shortcuts. I believe that's because under the hood on macOS, the keyboard bindings are bound to the menu system and have nothing to do with the palette code. I have no idea whether that's different on the other platforms.

Either way, it was only slow when executing using the palette itself, for me.

tangledhelix:
> Just noting, for future devs - it only appears to be slow on Macs, and is related to changing the cursor in the tree of widgets.

Do you think we should use an `is_mac()` guard here, or is that not worth doing?

windymilla:
> Do you think we should use an `is_mac()` guard here, or is that not worth doing?

No, let's remove for all platforms as you have done.

windymilla (approved):

srjfoo (approved):

PR #1389, opened 2025-08-27
Fix word boundary regex bug

### LINKED ISSUE #1369: Regex bug with word boundary?
Search: `\b.+?\b`
Replace: `X`

Matches a word, but instead of replacing with a single "X", it replaces with n "X" characters where n is the length of the word.

windymilla:
The combination of word boundary with a non-greedy match caused the replacement to be incorrect.

Example:
S: `\b.+?\b`
R: `X`

This should find words, i.e. groups of characters delimited by word boundaries, and replace it with a single `X`. In fact it replaced every character in the word with `X`.

This is because we can't use Tcl regex replacement since it is very limited, so we have to get the match into a python string, e.g. `word`. We now execute the replacement on that string. This means it does not have context, so word boundaries and look-ahead/behind at the start and end of the string don't work. We therefore removed them from the search string used on the Python string.

However, this means that the search term is now `.+?`, which matches just the `w` in `word`, which it replaces with `X`, then it matches the `o`, then the `r` then the `d`.

The fix is to replace a beginning of line `\b` with `^` and an end of line `\b` with `$`, and the same for lookahead/behind.

Fixes #1369

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -1074,14 +1074,15 @@ def get_regex_replacement(
 
     # Since below we do a sub on the match text, rather than the whole text, we need
     # to handle start/end word boundaries and look-behind/ahead by removing them.
-    # At some point the sub will be done manually, handing groups, execution of
-    # python code, etc., like in GG1. At that point, these fixes can probably go.
-    search_regex = re.sub(r"^\(\?<=.*?\)", "", search_regex)
-    search_regex = re.sub(r"\(\?=.*?\)$", "", search_regex)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
This could break any regex replacement (searching is unchanged) where the search term has `\b`, `(?=...)`, `(?!=...)`, `(?<=...)`, `(?<!...)`,  particularly at the beginning or end of the term (lookahead only at end, lookbehind only at beginning). 
I hope the fix will only kick in when the search term has a non-greedy match, e.g. `.+?` or `.*?` where it matches as few characters as it can.

PR #1391, opened 2025-08-27
Couple ebookmaker-related fixes

### LINKED ISSUE #1377: Local Ebookmaker directory error pushes dialog to the back
If there is no local ebookmaker directory set up, or the directory doesn't exist, the error dialog causes the EBM dialog to be pushed to the back. Then the user can't easily find the Browse button to search for the correct directory.

Either bring the EBM dialog to the front when there's a directory error and/or automatically pop the directory selection dialog.

### LINKED ISSUE #1379: Re-run ebookmaker doesn't show red "Working" label
Reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1373796#p1373796)

> When you run the online ebookmaker, you get the red "Working" message in the message space in the main window.
> If you click "Re-run" in the dialog box, the ebookmaker does start processing, but you don't get the red "Working" message for any subsequent re-run.

windymilla:
1. Make checker Re-run button show red "Working" label in bottom right corner, e.g. Ebookmaker API checker
Fixes #1379 
2. Raise ebookmaker dialog after error
Error dialog was pushing ebookmaker dialog to the back, so the user couldn't see it to Browse for the correct local directory.
Fixes #1377

### DIFF
--- a/src/guiguts/html_tools.py
+++ b/src/guiguts/html_tools.py
@@ -1251,9 +1251,16 @@ class EbookmakerChecker:
             logger.error(
                 f"Use 'Browse' button to select Ebookmaker 'pipenv install' {folder_dir_str(lowercase=True)}"
             )
+            Busy.unbusy()
+            self.dialog.lift()
             return
         if not os.path.isdir(cwd):
-            logger.error(f"{folder_dir_str()} {cwd} does not exist")
+            logger.error(
+                f"{folder_dir_str()} {cwd} does not exist\n"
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes for second commit: Run Ebookmaker (local installation) - you don't need ebookmaker installed to see the problem. Make sure the directory entered in the field next to the Browse button either doesn't exist or doesn't have ebookmaker installed in it. The error dialog will pop. When the dialog is dismissed, the ebookmaker dialog should now pop to the front, which it does not do in `master`.

PR #1394, opened 2025-08-28
Fix Page Up/Down to be symmetrical

### LINKED ISSUE #1373: Page Down then Page Up doesn't return to same line
Described [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1373500#p1373500)

windymilla:
The default Tk Page Up/Down behavior goes up one more line than it goes down.

Manually implement page up/down to move the same
number of lines, as well as Shift+Page Up/Down to
select/extend selection by a page symmetrically.

Fixes #1373

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1084,6 +1084,57 @@ class MainText(tk.Text):
                 bind_peer=True,
             )
 
+        # Override page up/down behavior since it is not symmetrical,
+        # so PgUp then PgDn doesn't bring you back to where you were.
+        def get_page_lines(text: tk.Text) -> int:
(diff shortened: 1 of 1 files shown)


windymilla:
Notes: if this has problems, it's probably best to back it out completely. It's taken quite a bit of work to get it to work, but it has added some complication to attempt to mimic the default behavior (but correctly). The original fault is not significant enough to warrant introducing bugs if we get suspicious of it needing to be more complex than this.

srjfoo (approved):
It works as advertised for me on macOS. I couldn't really think of an edgecases to poke at, but if you're in the least uncomfortable with the complexity of the code, the off-by-one of the current behavior doesn't bother me that much.

PR #1395, opened 2025-08-29
Permit NumLock & ScrollLock when typing in WF

### LINKED ISSUE #1371: Type-to-search not working on Linux
Even single letter typing in Word Frequency has stopped working on Linux.

windymilla:
Also in checker dialogs.

Fixes #1371


windymilla:
Testing notes: Please just check that type-to-search still works in WF and an example checker dialog, like bookloupe. It should work even if you put Caps Lock on, and Num Lock if you have it. But if you hold Cmd down when you type, e.g. Cmd+C, it should not jump to the first word beginning with C.

srjfoo (approved):

PR #1401, opened 2025-09-02
Attempt to speed up "Busy" on Macs

windymilla:
Remove calls to `update()` from the recursive calls to set the cursor shape. So instead of 100-200 calls to update, there should now be 2.

This still works acceptably on Windows & Linux (WSL), but needs testing on Macs

Previous suggestion, to not change the cursor recursively on the widgets, stop the cursor being changed on Windows, at least.


srjfoo (approved):
Saving is now instantaneous. I'm not entirely sure about Bookloupe and WF -- WF seemed to be slightly faster, but they're both lengthy checks on a long book, anyway.

PR #1403, opened 2025-09-03
Fix bug with illo wrongly marked as mid-para

windymilla:
Bad logic in checker fixed.


srjfoo (approved):
In addition, it corrrectly identifies several mid-para illustrations that weren't being identified before as mid-para.

PR #1411, opened 2025-09-08
Fix click-drag bug with NumLock

### LINKED ISSUE #1410: Left-clicking at start or end of highlighted text inserts clipboard contents
1. Load attached file.
2. Copy `no country.` to clipboard.
3. Open Search & Replace.
4. Paste from clipboard.
5. Search.
6. Text should be highlighted.
7. Click at the start or end of the highlighted text.
8. A copy of the search text is inserted.

This is weird behavior I have been seeing for a while, finally got a reproducible sequence. I don't know if all the steps are needed, or if this is the only way. Should be the latest code.

Guiguts version: 2.0.2

Python version: 3.13.7 (tags/v3.13.7:bcee1c3, Aug 14 2025, 14:15:11) [MSC v.1944 64 bit (AMD64)]
Tk/Tcl version: 8.6.15
OS Platform: Windows-11-10.0.26100-SP0
OS Release: 11

[18000.zip](https://github.com/user-attachments/files/22198666/18000.zip)

windymilla:
NumLock was being interpreted as "a modifier key", and thus causing "copy" behavior.

Fixes #1410

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -2401,8 +2401,7 @@ class MainText(tk.Text):
     def _is_drag_copy_mode(self, event: tk.Event) -> bool:
         """Return whether select drag should be copying."""
         # Shift, Ctrl, Cmd
-        print(f"{int(event.state):x}", flush=True)
-        return bool(int(event.state) & (0x0001 | 0x0004 | 0x0010))
+        return bool(int(event.state) & (0x0001 | 0x0004 | 0x0008 | 0x0010))
 
     def _get_truncated_drag_preview(self, text: str) -> str:
         """Get a truncated version of the drag text."""


windymilla:
> Drag works as expected now, clicking at start/end does not copy the text. Clicking within highlighted text deselects it, except when just barely within (< 1/2 char) start or end.

Thanks for the detail about the inconsistency, @rtonsing - it's the kind of thing that could annoy people. My latest commit makes that a bit more self-consistent, so the "1/2 char" thing only applies to copying (to allow you to copy "xyz" to make "xyzxyz"), not moving or clicking. You'll also notice that unlike previously, if you have a fresh file and select some text, then clicking in the first/last 1/2 char no longer marks the file as "edited".

windymilla:
I've added a debug print which should print a hex number multiple times as you drag text. 
Please could you select some text, then drag it, and note the number.
Then try Cmd-dragging and note the number.
Then try Option-dragging and note the number.

The dragging behavior won't be fixed - this is just for debugging. Once I get the numbers, I'll make the fix.
Thanks

And on the subject of using Option as a modifier key, I don't think tkinter will detect it, which is why we went for Cmd. I'll know for sure when you tell me the numbers.

rtonsing:
The issue is still occurring, I can highlight a word, click near the start or end of it, and a copy of the word gets pasted. Only with NumLock on.
It was working OK when I reviewed it, but that was before the macOS changes.

PR #1413, opened 2025-09-08
Use Ctrl key for column select on Linux

### LINKED ISSUE #1365: Add alternative key for column-select dragging on Linux?
On (at least some varieties of) Linux/Window Manager, Alt-click&drag is used by the window manager, so column-selection doesn't work, and user needs to do a regular selection, then convert it to a column selection.

Is there an alternative that could be used?

windymilla:
Since Alt-dragging is used by some Linux window managers, use Ctrl-dragging instead. No change to Mac/Win behavior.

Fixes #1365


windymilla:
Tested OK on WSL - Ctrl+drag does column select, whereas in `master` Alt+drag did it.

PR #1416, opened 2025-09-08
Report unconverted italics in PPhtml

### LINKED ISSUE #1375: PPhtml request: check for unconverted italics markup
Your special checks have done a nice job of catching some unusual things.

It would be nice to catch `_italics_` markup, maybe even if only at the start or end of a word. That character is pretty unusual in normal text, and then usually inside words.

windymilla:
Add to the miscellaneous special checks near the end. Should catch `_this_`, `_multi word on one line phrase_`, `_broken`, `missing_`, `P_art_`, but not `mid_word`. It's detecting when there's an underscore at the start and/or end of the word, but permits mid-word underscores.

Fixes #1375


srjfoo (approved):
If there's a test for unconverted italics, should there be a corresponding test for unconverted bold?

PR #1421, opened 2025-09-10
Make Ctrl-A select all in Entry fields

### LINKED ISSUE #1420: Make Cmd/Ctrl+A in Entry fields do "select all" not "go home"

windymilla:
Example of Entry field is in ASCII Table Effects dialog. Also applies to comboboxes, like S/R Search/Replace fields.

Only affects Linux, since Windows/macOS support
Select All natively via Cmd/Ctrl-A

Fixes #1420

@srjfoo - I've tested that the change happens and works on WSL, and that Windows behavior is unchanged. Please can you just check that Cmd-A still selects all text in Entry and Combobox fields on macOS. Thanks.

### DIFF
--- a/src/guiguts/root.py
+++ b/src/guiguts/root.py
@@ -50,6 +50,19 @@ class Root(tk.Tk):
         self.save_config = False
         self.bind("<Configure>", self._handle_config)
 
+        # Make Ctrl-A select all in Entries/Comboboxes on Linux only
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
cmd-a still works on macOS.

I don't know if it's relevant or not, but crtl-a and ctrl-e are still beginning of line and end of line on macOS.

PR #1452, opened 2025-09-21
Further CP improvements

### LINKED ISSUE #1449: Things to consider for CP features
Reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1376323#p1376323)
 
- It would be nice if, after selecting the checkbox for WF char suites, it opened the WF dialogue box automatically. It may not be otherwise obvious to people that it isn't going to highlight them in the main UI nor where to go. I added some content and images to the manual to help with this in the interim.
- From an order standpoint, if the goal is to have the menu items roughly in the suggested order, I think I would insert the Manage Character Suites before Dehyphenization and then put WF char suites after Char Suites. I think it would encourage people to pick the suites they know they will need up front (e.g. Greek). I kind of go back and forth on whether or not to do this before or after Filter Files, because if you do it before, it will show you the curly quotes as well, which obviously you don't need to deal with in this dialogue, but by doing it at this point, you allow for dehyphenization and filter files to clean up anything that might otherwise be messed up by having a random splotch between the quotes or the hyphens, so I think that's better on average as far as cleaning up the maximum amount.
- Would it be possible to expand the dehyphenization processes to also cover wrapping the emdashes to the next word within the same paragraph? This could be a really nice enhancement for proofers for books with a lot of EOL hyphens. Maybe not possible, but just occurred to me as I was looking through [this book](https://www.pgdp.net/c/tools/page_browser.php?project=projectID68cc6fe54d574&imagefile=011.png&mode=imageText&round_id=OCR) I just prepped.
- Would it be possible to make Keep and Remove in the dehyphenization different colors? It's not hard to separate them when you sort by type at the start, but as you go, it can become a little hard to distinguish them once you start swapping individual examples back and forth. I think red for remove and green for keep, for example, would help reduce mistakes. Again, really small actual impact, just a nice to have.

windymilla:
1. Change order of Char Suite options
2. Clarify wording on WF Char Count toggle for highlighting chars not in selected char suites
3. Use green for Keep prefix, red for Remove prefix
4. Add extra error prefix color to Settings dialog so it can be configured

Fixes #1449

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -783,11 +783,23 @@ class Guiguts:
         cp_menu.add_button(
             "Remove ~Headers/Footers...", lambda: HeadFootChecker().run()
         )
+        cp_menu.add_separator()
+        cp_menu.add_button(
+            "Manage Character ~Suites...", CPCharSuitesDialog.show_dialog
+        )
+        cp_menu.add_checkbutton(
+            "Highlight ~WF Char Counts if not in Selected Suites",
+            PrefKey.CP_HIGHLIGHT_CHARSUITE_ORPHANS,
+            WordFrequencyDialog.highlight_charsuite_refresh(),
+            WordFrequencyDialog.highlight_charsuite_refresh(),
+        )
+        cp_menu.add_separator()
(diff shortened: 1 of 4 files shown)


srjfoo (approved):
LGTM!

PR #1464, opened 2025-09-27
application.py - fix the creation of the preferences

tjeffress:
fix the creation of the preferences to use xdg-open if not on windows or mac

On Linux distributions that don't alias open to xdg-open, the open "$f" command won't work.  For the Custom menu, the default was still open "$f".  This change adds a check for mac and then defaults to xdg-open "$f" if not on windows or mac os.

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -603,7 +603,7 @@ class Guiguts:
         preferences.set_default(
             PrefKey.CUSTOM_MENU_ENTRIES,
             [
-                ["View HTML in browser", 'start "$f"' if is_windows() else 'open "$f"'],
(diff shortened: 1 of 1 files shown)


windymilla (changes requested):
Thanks @tjeffress 

The edit looks correct, and it works in my WSL setup.

The only thing that needs doing before it can be merged is for it to pass the tests in the Github actions. They are documented in the "Code style" section of the README (also see the Documentation, Type Checking and Testing sections below). The test that is failing is 
`poetry run black .` (including the period)
This just wraps long lines, ensures consistent spacing etc. 
After running that in your branch, just commit and push it, then I will combine the two commits when I merge it into `master`.

Thanks again

tjeffress:
I have run 'poetry run black .' on the project and re-pushed the commit to the branch.

PR #1489, opened 2025-10-05
Handle missing scanno/regex library file better

### LINKED ISSUE #1483: Broken/non-existent regex json files
As reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1377582#p1377582) if a regex library json is corrupt (or absent - not sure which) and the regex dialog is opened, which attempts to load the currently selected file from the dropdown, there's a problem that can't easily be fixed.

Consider: instead of (or as well as) the error, it could/should go back to loading the first of the default supplied files.

windymilla:
If the current regex/scanno file (as stored in the prefs file) is deleted by the user, next time they try to open the dialog, an error pops up, the dialog ends up behind the main window and the cursor looks busy, so user thinks they can't use the dialog.

Instead, make sure dialog gets popped to front once error is dismissed and no longer looks busy.

Fixes #1483

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -1757,6 +1757,8 @@ class ScannoRegexCheckerDialog(CheckerDialog):
         scanno_dict = load_dict_from_json(path)
         if scanno_dict is None:
             logger.error(f"Unable to load {self.type_adj} file {path}")
+            self.display_entries()
+            self.lift()
             return
         # Some scannos files require whole-word searching
         try:


windymilla:
Testing notes: 
1. Copy/create a regex library json file to user area
2. Open regex library dialog
3. "Load File" to load that file into library
4. Close Regex library dialog
5. Delete or rename the json file (or edit it so it is invalid json)
6. Open regex library dialog
7. Error box pops up (2 boxes in the case of invalid json) - dismiss errors
8. In `master` regex library dialog ends up behind main window, and program may appear "busy"
9. With this PR, dialog is popped to the front, and not "busy". User can now either rename/fix the missing/broken file, or open another one json using the dropdown, or Load File
10. Note that once a json file is deleted, **_and_** another file is selected using the dropdown, the deleted file will no longer appear in the dropdown - this is intended behavior to keep the dropdown tidy without providing a "Manage Dropdown" dialog.

PR #1514, opened 2025-10-23
Cope with `<div>` round Auto-illos

### LINKED ISSUE #1512: Auto illustration doesn't recognize `[Illustration]` inside a `div`
Reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1378768#p1378768).

windymilla:
When GG converts `[Illustration]` to HTML, it adds `<p>` markup around it. Auto-illus removes this when it converts the illustration to an HTML `<figure>`.

New behavior: If the user customizes that `<p>` markup with a class, or uses a `<div>` instead, leave their markup alone and just put the `<figure>` inside it.

Fixes #1512

### DIFF
--- a/src/guiguts/html_tools.py
+++ b/src/guiguts/html_tools.py
@@ -381,7 +381,10 @@ class HTMLImageDialog(ToplevelDialog):
         # Find and go to start of next unconverted illo markup
         illo_match_start = maintext().find_match(
             r"\[Illustration",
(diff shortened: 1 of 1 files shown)


windymilla:
@tjeffress - does this fix the issue you raised in the forums? (And not break other examples of `[Illustration]`)
Thanks

PR #1527, opened 2025-10-31
Add `--resetgeometry` command line option

### LINKED ISSUE #206: Provide better support for intermittent dual monitor users
Using more than one monitor, and sometimes using 2 and sometimes 1 monitor, can cause geometry-related issues:
1. If you drag the main window to a position with negative X or Y (i.e. left/above your main screen), then exit, then remove the second monitor, GG may start up off-screen and can't be dragged back onto the main screen. The GGprefs file will have `rootgeometry` set to something like `400x300+-200+500`, where the `-200` means negative X position. Should probably either always force the X & Y to be non-negative, or possibly detect if the second monitor is currently available, and if not, then force X & Y to be non-negative.
2. (It's possible that tooltip positioning may be weird with multiple monitors (not confirmed yet), e.g. if GG is dragged to a monitor with a different resolution to the main monitor, then the wrong screen width/height may be used, causing tooltips to be misplaced.) ETA - new tooltip positioning algorithm doesn't use screen width/height, so not a problem.
3. Investigate dragging other dialogs off onto second monitor and then removing the monitor. It could be annoying if all dialogs are forced back to non-negative when the user always has a second monitor. Detecting all monitors and only moving things back if it the dialog wouldn't appear on any of the monitors would be less annoying. May be necessary to have some kind of "reset geometry" setting or command-line arg (no good for user though). Maybe always force main window to start up on main monitor, then have a "reset geometry" option for the other dialogs?

windymilla:
No PPers have reported a problem with multiple monitors, but if the main window or dialogs are dragged onto a second monitor, then the monitor is removed before the next run of GG, the windows may be placed off-screen.

Use `guiguts --resetgeometry` to reset the size and position of the main window and all dialogs to the default values, which will bring them back onto the main screen.

I think this is sufficient for now for a situation that hasn't arisen during PPing.

Fixes #206

Testing notes: 

- Add second/third monitors, and drag various windows (including main window?) onto them. 
- Quit GG
- Turn one or more additional monitors off
- Start GG
- Notice dialogs are invisible
- Quit GG
- Start GG with the `--resetgeometry` argument
- Notice dialogs are now visible

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -181,6 +181,11 @@ class Guiguts:
         logger.info("Guiguts started")
 
         self.initialize_preferences()
+        if self.args.resetgeometry:
+            preferences.reset(PrefKey.ROOT_GEOMETRY)
+            preferences.reset(PrefKey.ROOT_GEOMETRY_FULL_SCREEN)
+            preferences.reset(PrefKey.ROOT_GEOMETRY_STATE)
+            preferences.reset(PrefKey.DIALOG_GEOMETRY)
 
         self.file = File(self.filename_changed, self.languages_changed)
         the_file(self.file)
@@ -287,6 +292,11 @@ class Guiguts:
             "--version",
             action="store_true",
             help="Display the version of Guiguts",
+        )
+        parser.add_argument(
+            "--resetgeometry",
+            action="store_true",
+            help="Reset layout of dialogs and main window to defaults",
         )
         self.args = parser.parse_args(args)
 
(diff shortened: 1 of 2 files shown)


windymilla:
> In my testing, situation 1 may still have problems -- if you accidentally press return before quitting to run `--resetgeometry`, you may unintentionally "edit" the file, thus causing the quit operation to hang waiting for a response. This happened to me. (And when I tried to force quit Python, the OS-level dialog came up on the second monitor 🤣 )
> ...
> I think that having this is better than not, but wanted to warn about the potential problem with not being able to quit the hidden window.

Yes - a small risk, and not too serious, IMO. This is a rare situation, and only likely to be encountered by users who are fairly competent technically, given that they have multiple monitors.
Thanks for investigating the different possiblilities.

Issue #1518, opened 2025-10-27
Check out python 3.14


windymilla:
Does GG2 run OK under 3.14?

Are there any docs that need updating?

PR #1531, opened 2025-11-01
Update versions of dev tools

windymilla:
Testing notes:
Potential areas that could be broken include: dragging a selection or column selection in main/peer window, selecting a character in the Unicode Block dialog, pasting a string into the HTML Generate dialog title box


windymilla:
I'm going to say that this concludes the investigation for Python 3.14 for now. It looks like the next batch of work for new versions will revolve around Python 3.15 and Tk 9.0.

Fixes #1518

PR #1549, opened 2025-11-10
Remove extra spaces after <cite> replacement

tangledhelix:
The `<cite>` replacement in italic_semantics.json had extra spaces after the replacement, which could result in incorrectly rendered space between a `<cite>` and punctuation, and possibly other issues.


windymilla (approved):
Oops - thank you!

PR #1558, opened 2025-11-13
Avoid exception when GG2 exits

windymilla:


srjfoo (approved):
Unable to duplicate the problem, but the code looks okay.

PR #1584, opened 2025-11-27
Add suggestions to Spelling dialog

### LINKED ISSUE #1579: Consider whether a spelling suggestion feature is feasible
When there is a word flagged by Spell check, see if there are any suggestions in the global & project dictionaries, and allow the user to choose one.

Maybe use `difflib.get_close_matches` to get some possible matches, and use Levenshtein `distance` to rank them

windymilla:
1. Add a dropdown of the 10 best suggestions
2. Add buttons to Fix the spelling, or to Fix all occurrences.
3. Also add buttons to Fix (All) and Hide the spelling from the list.

Fixes #1579

### DIFF
--- a/src/guiguts/spell.py
+++ b/src/guiguts/spell.py
@@ -121,8 +121,8 @@ class SpellCheckerDialog(CheckerDialog):
         )
         frame = ttk.Frame(self.custom_frame)
         self.custom_frame.columnconfigure(0, weight=1)
-        frame.grid(column=0, row=1, sticky="NSEW", pady=5)
-        frame.columnconfigure(0, weight=5)
+        frame.grid(column=0, row=1, sticky="NSW", pady=5)
+        frame.columnconfigure(0, weight=10)
         frame.columnconfigure(1, weight=0)
         frame.columnconfigure(2, weight=5)
         frame.columnconfigure(3, weight=3)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
1. Do the suggestions look reasonable? Words should be suggested from both the global dictionary and the project dictionary as appropriate. These dictionaries should be dynamically updated. For example, if you add "xyzzy" to the dictionary, it may suggest that for the next word if it is "xyzzx", without you needing to re-run or close and re-open the dialog.
2. Is the "case" behavior reasonable? It should normally match case-pattern of lowercase, uppercase, capitalized, i.e. suggest "dog" for "dof", "DOG" for "DOF", and "Dog" for "Dof", but "London" for "london" (because "London" appears in the dictionary as a proper noun)
3. Do the 4 fixing buttons work correctly?
4. Does adding to the project dict and/or global dict still work (either by button of clicking the word in the list with a suitable modifier key)? I've hijacked the "Fix" mechanism for both the project/global dict addition, and the suggestions fixing, so there's potentially a clash there.
5. Is the performance acceptable, even if you have say the English and French dictionaries loaded. Every time you select a word in the list, it combines the global dictionaries & the project dictionaries and does some processing of them to handle the upper/lowercase thing, then it searches the dictionaries for good matches, and sorts them, displaying the top 10. If the performance is too slow, we could make it an optional feature, turned on via a Pref, but I'd rather not if that's not necessary.

PR #1627, opened 2025-12-17
Improve behavior of Goto Line/Page dialog

windymilla:
Linux is more sensitive than Windows, so previous version caused exception


tjeffress:
I have tested the change and the Goto Page and Set Number dialogs are now working as they did previously. (Well except they don't accept octal or hex, but that's as expected.)

PR #1651, opened 2026-01-01
Avoid crash in Command Palette

### LINKED ISSUE #1646: Traceback from Command Palette
Pop dialog, then press Enter (sometimes traceback, sometimes runs last command, sometimes changes sort column)

```16:20:12: ERROR - Tkinter Exception
  File "C:\Users\xxx\.pyenv\pyenv-win\versions\3.14.0rc3\Lib\tkinter\__init__.py", line 2082, in __call__
    return self.func(*args)
           ~~~~~~~~~^^^^^^^
  File "C:\DP\guiguts-py\src\guiguts\misc_dialogs.py", line 2181, in execute_command
    if event is not None and self.list.process_click(event) == "":
                             ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "C:\DP\guiguts-py\src\guiguts\widgets.py", line 838, in process_click
    col = int(col_id[1])
              ~~~~~~^^^
IndexError: string index out of range
```

windymilla:
Certain combinations of popping the dialog, then pressing Return/Enter could cause a traceback. The code wasn't handling the Return/Enter correctly, and treating it like a mouse click.

Fixes #1646

Testing notes:
Please check running commands from within the Command palette still works, either by double clicking in the list, or single-clicking in the list then pressing Enter/Return, or by single-clicking in the list, then clicking the Run button, or focusing in the entry field at the top, then pressing Enter/Return. Pressing Enter/Return should behave the same as clicking the Run button, i.e. run the currently selected command. Whatever happens, it shouldn't give a traceback. You may no be able to get `master` to traceback - it seems timing/sequence related, and I could get it to happen only occasionally. I've fixed the bit of code where the traceback happened anyway.

### DIFF
--- a/src/guiguts/misc_dialogs.py
+++ b/src/guiguts/misc_dialogs.py
@@ -1988,7 +1988,7 @@ class CommandPaletteDialog(ToplevelDialog):
         self.list.tag_configure(self.SEPARATOR_TAG, foreground="gray")
 
         # Bind events for list and entry
-        self.list.bind("<Return>", self.execute_command)
+        self.list.bind("<Return>", lambda _: self.execute_command())
         self.list.bind("<Double-Button-1>", self.execute_command)
         self.list.bind("<Down>", lambda _: self.move_in_list(1))
         self.list.bind("<Up>", lambda _: self.move_in_list(-1))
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
I never did trigger the exception in `master`, but one difference I did note was that

> or focusing in the entry field at the top, then pressing Enter/Return

did not work in `master`, but did in the `command-palette-exception` branch.

PR #1662, opened 2026-01-06
Handle Ngrams properly when word contains -*

tangledhelix:
Allows user to select a `hyphenated-*suspect` and have Ngram handle it properly without having to remove the `*` first.

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -2796,8 +2796,10 @@ def convert_sequential() -> None:
 
 def open_ngram() -> None:
     """Open Google Books Ngram viewer, using selected text if any.
-    If a single hyphenated word, compare with unhyphenated."""
+    If a single hyphenated word, compare with unhyphenated. Ignores
+    an asterisk following a hyphen and treats as just the hyphen."""
     content = re.sub(r"\s+", " ", maintext().selected_text().strip())
+    content = content.replace("-*", "-")
     # If single hyphenated word, add non-hyphenated version
     if "-" in content and "," not in content and " " not in content:
         content = f"{content},{content.replace('-', '')}"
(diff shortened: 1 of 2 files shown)


windymilla (changes requested):
Although this handles the specific case of `hyphenated-*suspect` correctly, the GG Ngrams feature also supports such things as `regularize,regularise` where both forms to be compared are separated by commas. Thus it is possible to select `to-day,today` to compare their Ngrams. What this PR doesn't yet do is permit `to-*day,today`.
This could probably be done by inserting the single line
`content = content.replace('-*', '-')`
before the `if` on line 2802 in the original, instead of the new code.
Or there may be another way.

tangledhelix:
> This could probably be done by inserting the single line `content = content.replace('-*', '-')` before the `if` on line 2802 in the original

That does seem to do the trick and behave correctly in the different cases I tried. I updated that, rebased, force-pushed.

windymilla:
> Just confirming that this depends on the user actually adding the list of words to the text, and selecting that as a string before bringing up the tool. I'm not arguing with the results, because in testing, it works fine when just selecting text, just wanting clarification on how this is supposed to be used. It kind of surprised me to assume that.

The **_normal_** way of using it is where you have a hyphenated or possibly-hyphenated `*-` word in the text. You just select the word then use the tool. It then compares the hyphenated with the non-hyphenated version in the Ngram viewer. However, the point I made in my review was that the code also accepts where you have two (or more) words separated by commas, e.g. `colorize, colourize, colourise`. The original commit didn't cope if one of those was hyphenated with an asterisk whereas this later commit does. I wouldn't expect a PPer to be "adding the list of words to the text", but it could well happen that the alternative is already there in a proofer comment, e.g. `colorize[**colourize?]`, which a PPer could adjust to make it checkable via the tool. Alternatively, they could select `colorize` and run the tool, then either copy/paste or type `colourize` into the Ngram webpage.

windymilla:
@srjfoo - in case it's helpful, I used this file for some testing, with a combination of different hyphenations
```
bookshop, book shop, book-shop, book-*shop
flashlight, flash light, flash-light
hardhearted, hard hearted, hard-*hearted
to-*day
non-suspect
```

srjfoo (approved):
LGTM!

PR #1685, opened 2026-01-19
Change meaning of All in Hide/Fix All in scannos

windymilla:
Instead of ALL messages, All will now mean all messages that have the same highlighted string.


tangledhelix (approved):
Looks good!

PR #1688, opened 2026-01-20
Ignore Caps Lock, etc., when scrolling image

### LINKED ISSUE #1686: Image viewer scroll doesn't work with Num Lock turned on
At least on Windows.

windymilla:
Mouse wheel scrolling was not working correctly if Caps Lock, Num Lock, etc., were on.

Fixes #1686


rtonsing:
Verified it works. Thanks!

PR #1700, opened 2026-01-27
First draft at ppcomp

windymilla:
At bottom of Tools menu.

Load one of the files, e.g. the HTML file.
In the dialog, choose the HTML file and text file.
Click "Compare" (top right of dialog)

Some of the checkboxes work, such as Ignoring Case, and also suppressing Word Joiners - maybe some others 🤷 

[ppcomp sample.zip](https://github.com/user-attachments/files/24894775/ppcomp.sample.zip)

### DIFF
--- a/src/guiguts/tools/ppcomp.py
+++ b/src/guiguts/tools/ppcomp.py
@@ -323,6 +323,12 @@ class PPcompChecker:
 
     def run(self) -> None:
         """Run PPcomp."""
+        if preferences.get(PrefKey.PPCOMP_EXTRACT_FOOTNOTES) and preferences.get(
+            PrefKey.PPCOMP_SUPPRESS_FOOTNOTES
+        ):
+            logger.error("Cannot use both Extract Footnotes and Suppress Footnote Tags")
+            return
+
         empty_args: list[str] = []
         PPcompChecker.files = []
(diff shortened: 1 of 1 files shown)


rtonsing:
All just general feedback, not complaining, especially the time. I imagine it has to do with no native dwdiff on Windows, I'm impressed that it works at all.

Except for the show line numbers: it would be better to auto-detect which file is in the main window, it is useless to have the numbers for the other file on the left, as far as I can tell. Otherwise, clicking on the line causes the main window to go to a line that is unrelated.

Possibly auto-fill the relevant "Choose file" also. The user can always override it. 

Although there may be scenarios where a 3rd file is active, etc., I imagine it would be most common to want to go to the line in the file being edited. If it is a 3rd file, line selection would not work at all.

My 2 cents.

windymilla:
Latest commit:

1. Hide "Bold" option - not visible at ppwb
2. Hide "Greek title" option - we don't do transcription
3. Hide "Suppress Zero Space" - there is no code to support it. At ppwb, you can check the box but it does nothing
4. Re-organize options to make space for custom CSS.
5. Make Custom CSS be added to the transformations used.
6. Make "Extract Footnotes" work. As at ppwb - line number in HTML (& maybe text sometimes) will be wrong. This is noted in the tooltip over the checkbox.

PR #1730, opened 2026-02-17
Add check for leftover comments

rtonsing:


rtonsing:
I would like to try it. I regret not having participated more all along.

srjfoo:
> I have seen proofers leave only `[**]` with no actual comment before, I think...

Both the above form and the `[**no space after asterisk]` form were caught in my testing.

PR #1738, opened 2026-02-21
Describe ppcomp HTML parse errors in detail

### LINKED ISSUE #1735: PPcomp should give more detail if HTML fails to parse
Example files: [voyages2.zip](https://github.com/user-attachments/files/25451977/voyages2.zip)

windymilla:
1. Report parsing errors in dialog, rather than just a single "Failed to parse" error.
2. Output suitable header depending on whether text or HTML file is currently loaded.
3. Errors are clickable for ease of correction.

Fixes #1735

### DIFF
--- a/src/guiguts/tools/ppcomp.py
+++ b/src/guiguts/tools/ppcomp.py
@@ -332,7 +332,7 @@ class PPcompCheckerDialog(CheckerDialog):
         html_loaded = os.path.splitext(the_file().filename)[1].lower() in (
             ".htm",
             ".html",
-            "xhtml",
+            ".xhtml",
         )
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
Use [example files from linked issue](https://github.com/user-attachments/files/25451977/voyages2.zip) run ppcomp. 
Try with text file loaded and with HTML file loaded.
Make other edits (including more than one in the file) that may fail to parse as HTML, such as use of ampersand, badly spelled HTML keywords (`spanx`),  unclosed quotes within HTML tag, etc.

srjfoo (approved):

PR #1754, opened 2026-03-03
Run Ebookmaker API command in thread

windymilla:
1. When user presses "Run Ebookmaker" button, send the files off to ebookmaker
2. Disable the button, and change its label to "Running..."
3. Display message in dialog to tell user what is happening
4. Pass control back to the user, who can make edits, run other tools, etc.
5. When Ebookmaker is finished and any files downloaded, display the ebookmaker log messages as usual in the dialog and bring it to the front of the screen.
6. Restore the "Run Ebookmaker" button.

### DIFF
--- a/src/guiguts/html_tools.py
+++ b/src/guiguts/html_tools.py
@@ -1494,9 +1494,6 @@ class EbookmakerCheckerAPI:
         )
         self.dialog.display_entries(complete_msg=False)
 
-        # Optional: show busy state
-        # self.dialog.show_busy("Building via Ebookmaker…")
-
         thread = threading.Thread(target=self._run_worker, daemon=True)
         thread.start()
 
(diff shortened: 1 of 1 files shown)


src/guiguts/html_tools.py
```diff
@@ -1473,11 +1474,34 @@ def __init__(self) -> None:
...
+            "Running Ebookmaker in the background…",
+            "",
+            "When Ebookmaker processing completes, this message",
+            "will be replaced with the output.txt build info.",
+            "You can then look at any downloaded files or visit the cache.",
+        )
+        self.dialog.display_entries(complete_msg=False)
+
+        # Optional: show busy state
+        # self.dialog.show_busy("Building via Ebookmaker…")
```
tangledhelix: I realize this is only a comment, but is this even valid code? `.show_busy` doesn't exist elsewhere in the codebase.

If this were activated, it should be in a guard to not run on Mac; recall that we've had those long slowdowns by calling `Busy.busy()` on Mac and we avoid using it for that reason.

PR #1760, opened 2026-03-06
Improve "rnm" scanno check

tangledhelix:
The existing regex for the "rnm" scanno check avoided the word "government," which is common. This change expands on that to avoid "governments" and "governmental" as well.

### DIFF
--- a/src/guiguts/data/scannos/regex.json
+++ b/src/guiguts/data/scannos/regex.json
@@ -201,7 +201,7 @@
 "hint": "Find a word containing pbt and change it to pht"
 },
 {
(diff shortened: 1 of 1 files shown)


srjfoo:
I'm not good enough to quickly debug the change you made to the regex, but if I have a word that legitimally needs to have `rnm` changed to `mm`, it does not work correctly.

`cornments` was changed to `mco`, and `Harnmond` was changed to `mHa`. (governments and governmental successfully ignored, though).

srjfoo (approved):

windymilla (approved):

PR #1818, opened 2026-04-12
Improve dir handling for custom commands

### LINKED ISSUE #1812: Investigate having a dollar-replacement for project directory in custom commands
Reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1393239#p1393239)

Also, it's probably never correct to run the script in the GG installation folder - run it from the project folder instead (if it exists)

windymilla:
1. Run commands from the current project dir
2. Add `$d` token which is replaced by current project dir

Fixes #1812

### DIFF
--- a/src/guiguts/mainwindow.py
+++ b/src/guiguts/mainwindow.py
@@ -35,6 +35,7 @@ from guiguts.utilities import (
     bell_set_callback,
     process_accel,
     sound_bell,
+    folder_dir_str,
 )
 from guiguts.widgets import (
     ToplevelDialog,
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:

1. Add a command that tries to open a file, e.g. runs an editor to open a local file. I used `notepad myfile.txt`. In `master`, myfile.txt is assumed to be in the top level GG folder. In this branch, it is in the currently loaded file's folder.
2. Similar, but use `$d`, e.g. `notepad $d/myfile.txt` - maybe even `notepad $d/../mydir/myfile.txt

srjfoo (approved):

PR #1831, opened 2026-04-22
Fix WF word search failure with `'_`

### LINKED ISSUE #1828: WF find fails with single quote
Word Frequency/Diacritics finds
`Armesünderblum` but clicking on the word doesn't find it in the text where it is `die Armesünderblum'`

NTS: Don't break apostrophe handling, e.g. `dog` & `dog's`

windymilla:
When a trailing apostrophe is followed by an underscore, WF word search was not finding the word, because it is was attempting to avoid finding embedded apostrophes. For example, we don't want to find `don` in `don't`.

Avoiding treating underscore as a word character (which it is by default) fixes the situation.

Fixes #1828

### DIFF
--- a/src/guiguts/word_frequency.py
+++ b/src/guiguts/word_frequency.py
@@ -783,13 +783,13 @@ class WordFrequencyDialog(ToplevelDialog):
             # boundary is a word/space character
             emdash_bound = "—" if newline_word[0] == "—" else r"[^\w\s]"
             left_boundary = (
-                r"(?<!([^\W_]|\w[-'’]))"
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:
Run WF Diacritics on the following text: `_die Armesünderblum'_, the poor-sinner's-flower`
In `master`, it reports `Armesünderblum` but clicking doesn't jump to it, because the apostrophe after it is followed by a "word character", so counts as an embedded apostrophe, like in `don't`

srjfoo (approved):

PR #1836, opened 2026-04-26
Fix "multiple clicks on CP Head/Foot Fix&Hide"

### LINKED ISSUE #1835: Spamming "Fix & Hide" button can make weird stuff happen
1. Load test file: [raw.txt](https://github.com/user-attachments/files/27103600/raw.txt)
2. Run Content Providing, Headers/Footers
3. Hit "Fix & Hide" button as fast as possible several times (5 or 6 at least)

Sometimes, it jumps back to working on the first page in the file. Possibly, the list is being refreshed but the "hidden" items have not been hidden yet, so it's working on the first message in the list, even though that should be hidden.

windymilla:
In Content Providing, Header/Footer fixup, if the Fix&Hide button is pressed rapidly several times, it can appear to reset, with previously hidden messages appearing, or jumping to removing a line from the first page in the book.

This is because when Fix&Hide is done in this dialog, it automatically recalculates and refreshes the dialog, then hides the messages that were previously hidden. The rapid clicks could occur before this process was finished.

Fixed by disabling the Fix buttons when they are clicked and not enabling them until the whole job is done.

Fixes #1835

### DIFF
--- a/src/guiguts/content_providing.py
+++ b/src/guiguts/content_providing.py
@@ -540,6 +540,8 @@ class HeadFootCheckerDialog(CheckerDialog):
         self, process: bool, remove: bool, all_matching: bool
     ) -> None:
         """Overridden from CheckerDialog in order to facilitate re-run."""
+        for btn in self.fix_btn, self.fixall_btn, self.fixrem_btn, self.fixremall_btn:
+            btn["state"] = tk.DISABLED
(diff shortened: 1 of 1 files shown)


windymilla:
Test file & instructions in [linked issue](https://github.com/DistributedProofreaders/guiguts-py/issues/1835)

srjfoo (approved):

PR #1856, opened 2026-05-14
Fix two-finger tap on macOS with Tk9

windymilla:
With Tk9, 2-finger tap is generating mouse 3, whereas it
generated mouse 2 with Tk9. Bind both of these, so it
should work with either Tk version.

Testing of debug version:
In Tk9 environment, and Tk8 environment, do two-finger tap on main text window, then look at output in the terminal window. Compare with Ctrl-click.

### DIFF
--- a/src/guiguts/widgets.py
+++ b/src/guiguts/widgets.py
@@ -1121,7 +1121,7 @@ def mouse_bind(
 ) -> None:
     """Bind mouse button callback to event on widget.
 
-    If binding is to mouse button 3, then if on a Mac bind
+    If binding is to mouse button 3, then if on a Mac also bind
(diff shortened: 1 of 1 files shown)


srjfoo:
`ctrl`+click got nothing on either Tk 8 or 9. (In other apps, `ctrl`+click and two-finger tap with my system setup get the same response.)

Tk8 two-finger tap: `<ButtonPress event num=2 x=199 y=298>` is typical, where the x/y coorinates change depending on where the touchpad pointer is located.

Tk9 two finger tap: `<ButtonPress event num=3 x=199 y=298>`, ditto.

Go figure? 🙄

srjfoo (approved):

PR #1857, opened 2026-05-16
Test ttk-themes

windymilla:
Ignore the fact the theme isn't fully incorporated, and it doesn't look very good, and you can't change theme - only dark.

The question is whether the checkboxes work, i.e. show an empty box for "unset" and either a box with a tick or a filled box for "set"

You'll need to "poetry install"


windymilla:
Testing notes:
1. `poetry install` needed
2. There are lots of themes - we could/should cut down the list
3. If you had "Dark" theme selected, it now chooses "black"
4. If you had "Light" theme selected, it now chooses "scidblue"
5. The "Default" hasn't been properly dealt with. Before we only had dark/light, so it was obvious what to do with Default when the OS switched between light/dark. Now there are 2 dark and lots of light themes. I've just picked "black" and "scidblue" to be selected if you have the Default theme. However, there is no "Default" in the menu, so you can no longer choose "Default". This needs a decision - my personal view is that we could live without "Default" - just choose the theme you want, but that may not be popular with some. We could let the user choose a light and a dark theme, but that's added complication.

windymilla:
I've added the default Tk themes too

> Also add default Tk themes: alt, clam, classic, default
On Mac, add aqua
On Windows, add winnative, xpnative, vista

Also radio button to choose which theme to use Light/Dark/Auto (match OS setting)

Default Light theme is "default"
Default Dark theme is "black"

Please test

- Using a clean GGprefs that all works OK to simulate a new user.
- Also, simulate a user upgrading to this version: with clean GGprefs, first run `master` and set theme Dark, then run this branch, and it should start up with "black". Again, clean GGprefs (make sure it does not contain `theme_name_light` or `theme_name_dark`) run `master` and set theme Light, then check this branch starts with "default" light theme. Finally, same for the "Default" setting in `master` - should choose "black"/"default" depending on your OS setting.

I'm wondering if we should just provide all the themes there are, and let people choose what they like. We should document that some have more "negative space" meaning a dialog might need to be resized so that all of it shows, and some may run more slowly on certain platforms ( 👀 ). Essentially, if you don't like a theme, use a different one. "black" and "default" should be a safe option for everyone.

What do you think? The disadvantage of supplying all themes, is that we can't take them away as easily as we can add them, if we decide we really don't want some of them.

srjfoo:
Tested on the M5 using the above instructions, though not a full usability test at this point.

I'd ditch equilux. It's low enough contrast that honestly even I don't like it other than it uses a ✔️ instead of an ✖️ for checkboxes. And I've confirmed that scrolling is choppy. It appears to be a port from \*nix, and I'm not terribly impressed with the quality of the ttk port. Maybe if they fix it a bit at some point?

windymilla:
@srjfoo - I've removed equilux as suggested.
I've added aqua to the list of dark themes (Mac only), so in theory you could choose aqua for your light theme and your dark theme. Note that it will (probably) reflect your OS setting - I don't think it's possible to force aqua to be dark if your OS setting is light, but you probably wouldn't want that anyway. If you want a dark theme but light OS, use "black". If you want a light theme but dark OS, use almost any theme except "black" and "aqua". Hopefully auto-detect of OS dark/light should work correctly if aqua is one, or both, of your selected themes, and you choose "Default" in the "Use" radio buttons.

srjfoo (approved):

PR #1859, opened 2026-05-20
Add `<main>` tags around whole book

windymilla:
Aids accessibility. After ebookmaker adds the PG boilerplate before/after the book, these tags will continue to mark the "main" part of the document - the book itself.


windymilla:
Testing: Take any pre-HTML file, and use HTML autogenerate. It should get `<main>` immediately after `<body>` and `</main>` immediately before `</body>`

srjfoo (approved):

PR #1861, opened 2026-05-24
[tk9] Improve touchpad scrolling on Mac

tangledhelix:
In Tk9, on Mac at least, scrolling with the touchpad only emits the `<TouchpadScroll>` event, which we did not bind. This PR binds the event to correct the misbehaving scroll speed I observed previously.

The `<MouseWheel>` event handling was not touched, and should function exactly as before. I did limited testing of a mouse-wheel and it seemed to scroll fine.

Testing notes: just try to scroll the text area with a mouse-wheel (if you have one) and with a Mac touchpad (if you have one). The mouse-wheel should behave exactly as before. The trackpad should scroll as most Mac apps do; if you scroll slowly, it does too, if you flick your finger it will go faster.

Test on python.org v3.14.5.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -1317,25 +1317,45 @@ class MainText(tk.Text):
         self.theme_set_tk_widget_colors(self.peer)
 
         # On Tk9 macOS, trackpad gestures fire <TouchpadScroll> not <MouseWheel>.
-        # An instance-level binding is required; the Tk class-level binding alone
-        # does not engage correctly on this widget.
+        # Rate-limit scroll updates to ~60fps by accumulating deltas between frames.
         if is_mac():
 
-            def _touchpad_scroll(event: tk.Event) -> str:
-                def to_signed_16(n: int) -> int:
-                    return n if n < 0x8000 else n - 0x10000
+            def _make_touchpad_handler(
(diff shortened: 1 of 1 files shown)


windymilla:
Looks good - i.e. has no effect on Windows.

tangledhelix:
For benefit of this thread, since I only said it in Slack: this isn't quite right yet. I'm going to mark it as a draft and keep tinkering with it.

tangledhelix:
Mentioned in Slack, but not here - I updated this and now it is working well for me. Removing draft status.

windymilla (approved):

srjfoo (approved):
Some work still needed to get parity between Tk8 and Tk9.

Tk8: mouse and trackpad scrolling work just fine for me. No choppiness. If the main window is in the background, scrolling works fine over scrollbar and main text, but not over line numbers; with the main window in the foreground; scrolling works fine even over line numbers.

Tk9: The above is true for the mouse, but not for the trackpad. That is, everything is okay except scrolling when hovering over line numbers with the trackpad.

tangledhelix:
> Tk9: The above is true for the mouse, but not for the trackpad. That is, everything is okay except scrolling when hovering over line numbers with the trackpad.

Opened #1866 for that.

PR #1864, opened 2026-05-25
Make menu bar visible whatever dialog has focus

windymilla:
Tk9 on Macs only displays the root menu bar when the root window has focus, not if a dialog has focus.
Assign the root menu bar to all dialogs.

Testing: Check if it breaks Tk8 version on Macs. If it does, we'll need `if is_mac() and is_tk9()` but I'd rather avoid explicit Tk9 testing where possible.

### DIFF
--- a/src/guiguts/widgets.py
+++ b/src/guiguts/widgets.py
@@ -74,6 +74,10 @@ class ToplevelDialog(tk.Toplevel):
 
         self.base_title = title
         self.resizable(resize_x, resize_y)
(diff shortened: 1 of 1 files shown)


tangledhelix (approved):
Tested with Python 3.13.5 / Tk 8 and Python 3.14.5 / Tk 9. Menus stay consistent as I open S/R, Bookloupe, WF, Regex Library, etc. Looks good.

srjfoo (approved):
LGTM, also!

PR #1867, opened 2026-05-28
Don't do #1864 fix when Tk version < 9

windymilla:
On Macs, the fix causes problems with Tk 8.6. It's only needed for Tk 9+, so just do it when needed.


srjfoo (approved):

srjfoo (approved):

windymilla:
Note problem was only with python 3.11 & 3.12, not 3.13, so apparently not specifically Tk8/9 related. Fix is OK though.

PR #1869, opened 2026-05-28
Improve theme load failure

windymilla:
Add a slight pause before retrying, and if it still fails, load default theme.
Remove keramic_alt theme - suspicious it causes problems on Windows


windymilla:
Testing notes: Changing themes should not give an exception. Probably not reproducible in `master` but as long as this isn't worse.

srjfoo (approved):

PR #1889, opened 2026-06-25
Add "horizontal rule" option to images dialog

### LINKED ISSUE #1882: Add decorative `hr` feature somewhere
Maybe in the (auto) illo dialog, where we have "decorative", we could also have "thought break" as an option, i.e. 3-way radio button:
Illustration/Decorative/ThoughtBreak

Thought Break would use `hr` (with background image?)

windymilla:
Auto-illustration / HTML images dialogs now have a "Horizontal rule" checkbox, that causes the "illo" to be added as `<hr>` with the illo as a background image. This is intended for where there is an embellished thought break that the PPer wants to preserve instead of just using a plain straight line.

Fixes #1882

Testing notes:
1. Checking/unchecking the new box should make most other widgets in the dialog (like width, alignment, etc) disabled/enabled.
2. Adding regular illustrations should still work
3. Adding a thought break illo should now work, with CSS to control all the common settings, and the file name specified in the style attribute of the hr tag

Example tb illo:
image

### DIFF
--- a/src/guiguts/application.py
+++ b/src/guiguts/application.py
@@ -640,6 +640,7 @@ class Guiguts:
         preferences.set_default(PrefKey.HTML_IMAGE_ALIGNMENT, "center")
         preferences.set_default(PrefKey.HTML_IMAGE_CAPTION_P, True)
         preferences.set_default(PrefKey.HTML_IMAGE_DECORATIVE_ONLY, False)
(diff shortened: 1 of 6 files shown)


srjfoo (approved):

windymilla:
Ebookmaker currently gives errors during epub validation with the code this generates, so changing to Draft for now.

srjfoo (approved):

PR #1891, opened 2026-06-27
Cope with `[][]` regex better

### LINKED ISSUE #1890: `[][]` regex doesn't fail nicely
1. It keeps popping the "highlighting turned off" error box.
2. It should display the regex in red if it's going to fail.

windymilla:
1. Report is by changing the regex to red in the SR dialog - the validation routine wasn't using V1 regex rules
2. Turn off regex highlighting if the regex fails

Fixes #1890 

Testing notes:
1. Make sure your file has some square brackets in it
2. Enter `[][]` in the S/R search field, check "regex" - note that in `master`, it does not get displayed in red text as a warning that it's not legal.
3. Press Highlight All - note that in `master` you get an infinite loop of error dialog.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -3205,16 +3205,18 @@ class MainText(tk.Text):
             logger.error(
                 "Regex timed out. Try changing the regex or flags;\n"
                 "or increase the timeout in the Preferences dialog, Advanced tab.\n\n"
-                "Search highlighting turned off temporarily."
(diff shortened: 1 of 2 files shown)


tangledhelix (approved):
- Regex now turns red
- If I try to "highlight all" as before, I still get the popup dialog, but clicking "OK" dismisses it without entering a loop

Looks good!

srjfoo (approved):
LGTM!

PR #1898, opened 2026-07-12
Add a regex replace escape for regnal numbers

tangledhelix:
The existing `\N` escape converts a Roman number to an Arabic number. Given the books we work on, it's common to encounter Roman numbers which are regnal numbers, i.e., the number in a series of monarchs (kings, queens, popes).

This change implements a new escape (`\M`). I used M for Monarch, since `\R` (for Regnal) was taken already. With support for numbers in the range 1-99, it will take a Roman number as input (e.g. "II", "XIV", "XXIII") and output the regnal string prefixed by "the".

    II    --> the Second
    XIV   --> the Fourteenth
    XXIII --> the Twenty-third

I'm not aware of any real-world regnal numbers higher than 23, but it was as easy to implement the 20s as it was to go through the 90s.

### DIFF
--- a/src/guiguts/search.py
+++ b/src/guiguts/search.py
@@ -42,6 +42,59 @@ PADY = 2
 # Passed into eval call when `\C...\E` used in regex replacement
 lglobal: dict[str, str | int] = {}
 
+# Ordinal words for regnal numbers 1-19, used by `\M...\E` in regex replacement
+REGNAL_ONES = {
+    1: "First",
+    2: "Second",
+    3: "Third",
+    4: "Fourth",
+    5: "Fifth",
+    6: "Sixth",
+    7: "Seventh",
+    8: "Eighth",
+    9: "Ninth",
+    10: "Tenth",
+    11: "Eleventh",
+    12: "Twelfth",
+    13: "Thirteenth",
+    14: "Fourteenth",
+    15: "Fifteenth",
+    16: "Sixteenth",
+    17: "Seventeenth",
+    18: "Eighteenth",
+    19: "Nineteenth",
+}
+# Cardinal words for the tens, 20-90, used by `\M...\E` in regex replacement
+REGNAL_TENS = {
+    2: "Twenty",
+    3: "Thirty",
+    4: "Forty",
+    5: "Fifty",
+    6: "Sixty",
+    7: "Seventy",
+    8: "Eighty",
+    9: "Ninety",
+}
+
+
+def regnal_ordinal(number: int) -> str:
+    """Convert a number from 1 to 99 into its ordinal word form,
+    e.g. 2 --> "Second", 20 --> "Twentieth", 45 --> "Forty-fifth".
+
+    Args:
+        number: Number to convert; must be between 1 and 99 inclusive.
+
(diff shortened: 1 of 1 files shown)


tangledhelix:
Testing notes: I use this regular expression to find roman numbers.

```
(?<=[\s>-])(?!(?:x-ebookmaker|ill|civil|did|mild))([IVXLCDM]+|[ivxlcdm]+)(?![A-Za-z])(\.)?(?=[,\s<-])
```

Use this replace for testing:

```
\M\1\E
```

The most obvious use case in my mind is for abbreviations. An example replace for that case:

```
<abbr title="\M\1\E">\1\2</abbr>
```

tangledhelix:
Finally, a reminder: if this is merged, we'd want to update the wiki documentation to add `\M`, probably [here](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/Search_Menu#Regex_examples_using_extended_backslash_commands).

tangledhelix:
Forgot this in the testing notes: that regex can cope with either upper or lowercase roman numbers, with or without a terminating `.`, and it *should* exclude some commonly encountered words like "civil", "did", etc, that are made up of entirely roman numeral characters. The pathological case is the word "`I`" which is impossible to exclude without more intelligence than a regex pattern can muster.

I'd be interested to hear of any other words to exclude that aren't already in the pattern, too, if you encounter any.

windymilla:
> Finally, a reminder: if this is merged, we'd want to update the wiki documentation to add `\M`, probably [here](https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_2_Manual/Search_Menu#Regex_examples_using_extended_backslash_commands).

Thanks for the link - I normally update the manual just before I do a release

windymilla:
> Testing notes: I use this regular expression to find roman numbers.
> 
> ```
> (?<=[\s>-])(?!(?:x-ebookmaker|ill|civil|did|mild))([IVXLCDM]+|[ivxlcdm]+)(?![A-Za-z])(\.)?(?=[,\s<-])
> ```
An alternative would be to trap Roman numeral construction more precisely than `[IVXLCDM]+` with something like
```
(?=[MDCLXVI])(M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3}))
```

Also, if you put `(?i)` at the start of one of our regexes it makes it case insensitive which sometimes simplifies things.

windymilla (approved):
Code and behavior look good to me - thanks

tangledhelix:
> An alternative would be to trap Roman numeral construction more precisely than `[IVXLCDM]+` with something like
> 
> ```
> (?=[MDCLXVI])(M{0,3}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3}))
> ```

I'll tinker with this; I think it might need to be `{0,4}` for some of these, as you do run into the incorrectly formatted type. My current book has `MCCLXXXX` for 1290, for example, instead of `MCCXC`.

Dan

srjfoo (approved):

PR #1902, opened 2026-07-23
Add "expanded" mode to PPcomp

windymilla:
Make the output more spaced out, with more context, so that it is more similar to PPWB version.

Radio buttons at top of dialog set Compact/Expanded mode for next run.
Switching file should still work, though context lines may end up with an extra/missing leading space or two after switching - shouldn't be a significant problem.


rtonsing:
A sample comparison. Mainly, expanded makes it _much_ more clear that illos are just moved to another location.

compact
expanded

srjfoo (approved):

PR #1910, opened 2026-07-29
Improve ppcomp expanded output over line breaks

windymilla:
Previously, in expanded mode, all the extra text was reported on one line of the dialog, until the line number in "File 1" changed. Now, when an addition in "File 2" is being reported, if the addition crosses a line break, that is reflected in the report.
Also, if a character is inserted at the start of a line, it should be displayed at the start of a line, rather than the end of the previous line. Note that "File 1" still mostly controls where line breaks are displayed - it's only where there's an insert in "File 2" that its line breaks are taken into account.

### DIFF
--- a/src/guiguts/tools/ppcomp.py
+++ b/src/guiguts/tools/ppcomp.py
@@ -71,6 +71,7 @@ FLAG_CH_1_L = "⦓"
 FLAG_CH_1_R = "⦔"
 FLAG_CH_2_L = "⦕"
 FLAG_CH_2_R = "⦖"
+FLAG_NL = "⦑"
 NO_SPACE_BEFORE = "])}.,:;!?"
 NO_SPACE_AFTER = "[({"
 PAIR_RE = re.compile(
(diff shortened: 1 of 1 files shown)


windymilla:
Test file: [eyesthatcannotsee.zip](https://github.com/user-attachments/files/30511850/eyesthatcannotsee.zip)

1. Choose File 1 as the HTML file
2. Choose File 2 as the text file
3. Load HTML file into the main window
4. Run PPcomp
5. Load text file into main window
6. Run PPcomp
7. Choose File 2 as the HTML file
8. Choose File 1 as the text file
9. Load HTML file into the main window
10. Run PPcomp
11. Load text file into main window
12. Run PPcomp

In each case, compare the PPcomp output with `master` - displayed results should either be more or less the same, or improved in this branch. In particular, where additional text in File 2 crosses a line break, the line break should be shown in the results.

The edits made should not affect "Compact" display mode, so you don't need to check all of the above in "Compact" as well as "Expanded". However, you may want to check "Compact" for the case where the above is very different between `master` and this branch.

Screenshots of different display (branch first, then `master`):

image

.

image

.

Annotated version of first screenshot ("Newline from File 2" labels mark the change in this branch):

image

srjfoo (approved):

PR #1911, opened 2026-07-30
Improve speed of Page Sep Fixup

windymilla:
Remove some time-consuming stuff at the cost of slightly more complex code, but worthwhile.

Also, show "Working" label while working.

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -51,6 +51,7 @@ from guiguts.widgets import (
     FileDialog,
     get_global_font,
     themed_style,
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Wow ... about an order of magnitude speed-up on the M1.

windymilla:
> Wow ... about an order of magnitude speed-up on the M1.

Yes - it was taking extra long on longer files, probably proportional to the square of the number of pages, because the code was written in a simple way. Since performance has now shown itself to be a problem, I found 2 or 3 changes to make it more efficient.

PR #1923, opened 2026-08-09
Set Tk selection anchor when using Home/End keys

### LINKED ISSUE #1920: Selection highlighting changes unpredictably when adjusting with Shift + Arrow keys
I have difficulty making precise selections using the keyboard. I expected someone else might report this issue, but I have not seen it mentioned, so I am reporting it.

Please consider this an **accessibility issue**. Reliable keyboard selection behavior is important for users who depend on precise cursor movement and keyboard-based editing.

Steps to reproduce:

**Case 1: Reducing a selection from the end of a line**

1. Position the cursor at the extreme left of a line.
2. Press **Shift + End** to select to the end of the line.
3. Press **Shift + Left Arrow** once.

**Actual behavior:**
The left side of the highlighted area moves right by one character, and the right side moves left by a variable number of characters.

**Expected behavior:**
The left side of the selection should remain fixed, and only the right side should move left by exactly one character.

---

**Case 2: Reducing a selection from the beginning of a line**

1. Position the cursor at the end of a line.
2. Press **Shift + Home** to select to the beginning of the line.
3. Press **Shift + Right Arrow** once.

**Actual behavior:**
The left side of the highlighted area moves right by one character, and the right side moves left by a variable number of characters.

**Expected behavior:**
The left side of the selection should move right by exactly one character, and the right side should remain fixed.

This makes precise keyboard-based editing difficult, especially when trying to select or adjust small portions of text. Consistent selection behavior would improve usability and accessibility.

windymilla:
When used without Shift, Home/End (Cmd-Left/Right on Macs) should set the selection anchor so that future adjustment of the selection works more predictably.

Fixes #1920

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -951,14 +951,7 @@ class MainText(tk.Text):
         self.bind_event(
             "<Escape>", lambda _: self.cancel_drag_sel(), add=True, bind_peer=True
         )
-
-        def set_anchor(_: tk.Event) -> str:
-            """Set anchor point at current position, then allow Shift-click to
-            execute its default behavior."""
-            self.mark_set(TK_ANCHOR_MARK, f"{self.index(tk.INSERT)}")
-            return ""
-
-        self.bind("<Shift-Button-1>", set_anchor)
+        self.bind("<Shift-Button-1>", self.set_tk_selection_anchor)
 
         # Override default left/right/up/down arrow key behavior if there is a selection
         # Above behavior would affect Shift-Left/Right/Up/Down, so also bind those to
@@ -3920,6 +3913,11 @@ class MainText(tk.Text):
         while start := self.search(" +$", start, regexp=True):
             self.delete(start, f"{start} lineend")
 
+    def set_tk_selection_anchor(self, _: tk.Event) -> str:
+        """Set anchor point at current position."""
+        self.mark_set(TK_ANCHOR_MARK, f"{self.index(tk.INSERT)}")
+        return ""
+
     def go_home(self, event: tk.Event) -> str:
         """Handle the Home key. Either move to start of line, or if
         already there, move to first non-space on line."""
@@ -3955,6 +3953,7 @@ class MainText(tk.Text):
         # If no Shift, just moving, so clear selection
         else:
             self.clear_selection()
+            self.set_tk_selection_anchor(event)
         self.see(tk.INSERT)
         return "break"
 
@@ -3986,6 +3985,7 @@ class MainText(tk.Text):
         # If no Shift, just moving, so clear selection
         else:
             self.clear_selection()
+            self.set_tk_selection_anchor(event)
         self.see(tk.INSERT)
         return "break"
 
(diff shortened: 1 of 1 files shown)


tangledhelix:
This still does the wrong thing when (on Mac only) you move the cursor to start/end of line with Ctrl-A (start) or Ctrl-E (end).

tangledhelix:
Strangely, though, it *does* work if you don't involve arrow keys in it!

Works:
* Click in middle of some line
* `Ctrl-A` to jump to start of line
* `Shift-Ctrl-E` to select to end of line
* `Shift-Ctrl-B` to move backward 1 char (selection is now entire line minus last character)

(The same is true in reverse: Ctrl-E, Shift-Ctrl-A, Shift-Ctrl-F.)

Doesn't work:
* Click in middle of some line
* `Ctrl-A` to jump to start of line
* `Shift-Cmd-Right` to select to end of line
* `Shift-Cmd-Left` to move backward 1 char

This leaves the selection from where you clicked until just before the last char in the line.

srjfoo:
> Doesn't work:
> * Click in middle of some line
> * `Ctrl-A` to jump to start of line
> * `Shift-Cmd-Right` to select to end of line
> * `Shift-Cmd-Left` to move backward 1 char
> 
> This leaves the selection from where you clicked until just before the last char in the line.

For the last bullet point, `Shift-Cmd-Left` should take you back to the beginning of the line. Did you mean `Shift-Left`?

tangledhelix:
Yes

srjfoo:
The simple case now works. That is ...
* Click in middle of line
* Use cmd (on Mac) plus arrow1 key to go to one end of the line.
* Use shift+cmd+arrow2 key to select to the other end of the line
* Release cmd and use just shift+arrow1 to trim selection

Where arrow1 and arrow2 are left and right arrow keys (or right and left, depending on which direction the initial move is).

> > Doesn't work:
> > 
> > * Click in middle of some line
> > * `Ctrl-A` to jump to start of line
> > * `Shift-Cmd-Right` to select to end of line
> > * `Shift-Cmd-Left` to move backward 1 char
> > 
> > This leaves the selection from where you clicked until just before the last char in the line.
> 
> For the last bullet point, `Shift-Cmd-Left` should take you back to the beginning of the line. Did you mean `Shift-Left`?

Confirmed this result of Dan's

Also discovered another failure mode:

If, when I click in the middle of a line, I have the shift key down and arrow to the left or the right, then catch myself and change from shift to cmd and go to one end of the line or the other, I see the same behavior as above (which mirrors the original problem).

windymilla:
I'm not certain, since I can't test it, but hopefully this fixes the Ctrl-a/e problem. Please check they work as expected generally, not just this specific case. Also, please re-rest Cmd-left/right because the Ctrl-a/e fix slightly affects that too.

@srjfoo if you test your second bug, that may have been fixed by this. If not, please can you describe it more precisely for me, specifically, what you expect to happen and what you see that is unexpected. I tested it as best I could on Windows, but couldn't do exactly what you describe of course.

srjfoo (approved):
Looks good to me.

tangledhelix (approved):
The case I outlined earlier that didn't work, now works 👍🏻 

The other case I mentioned, which was already working, _still_ works too. LGTM!

PR #1924, opened 2026-08-09
Add `"T`/`"I` and `T'm`/`I'm`, etc to CP Filter

### LINKED ISSUE #1921: Consider CP fix for `"I`, `"I'm`, etc. being OCRed as `"T`, `"T'm`, etc.
Described here: https://www.pgdp.net/phpBB3/viewtopic.php?p=1403237#p1403237

windymilla:
Under the "2/3 letter scannos" sub-toggle, change capital "T" to "I" for some common words, and for standalone "T" following double quotes.

Fixes #1921

Testing: should fix cases described here: https://www.pgdp.net/phpBB3/viewtopic.php?p=1403237#p1403237

### DIFF
--- a/src/guiguts/content_providing.py
+++ b/src/guiguts/content_providing.py
@@ -1103,8 +1103,8 @@ class CPFilteringDialog(CheckerDialog):
         re_exc2 = re.compile(r"'!!")
         re_h = re.compile(r"!!(?=\w)")
         re_exc = re.compile(r"(?<=\w)!(?=\w)")
-        re_ti = re.compile(r"\bT('ll|'m|'d|f|s|t)\b")
-        re_tiw = re.compile(r'"T\b')
+        re_ti = re.compile(r"\bT(['’]ll|['’]m|['’]d|f|s|t)\b")
+        re_tiw = re.compile(r'(["“])T\b')
         re_1i = re.compile(r"(?<![^'\" ])1\b(?!\.)")
         re_0o = re.compile(r"(?<![^'\" ])0\b")
         re_li = re.compile(r"(?<![^'\" ])l\b(?!')")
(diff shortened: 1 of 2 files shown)


srjfoo:
I took the text from the forum post mentioned and added a couple based on real-life testing:

```
One of the common issues I see with Tesseract OCR and
certain fonts is interpreting capital I as capital T.
Would it be possible to add a few bits of regex to
either the filters or the scannos to deal with these?

The most common issues I see are at the beginning of
a line of dialogue, where

"I

is OCR'd as

"T

Edited to add: also T'll, T'm, and T'd, and Tf and Ts and Tt.

Also,

T meaning "I" at the beginning of a line with no quote
     is not changed to "I"

What about 

T'll
```

The results of running the filter (with the two header paragraphs stripped out:

```
"I

is OCR'd as

"I

Edited to add: also I, I, and I, and I and I and I.

Also,

T meaning "I" at the beginning of a line with no quote
     is not changed to "I"

What about

I

```

So -- testing results:
* T'll, T'm, T'd, Tf, Ts and Tt were all changed to `I` with their following characters stripped
* Standalone T at the beginning of a line, was not changed to `I` unless followed by one of the above, in which case it became just a standalone `I` as detailed above. (How likely is "T" to occur standalone, whether at the beginning of a line or in the middle, unless it's a scanno for the capital `I`?

windymilla:
> * T'll, T'm, T'd, Tf, Ts and Tt were all changed to `I` with their following characters stripped

Oops - my replacement regex was wrong. Can't believe I didn't notice that myself! Pushed another commit that should now change `T'll` to `I'll`, etc.

> * Standalone T at the beginning of a line, was not changed to `I` unless followed by one of the above, in which case it became just a standalone `I` as detailed above. (How likely is "T" to occur standalone, whether at the beginning of a line or in the middle, unless it's a scanno for the capital `I`?

I don't feel happy changing all standalone `T` to `I`. The initial request was for `"T` to `"I`, which I believe the branch should do. This is a CP tool, which doesn't aim to fix all scannos, and I think should **strongly** err on the side of not making incorrect changes to valid OCR, especially as some CPers use it fairly automatically, not checking every change it makes. Also, the rounds need something to do!

srjfoo:
> > * Standalone T at the beginning of a line, was not changed to `I` unless followed by one of the above, in which case it became just a standalone `I` as detailed above. (How likely is "T" to occur standalone, whether at the beginning of a line or in the middle, unless it's a scanno for the capital `I`?
> 
> I don't feel happy changing all standalone `T` to `I`. The initial request was for `"T` to `"I`, which I believe the branch should do. This is a CP tool, which doesn't aim to fix all scannos, and I think should **strongly** err on the side of not making incorrect changes to valid OCR, especially as some CPers use it fairly automatically, not checking every change it makes. Also, the rounds need something to do!

That's fair. It would also pick up any standalone Ts as initials, which might be disastrous in some books.

srjfoo (approved):

srjfoo (approved):

PR #1935, opened 2026-08-20
PPcomp: Add option to suppress text file illo text

### LINKED ISSUE #1932: PPcomp: Add "Suppress illo text" option
Currently, the text option to suppress illo markup changes
`[Illustration: My description of the illo]` to `My description of the illo`
and removes empty `[Illustration]` tags completely.

Recently, as some PPers have begun to add more meaningful alt text to their HTML illos, they have then used that alt text as their text file illo description, where historically the text illo description was often briefer or omitted when there wasn't a caption. Since alt text (correctly) isn't considered part of the text of the HTML file, this gives rise to a diff from PPcomp, i.e.
```
<img alt="My description of the illo"...>
```
and
```
[Illustration: My description of the illo]
``` 
lead to a diff.

The current behavior is correct, since the text in the `[Illustration:...]` tag can often be a caption (which doesn't cause a diff). With the recent change it would be good to have the option to suppress this copied-over alt text. The suggested option would thus suppress the text file illo caption/description/text. 

Note that schemes like trying to extract the alt text from the HTML file and count it as "real text" are probably not a good idea. Illos can/should often have captions AND alt text, and there may also be many situations where the HTML alt text is not appropriate for use in the text file.

windymilla:
To support when PPer copies alt text from HTML
to use as text in `[Illustration:...]` markup in text file.

Fixes #1932


windymilla:
Testing notes:
1. Use an HTML file containing an image that has alt text but no caption, and PPcomp it with the text file.
2. Ensure the text file just says `[Illustration]` (i.e. no caption)
3. This should not cause a diff
4. Now copy the alt text into the text file: `[Illustration: Copied alt text]`
5. In `master`, and in this branch by default, this should cause a diff, reporting that the text file has text that the HTML does not.
6. Turn on the new option, and the diff should go away.

I'm also hoping for a test file from the originator of the request to be certain it addresses the issue they raised.

windymilla:
Test files from requester: [faerie.zip](https://github.com/user-attachments/files/31970584/faerie.zip)

Ensure Text: Suppress Illo Tags is ON and HTML: Add Illo Tags is OFF

First run with new Suppress Illo Text option OFF - note diff at lines 409/683, and at 512/1104, etc., consisting of the text "caption" within the `[Illustration...]` tag in the text file.

Now turn on new option, those diffs should be gone.

You may see other diffs appear (e.g. lines 1277/2415) - this is where the HTML has a caption, rather than merely alt text. So because you have suppressed Text file illo text, the HTML has the caption but the text file doesn't. There's no obvious way to deal with a mixture of different cases (HTML has only alt text, or only caption, or both alt text and caption, or no caption and empty alt text, and Text has text or not). Whatever PPcomp does is likely to generate diffs for some combinations of those.

srjfoo (approved):

PR #1937, opened 2026-08-21
Update install instructions

### LINKED ISSUE #1933: Possible updates to install information
See comment [here ](https://www.pgdp.net/phpBB3/viewtopic.php?p=1404025#p1404025)about Windows installation not setting PATH. Documentation could/should say what to do if PATH not set, e.g. use `python -m...`

Something similar [here ](https://www.pgdp.net/phpBB3/viewtopic.php?p=1404039#p1404039)relating to Macs

Is it worth documenting "If you want an icon to click". Some Mac instructions [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1404422#p1404422) and could explain how to set up a shortcut on Windows.

windymilla:
Windows Python install uses Install Manager now.
MacOS users can use Automator to create an app.

@srjfoo - Please check changes, especially suggested alternative for Automator script which states the path in the python command, rather than exporting an updated PATH variable. If it works OK, it seems simpler for users. Thanks.
I also left a [**...] note in point 1 of the macOS user install, about the PATH. Do we need to tell the user how to add to the PATH, and if so, what should we say?

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,20 +77,6 @@ previous version of GG2).
 
     c. Type `guiguts` or `~/.local/bin/guiguts` to run GG2.
 
-6. If you would like to have an icon that runs GG2,
-
-    a. Launch the Automator app from your Applications folder
-
-    b. When prompted to "Choose a type for your document," select "Application."
-    Then enter "Run Shell Script" from the list of possible actions.
-
-    c. In the "Run Shell Script" panel, type `/usr/local/bin/python3 -m guiguts`
-
-    d. Press the "Run" button to test the script.
-
-    e. Choose "Save" to save the new application. You can now click on it to open,
-    like any other Mac app.
-
 ### Linux (Debian/Ubuntu)
 
 Python version 3.13 or above is recommended if possible, though GG has been tested in
--- a/README.md
+++ b/README.md
@@ -18,17 +18,24 @@ Python.org now uses a Python install manager which can be installed using the
 Microsoft Store App.
 The website says "For most users, this is the recommended approach."
 
-2. Once the install manager is installed, in a command window, you should be
+2. Open a Windows Command Prompt window into which you will type the commands below.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,24 +40,41 @@ previous version of GG2).
...
 
 1. Install Python 3.13 or above from [python.org](https://www.python.org/), not a Homebrew-installed Python.
-GG has been tested in versions up to 3.14.4. Using 3.14.5 or above is not recommended for Mac users at the
-moment, because Tcl/Tk 9.03 is bundled with it. Work is ongoing to support this version within GG2.
+GG has been tested in versions up to 3.14. You may need to add the location of the Python installation to
+your PATH [**MORE DETAIL HERE]
 
 2. In a terminal window, type `pip3 install guiguts` (or `pip3 install guiguts --upgrade` to upgrade from a
-previous version of GG2).
+previous version of GG2). If you cannot use `pip3`, try `python3 -m pip install guiguts` instead.
```
tangledhelix: The second variant should work either way; we could simplify the instructions by omitting `pip3` mention entirely?

README.md
```diff
@@ -33,24 +40,41 @@ previous version of GG2).
...
 
-3. In a terminal window, type `guiguts` to run GG2.
+3. In a terminal window, type `guiguts` to run GG2. If this does not work, due to not being on your PATH,
```
tangledhelix: Again, the second variant should work always, even if the first won't (and it seems the Python installers have shifted something so maybe it will become more common that just `guiguts` won't work reliably). We could simplify the instructions by only giving one suggestion.

tangledhelix (approved):
Left some suggestions, will let @windymilla and/or @srjfoo decide whether to make them or leave as-is...

windymilla:
Fixes #1933

tangledhelix (approved):

srjfoo (approved):

PR #1942, opened 2026-08-23
Avoid exception with lowercase `doctype`

### LINKED ISSUE #1939: PPcomp gives a traceback if HTML header is lowercase
It is reported that this also happens in the PPWB version of PPcomp.

I think there are two bugs:
1. It should obviously accept `<!DOCTYPE html>` case-insensitively
2. If it doesn't find that string, it then attempts to parse it as HTML4, rather than HTML5. That parsing fails, but it should output a polite message, not give a traceback

Example files to PPcomp: [ducky-entertains.zip](https://github.com/user-attachments/files/31349465/ducky-entertains.zip)

To test:
1. Try varying the case of the first line of HTML
2. Try changing the first line so it doesn't match, even case-insensitively, e.g. `<!DOCTYPE rubbish>`

windymilla:
1. If there is a doctype error, e.g. `doctype rubbish`, output sensible error
2. Ignore case of `doctype` line of HTML file

Fixes #1939

### DIFF
--- a/src/guiguts/tools/ppcomp.py
+++ b/src/guiguts/tools/ppcomp.py
@@ -515,7 +515,15 @@ class PPcompChecker:
                 )
                 # Extract line & column, message and optional supplementary data (dict)
                 # Display in dialog
-                for (line, col), msg, data in exc.parse_errors:
+                for err in exc.parse_errors:
+                    # Might be HTML5 parsed errors, or HTML4 - structures are different
+                    try:
+                        (line, col), msg, data = err  # html5parser
+                    except TypeError:
+                        msg = err.message  # etree parser
+                        data = ""
+                        line = err.line
+                        col = err.column
                     if data and isinstance(data, dict):
                         extra = (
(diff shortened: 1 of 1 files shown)


windymilla:
Test files: [ppcomp_test.zip](https://github.com/user-attachments/files/31733524/ppcomp_test.zip)

In `master` this causes an exception
In this branch, I get the following - the HTML parser fails to parse the HTML file:

```
Failed to parse HTML file
Fix the errors below in the HTML file before re-running ppcomp

260.25: Tag nav invalid
321.87: Tag aside invalid
324.46: Tag section invalid
331.87: Tag aside invalid
333.87: Tag aside invalid
335.87: Tag aside invalid
337.87: Tag aside invalid
339.87: Tag aside invalid

Check complete
```

srjfoo (approved):
This setup doesn't appear to fail on macOS, but if the PR fixes it for Win/Linux, go for it.

Issue #1866, opened 2026-05-26
[tk9] TouchpadScroll on Mac doesn't work on number gutter


tangledhelix:
As reported in #1861, touchpad scroll over the number gutter isn't working in Tk9. MouseWheel still works, and in Tk8 either work (because the touchpad is emitting MouseWheel in that case).

windymilla:
Fixed by #1943

PR #1943, opened 2026-08-23
[mac] Touchpad scroll improvements/support for scrollbar and line number gutter

tangledhelix:
When using Tk9, the new touchpad events are introduced (for supported platforms; which right now I believe are only Macs). A while back, this event was implemented for the main text document.

This PR contains 2 updates:
1. When hovering over the text document scroll bar, scrolling with touchpad was previously much faster than over the main document. This change makes them about equivalent instead.
2. When hovering over the line number gutter, the touchpad previously did nothing at all. This change implements touchpad scrolling there, and makes it roughly as fast as the text document scrolling.

This should only happen on Macs, but may as well do a quick test in Windows.

It appears to work properly in my testing, but @srjfoo should verify in both Tk8 and Tk9.

### DIFF
--- a/src/guiguts/maintext.py
+++ b/src/guiguts/maintext.py
@@ -253,6 +253,45 @@ class TextLineNumbers(tk.Canvas):
         self.bind("<Enter>", lambda _: bind_mouse_wheel(self, self.textwidget))
         self.bind("<Leave>", lambda _: unbind_mouse_wheel(self))
 
+        # On Tk9 macOS, trackpad gestures fire <TouchpadScroll> rather than
+        # <MouseWheel>, and (unlike MouseWheel) it's delivered directly to the
+        # widget under the pointer, so bind it here rather than via the
+        # Enter/Leave + bind_all approach used for <MouseWheel> above. Rate-limit
+        # to ~60fps by accumulating deltas between frames, matching the text
+        # widget and scrollbars, so scrolling feels the same over the gutter.
+        if is_mac():
+            pending_x: float = 0.0
+            pending_y: float = 0.0
+            after_id: Optional[str] = None
+
+            def to_signed_16(n: int) -> int:
+                return n if n < 0x8000 else n - 0x10000
+
+            def flush_scroll() -> None:
+                nonlocal pending_x, pending_y, after_id
+                after_id = None
+                y_scroll = int(-pending_y)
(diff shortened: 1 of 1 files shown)


tangledhelix:
It's worth testing this with a scroll-wheel mouse also.

tangledhelix:
Testing notes:

* Hover over the main text window, scroll with touchpad; make note of the scrolling speed
* Hover over the text window scroll bar, scroll with touchpad. Speed should be similar to main window.
* Hover over the number line gutter, scroll with touchpad. Speed should be similar to main window.
* Should also work with split views.
* Should probably try with a scroll-wheel mouse to make sure it still works too
* Should probably try in tk8 also to make sure it doesn't break anything

windymilla (approved):
I tried it with Python 3.11 and 3.14, but don't have Tk9 of course.
Since the main part is protected by `is_mac()` it should be OK.
The code looks fine too.
Thanks

windymilla:
@tangledhelix - Are you happy this fixes #1866?

tangledhelix:
I'll do some more testing when I'm back home in a couple of days. I have a mouse there too.

srjfoo (approved):
I was able to test only with Tk9 -- I apparently no longer have a test environment running Tk8.

I was able to test and confirm that scrolling over line numbers, body text and scroll bar all work just fine in both split and non-split views.

tangledhelix:
> I was able to test only with Tk9 -- I apparently no longer have a test environment running Tk8.
> 
> I was able to test and confirm that scrolling over line numbers, body text and scroll bar all work just fine in both split and non-split views.

I setup a Py3.11 / Tk 8 environment and it seemed to work fine with the touchpad. I'll wait and confirm with a scroll mouse later this week before considering it all OK.

tangledhelix:
## Tk9 tests

Python 3.14.7, Tk 9.0.4:

- mousewheel behaves the same as on master branch
- trackpad scrolls approximately the same speed on all 3 regions (text, scrollbar, gutter)

## Tk8 tests

Python 3.11.16, Tk 8.6.18:

- mousewheel behaves the same as on master branch
- trackpad scrolls approximately the same speed on all 3 regions

@windymilla I'm satisfied this fixes #1866 although I'm not sure the commit messages linked them; it might not auto-close that ticket.

PR #1945, opened 2026-08-25
Handle `/#[n]`, etc., in Unmatched Block Markup

windymilla:
Some of the checks were only looking for the default `/#`, rather than customized versions, like `/#[8]`, `/#[4.8,24]`, etc.

This meant you could get incorrect reports, either saying there was a problem when there wasn't, or vice versa.

### DIFF
--- a/src/guiguts/misc_tools.py
+++ b/src/guiguts/misc_tools.py
@@ -944,6 +944,9 @@ def unmatched_html_markup() -> None:
 def unmatched_block_markup() -> None:
     """Check for unmatched block markup."""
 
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:

Tools-->Unmatched-->Block Markup

Should report unmatched: [blocktest1.txt](https://github.com/user-attachments/files/31970351/blocktest1.txt)
Should not report unmatched: [blocktest2.txt](https://github.com/user-attachments/files/31970352/blocktest2.txt)

In `master`, neither of them report unmatched.

srjfoo (approved):

PR #1947, opened 2026-09-06
Report `.DS_store` better in PPhtml

### LINKED ISSUE #1938: [mac] Ignore .DS_Store file in pphtml
Discussed [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1404601#p1404601).

PPhtml (and potentially other checks) should ignore the presence of a `.DS_Store` file on Mac. The system creates them any time the folder is opened in Finder (the file explorer). So it's a bit of a waste of PPer effort to repeatedly delete them just to make a validator happy (IMO).

windymilla:
Instead of mistaking it for an image file, just tell the user they need to delete it before they upload.

Fixes #1938

### DIFF
--- a/src/guiguts/tools/pphtml.py
+++ b/src/guiguts/tools/pphtml.py
@@ -17,6 +17,8 @@ from guiguts.maintext import maintext, HighlightTag
 from guiguts.preferences import preferences, PrefKey, PersistentBoolean
 from guiguts.utilities import IndexRange, IndexRowCol, sing_plur
 
+DS_STORE = ".DS_store"  # Don't report hidden Mac file as a bad image file
+
 
 class PPhtmlCheckerDialog(CheckerDialog):
     """Dialog to show PPhtml results."""
@@ -149,8 +151,13 @@ class PPhtmlChecker:
             if " " in filename:
                 errors.append(f"  Filename '{filename}' contains spaces")
                 test_passed = False
-            if re.search(r"\p{Lu}", filename):
+            if re.search(r"\p{Lu}", filename) and filename != DS_STORE:
                 errors.append(f"  Filename '{filename}' not all lower case")
+                test_passed = False
+            elif (
+                filename == DS_STORE
+            ):  # Hidden Mac file should not be uploaded for PPV/WW
+                errors.append(f"  Filename '{filename}' must be deleted before upload")
                 test_passed = False
 
         # Make sure all are JPEG, PNG or SVG images
@@ -189,6 +196,8 @@ class PPhtmlChecker:
                     errors.append(f"  File '{filename}' is not valid SVG format")
                     test_passed = False
                     continue
+            elif filename == DS_STORE:
+                continue
             else:
                 errors.append(
                     f"  File '{filename}' does not have extension jpg, png or svg"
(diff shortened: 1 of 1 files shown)


src/guiguts/tools/pphtml.py
```diff
@@ -17,6 +17,8 @@
 from guiguts.preferences import preferences, PrefKey, PersistentBoolean
 from guiguts.utilities import IndexRange, IndexRowCol, sing_plur
 
+DS_STORE = ".DS_store"  # Don't report hidden Mac file as a bad image file
```
tangledhelix: This should be `.DS_Store`, the case mismatch causes it to fail currently.

tangledhelix (changes requested):

src/guiguts/tools/pphtml.py
windymilla: Thanks. Obviously copied it wrong and I don't have a Mac for testing. I'll put up a fix tomorrow

srjfoo:
@windymilla -- a couple more test zips for you. I'll put them in your dpscans -- this is one of the former test files I still have lying around, and when I attempted to add it here, I was told that the file was too large. Easier to just put it in dpscans than to worry about it.

The first one (genthouse-test.zip), I messed around with and did a standard macOS zip on it, then looked at it with unzip `-l`. It contains no `.DS_Store` files, but it contains a ton of `__MACOSX` folders and their contents. I can't see these using `ls -la` on a Mac (@tangledhelix, if you have advice on showing them, let me know), I assume, because if opened on a Mac, they get used by the OS. The second (genthouse-test-clean.zip) doesn't have them, and was compressed using JuneCloud's Create Clean Archive utility.

Hopefully you'll be able to see the extra garbage on a Windows system, even if I can't see it. I'm pretty sure that stuff needs to be cleaned out, too, so it isn't just the `.DS_Store` that's a problem. If you try to run pphtml on the html in `genthouse-test.zip`, do you get errors?

windymilla:
> @windymilla -- a couple more test zips for you. I'll put them in your dpscans -- this is one of the former test files I still have lying around, and when I attempted to add it here, I was told that the file was too large. Easier to just put it in dpscans than to worry about it.
> 
> The first one (genthouse-test.zip), I messed around with and did a standard macOS zip on it, then looked at it with unzip `-l`. It contains no `.DS_Store` files, but it contains a ton of `__MACOSX` folders and their contents. I can't see these using `ls -la` on a Mac (@tangledhelix, if you have advice on showing them, let me know), I assume, because if opened on a Mac, they get used by the OS. The second (genthouse-test-clean.zip) doesn't have them, and was compressed using JuneCloud's Create Clean Archive utility.
> 
> Hopefully you'll be able to see the extra garbage on a Windows system, even if I can't see it. I'm pretty sure that stuff needs to be cleaned out, too, so it isn't just the `.DS_Store` that's a problem. If you try to run pphtml on the html in `genthouse-test.zip`, do you get errors?

I see the junk files. Superficially, they look like a duplicate of what's in the zip, i.e. one 1KB  file for each file in the zip, but they just contain some Mac metadata for each file. Here's the contents of one of them:
```
    Mac OS X            	   2       7                                     ATTR      7   È   o                  È   4  $com.apple.metadata:_kMDItemUserTags     ü   ;  com.apple.quarantine bplist00¡WGreen
2
                            q/0082;693e2b99;Slack;DBFE7567-80AF-45AB-B240-DEA0FAC4D088 
```

I think we can just ignore them unless they start causing problems.

src/guiguts/tools/pphtml.py
windymilla: New commit to fix the case mismatch - thanks

tangledhelix (approved):

srjfoo (approved):
Looks good -- definitely a clearer error message.

PR #1958, opened 2026-09-19
Improve reporting of uncommon characters

windymilla:
Ignore the verbose setting for this check - if a char appears a lot, it's probably deliberate. If it's not, the PPer will still get the first few reports followed by a "...more" and they can use S/R to search & fix them.

Also, add some exceptions for characters commonly used in the languages most often PPed at DP.


windymilla:
Testing notes:

This file (taken from the code) might be good to try:
```
    weirdos_not_language: Dict[str, str] = {
        "de": "äöüÄÖÜß",
        "es": "áéíóúüñÁÉÍÓÚÜÑ",
        "fr": "àâæçéèêëîïôœùûüÿÀÂÆÇÉÈÊËÎÏÔŒÙÛÜŸ",
        "it": "àèéìîòóùÀÈÉÌÎÒÓÙ",
        "nl": "ëï",
        "pt": "áàâãçéêíóôõúüÁÀÂÃÇÉÊÍÓÔÕÚÜ",
    }
```

If you PPtext it as an "en" file, you get lots of reports. If you change the lang to "fr", for example, you should not get any reports about the "fr" line. You can also say "fr es" to indicate multiple languages, and you shouldn't get reports about either of those lines.

srjfoo (approved):

PR #1959, opened 2026-09-19
Remove unused code from ppcomp.py

windymilla:
Enough time has elapsed to tidy this up a bit

Add typechecking

Also fixes bug that would probably crash if you did footnotes separately and the footnotes contained quotes, fractions, etc.


windymilla:
Testing: Try ppcomping two files. Ideally try files with at least 1 footnote that contains `½` character, and use the `Extract Footnotes` option.

srjfoo (approved):

